PR #98, opened 2019-10-21
Workflow status enum should be converted to int and not string when persisting with EntityFramework

cando:
This is needed since the SQL [_WorkflowInstances]_ table expects:
`[Status] int NOT NULL`


sfmskywalker:
Or perhaps we can fix the migrations such that it generates a string not null column instead of an int not null?

cando:
@sfmskywalker You mean something like that? Forgive me if I've done something wrong with migrations, but it's the first time I've used them :)

sfmskywalker:
Exactly like that 👍🏻

But since we haven’t released anything yet officially, feel free to first delete the Migrations folder first and then generate them again like you did. 

If your unsure then that’s fine too. Just let me know.

cando:
It turned out that the InitialCreate migration was alredy correct.

The only correction needed is on the [sqlserver-create.sql](https://github.com/elsa-workflows/elsa-core/blob/develop/src/persistence/Elsa.Persistence.EntityFrameworkCore/sqlserver-create.sql) in the root folder.

sfmskywalker:
Ah, good find! No worries, I appreciate you looked into it.

Issue #124, opened 2019-11-04
Prevent activities from being registered multiple times


sfmskywalker:
Currently, it's possible to register activity types multiple times, which is problematic when the activity resolver tries to project them into a dictionary (a duplicate key exception will occur, since multiple activity types with the same type name will exist).

Possible solutions:

1. Do nothing, except require the developer to explicitly register the activities they want to use.
2. Update `AddActivity<T>` sich that it will replace any existing registration with a new one.
3. Allow multiple activityies of the same type to be registered, but make the activity resolver take the last one that was registered (or doing a DISTINCT first).

See also #117 for background.

PR #128, opened 2019-11-04
Fixed bug where multiple activity registrations of the same type would cause duplicate key exception

jamesdoran:
To reproduce, run Sample12 on 1.1.1 branch and you'll get a duplicate key exception.

The registration was being performed twice, once by AddElsa() and again by AddUserTasks(). You could probably remove the AddUserTasks() extension method altogether.


jamesdoran:
Apologies, I've just seen #124 . feel free to put in a different fix if you think that's better. I only saw this because Sample12 was registering UserTask twice.

sfmskywalker:
No worries, this is a good fix. We can revisit the activity registration story later if we deem it necessary. Thanks for the PR 👍🏻

Issue #137, opened 2019-11-08
Fault workflows on evaluation exception


jamesdoran:
It would be good to know when there's an error evaluating an expression in an activity, rather than returning the error in the result of the expression so that the workflow can be faulted.

The following workflow tries to make use of an `unknownFunction()` javascript function:

```
public class WorkflowWithExpressionError : IWorkflow
{
    public void Build(IWorkflowBuilder builder)
    {
        builder
            .StartWith<WriteLine>(activity => activity.TextExpression = new JavaScriptExpression<string>("unknownFunction()"))
            .Then<WriteLine>(activity => activity.TextExpression = new LiteralExpression("Completed"));
    }
}
```
When run, the expression evaluates to the error string and completes successfully, moving to the next activity.

By throwing an exception in the catch block of `WorkflowExpressionEvaluator.EvaluateAsync`, the workflow goes to a faulted state, with the fault information set.

@petedavis and I took a look at the Orchard code and there didn't appear to be a try..catch there at all for Liquid Expressions, so it did go to a faulted state.

I'll send over a PR for discussion

sfmskywalker:
I agree - the workflow should indeed fault when the expression fails to be evaluated.

PR #138, opened 2019-11-08
#137 WorkflowExpressionEvaluator Exception

jamesdoran:


sfmskywalker:
I actually swallowed the exception for a reason, but I assume you want the exception to be thrown for another reason?

sfmskywalker:
Ah never mind, I just noticed the issue you created 👍 Checking it now.

Issue #123, opened 2019-11-04
Feature Request: XMLContentFormatter + POCO loader


Andale-online:
Hey dude...

I need XML as ContentFormatting - there is no such feature, isnt it?

Even more it should be able to cast the requests in a type safe manner (e.g. against a poco dll)

I did this before for Web Api - think i can do this and commit it. I rly rly rly need that.... Do you wanna have this?

Greetings

sfmskywalker:
We don’t have it yet, but I have a sneaky suspicion you will send a PR with that feature 😉

Perhaps you could describe it in a little bit more detail / provide a use case?

For example, are you posting an XML payload to a workflow, and you want that XML to be deserialized into a POCO?

If so, we need to come up with a way for a user to specify what type of poco it should be. And maybe by default it would be deserialized to a JObject, so that you could easily access its members from Liquid and JavaScript expressions.

Andale-online:
Err.. For sure... I will do so.

But first: I cloned the JsonContentFormatter to:

`namespace Elsa.Activities.Http.Formatters
{
    public class XMLContentFormatter : IContentFormatter
    {
        public int Priority => 2;
        public IEnumerable<string> SupportedContentTypes => new[] { "application/xml", "text/xml" };

        public Task<object> ParseAsync(byte[] content, string contentType)
        {
            string rootNodeName = "";
            using (XmlReader xmlReader = XmlReader.Create(new MemoryStream(content)))
            {
                if (xmlReader.MoveToContent() == XmlNodeType.Element)
                    rootNodeName = xmlReader.Name;
            }

            if(rootNodeName != "")
            {
               
                
            }

            return null;


        }
    }
}
`

But my formatter doesnt get loaded to 

        public ReceiveHttpRequest(
            IHttpContextAccessor httpContextAccessor,
            IEnumerable<IContentFormatter> contentFormatters)

Any idea?

Andale-online:
Too early in the mornin... Forget the question... Think i found it

Andale-online:
.AddSingleton<IContentFormatter, XMLContentFormatter>()

sfmskywalker:
Good morning 😉

Andale-online:
Mornin.... :-D

So... In theory i got the receiver formatter working :-D Got some dll loading issue, but that will be solved after another coffee...

Andale-online:
[...]

PR #141, opened 2019-11-09
#123 xml formatterand poco loader

Andale-online:


src/activities/Elsa.Activities.Http/Activities/WriteHttpResponse.cs
```diff
@@ -106,6 +111,17 @@ public WorkflowExpression<string> ResponseHeaders
             set => SetState(value);
         }
 
+
+        /// <summary>
+        /// Variable name for model body - Read model body from named variable. 
+        /// </summary>
+        [ActivityProperty(Hint = "Variable name for content model - Read content model from named variable.")]
+        public string InputVariable
```
sfmskywalker: We don't need this property - Variables can be retrieved via expressions. E.g. `MyVariableName`.

See also: https://elsa-workflows.github.io/elsa-core/docs/expressions-javascript (I added this documentation today, so I don't blame you for not knowing)

src/activities/Elsa.Activities.Http/Formatters/DefaultContentFormatter.cs
```diff
@@ -13,5 +13,10 @@ public Task<object> ParseAsync(byte[] content, string contentType)
         {
             return Task.FromResult<object>(content);
         }
+
+        public Task<byte[]> ParseAsync(object content, string contentType)
```
sfmskywalker: It looks like your fork is out of date - the `IContentFormatter` no longer exists. It was replaced with `IHttpRequestBodyParser` and `IHttpResponseBodyParser`.

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Elsa.Activities.Http.Formatters
+{
+    public static class XMLContentFormatterCache
+    {
+        public static List<Type> TypeCache { get; set; }
+    }
+
+
+    public class XMLContentFormatter : IContentFormatter
```
sfmskywalker: Please update your fork with latest commits from `develop`, then implement `IHttpRequestBodyParser`.
Sorry for the inconvenience.

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private static void GetTypeCache()
+        {
+            if (XMLContentFormatterCache.TypeCache == null)
+            {
+                string baseDirectory = AppDomain.CurrentDomain.BaseDirectory;
+                foreach (string fileName in Directory.EnumerateFiles(baseDirectory, "*.dll", SearchOption.AllDirectories))
+                {
+                    AssemblyLoadContext.Default.LoadFromAssemblyPath(Path.Combine(baseDirectory, fileName));
```
sfmskywalker: I don't like this at all. What's the reason we need to load all assemblies like this?

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
```diff
@@ -0,0 +1,123 @@
...
+
+
+
+                XMLContentFormatterCache.TypeCache =
+                    AppDomain.CurrentDomain.GetAssemblies().
+                    Where(a =>
+                        {
+                            string name = a.GetName().Name;
+                            return !name.StartsWith("System") &&
+                                    !name.StartsWith("Microsoft") &&
```
sfmskywalker: This logic seems way too specific and not open for extensibility.

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            get => GetState<string>(null, "MethodName");
+            set => SetState(value, "MethodName");
+        }
+
+        protected override async Task<ActivityExecutionResult> OnExecuteAsync(
+            WorkflowExecutionContext context,
+            CancellationToken cancellationToken)
+        {
+            await Task.Delay(1);
```
sfmskywalker: What's the reason for delaying?

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+            var inputValues = InputVariableNames.Split(',').Select(s => context.GetVariable(s)).ToArray();
+            return Execute(context, inputValues);
+        }
+
+        protected override ActivityExecutionResult OnResume(WorkflowExecutionContext context)
```
sfmskywalker: I'm not sure I see why you need to implement `OnResume`. Does the activity return a `HaltResult` somewhere?

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            var inputValues = InputVariableNames.Split(',').Select(s => context.GetVariable(s)).ToArray();
+            return Execute(context, inputValues);
+        }
+
+        private ActivityExecutionResult Execute(WorkflowExecutionContext workflowContext, object[] receivedInputValues)
+        {
+            string path = Directory.EnumerateFiles(AppDomain.CurrentDomain.BaseDirectory, AssemblyName, SearchOption.AllDirectories).FirstOrDefault();
+            Assembly assembly = Assembly.LoadFrom(path);
```
sfmskywalker: Why not just use `Type.GetType("Assembly Qualified Type Name")` instead of loading assemblies from disk?

src/activities/Elsa.Activities.Reflection/Activities/SplitObject.cs
```diff
@@ -0,0 +1,90 @@
...
+                foreach (var Property in Properties)
+                {
+                    try
+                    {
+                        object PropValue = FollowPropertyPath(splitObject, Property);
+                        context.SetVariable(Property, PropValue);
+                    }
+                    catch (Exception ex)
+                    {
+                        // dont shout ex, because it could be a typo... but how do we tell the user?
```
sfmskywalker: You should remove the try/catch block and just let the exception occur. It will cause the workflow to fault, which is fine - the user will see the exception when inspecting the workflow instance, which will record the exception.

sfmskywalker: If we hide the exception (which is really caused by a user error which they should fix), it might lead to hard to find bugs for the user. Better to fail fast and let the user fix their mistake.

src/activities/Elsa.Activities.Reflection/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,15 @@
+using Elsa.Activities.Reflection.Activities;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Elsa.Activities.Reflection.Extensions
```
sfmskywalker: I like this category!

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        [ActivityProperty(Hint = "The variables to use as parameters, seperated by comma in order of method call.")]
+        public string InputVariableNames
+        {
+            get => GetState<string>(null, "InputVariableNames");
+            set => SetState(value, "InputVariableNames");
+        }
+
+        [ActivityProperty(Hint = "The name of the variable to store the returned value into.")]
+        public string OutputVariableName
```
sfmskywalker: Given the fact that this activity represents a function to execute, I think it's only appropriate to have this activity store the function result into the activity's Output, rather than letting the user specify a variable name to store it into. The result value can then be accessed using an expression, e.g. `{{ Activities.MyMethodActivity.Result }}`

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                var staticClassType = assembly.DefinedTypes.Where(t => t.Name == ClassName).FirstOrDefault();
+                var staticMethod = staticClassType.DeclaredMethods.Where(m => m.Name == MethodName).FirstOrDefault();
+                receivedOutput = staticMethod.Invoke(null, receivedInputValues);
+            }
+            else
+            {
+            }
+
+            workflowContext.CurrentScope.SetVariable(OutputVariableName, receivedOutput);
```
sfmskywalker: Related to my previous comment about the output variable, let's do this instead:

`Output.SetVariable("Result", receivedOutput)`.

Then you can access the received output using the following Liquid expression: "{{ Activities.MyActivity.Result }}"

We need a similar function for JavaScript, so we can do something like `activityOutput('MyActivity', 'Result')`, or if there's only one value in the activity's Output dictionary, just do `activityOutput('MyActivity')`.

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (IsStaticClass)
+            {
+                var staticClassType = assembly.DefinedTypes.Where(t => t.Name == ClassName).FirstOrDefault();
+                var staticMethod = staticClassType.DeclaredMethods.Where(m => m.Name == MethodName).FirstOrDefault();
+                receivedOutput = staticMethod.Invoke(null, receivedInputValues);
+            }
+            else
+            {
+            }
```
sfmskywalker: Looks like you can remove this block.

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
```diff
@@ -0,0 +1,123 @@
...
+            using (MemoryStream stream = new MemoryStream())
+            {
+                xmlSerializer.Serialize(stream, content);
+                return Task.FromResult<byte[]>(stream.ToArray());
+            }
+        }
+
+        private static Type GetRootType(string rootNodeName)
+        {
+            return XMLContentFormatterCache.TypeCache.Where(t => t.Name == rootNodeName).FirstOrDefault();
```
sfmskywalker: I'm very skeptical we need all the assembly loading logic. It seems overkill just to find a single type to match against the XML root element. I understand why you do it, but what if the XML element contains a root node whose corresponding type lives in an assembly not included in the type cache because it's filtered by assemblies starting with the name "Elsa"?

sfmskywalker:
Since you need to update your fork anyway, would it be OK for you to submit each feature as a separate PR?

src/activities/Elsa.Activities.Http/Activities/WriteHttpResponse.cs
Andale-online: I dont like that... This way my objects get mangeled twice... Once for reading and once for writing the object. Its not needed and costs a lot of CPU. How about a new expression type called Variable?

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
Andale-online: Let ne Take a look in that.

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
Andale-online: No way... I have to fill it. There goes the non static handling

src/activities/Elsa.Activities.ControlFlow/Activities/Switch.cs
```diff
@@ -24,7 +25,10 @@ public class Switch : Activity
         public Switch(IWorkflowExpressionEvaluator expressionEvaluator)
         {
             this.expressionEvaluator = expressionEvaluator;
-            Cases = new List<string>();
+            Cases = new List<string>()
+            {
+                "!unknown!"
```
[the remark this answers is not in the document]
Andale-online: Yup

src/activities/Elsa.Activities.Http/Activities/ReceiveHttpRequest.cs
```diff
@@ -85,6 +85,16 @@ public bool ReadContent
             set => SetState(value);
         }
 
+        /// <summary>
+        /// Variable name for model body - Read model body to named variable. 
+        /// </summary>
+        [ActivityProperty(Hint = "Variable name for content model - Read content model to named variable.")]
+        public string OutputVariable
+        {
+            get => GetState<string>(null, "OutputVariable");
```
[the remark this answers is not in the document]
Andale-online: Thx

src/activities/Elsa.Activities.Http/Activities/ReceiveHttpRequest.cs
```diff
@@ -85,6 +85,16 @@ public bool ReadContent
...
+            set => SetState(value, "OutputVariable");
```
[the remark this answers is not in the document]
Andale-online: Thx

src/activities/Elsa.Activities.Http/Formatters/DefaultContentFormatter.cs
Andale-online: Oh.... okay... ill check that

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
Andale-online: We need some assembly loading, because GetType() didnt get me the requested type (think because its not used somewhere). So imho we need to load the dll. The only thing we should remove is the lazy loading. I think a ModelType and a binary path maybe enough, so we just load the named binary (if needed) and pass the type to the deserializer

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
Andale-online: Yup, will do so....

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
Andale-online: See other comments on this.

src/activities/Elsa.Activities.Http/Formatters/XMLContentFormatter.cs
Andale-online: Can be removed if we supply the wanted binary and the ModelType

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
Andale-online: Just giving the system time to breath... can be removed, but then the async gives a warning (no awaiting)

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
Andale-online: Nope, dont think so... Just a n00bish copy from another module. Can be removed?

src/activities/Elsa.Activities.Reflection/Activities/ExecuteFunction.cs
Andale-online: If the assembly's not loaded this doesnt work... If i e.g. place three poco dll into the server (without referincing it in the project) they wont get loaded. Should at least load an (optional) binary from specified path to dll. I even dont like the lazy loader anymore since it costs a lot of performance

src/activities/Elsa.Activities.Reflection/Activities/SplitObject.cs
Andale-online: Yup, you're right...

src/activities/Elsa.Activities.Reflection/Extensions/ServiceCollectionExtensions.cs
Andale-online: Thx

Andale-online:
"OR, instead of creating a new activity, we add the model binding ability to the ReceiveHttpRequest, where we add an optional property called e.g. ModelType. When specified, it performs the model binding logic as just described."

AND a field called path to dll - then we are both happy and i can remove the dll lazy loader.

ABOUT the xml loader:

i used the base asp xml route before - it suxx: it crashes on pocos, base configuration is insecured, yada yada yada
If we want to use the base asp route we need to create the input and output processors for xml by hand or we create something that will be insecure and hard to handle for the end user...

Andale-online:
If you tell me which feature goes into what pr i will try to split.
Since i wasnt developing on branch develop i have to update a lot - and working on elsa isnt my main goal ;-)
Let me update, i will see what i can do...

Andale-online:
Merged develop

Andale-online:
Damned... The merge broke my Code. I think we have to go with seperate activities for type safe Handling. Its also okay for me so theres less overhead = more Performance

Andale-online:
Bzw lmnt1 ist also my Account, Gitta Check that

Andale-online:
Needs some cleanup. Ill clos3 this pr

sfmskywalker:
Thanks for all the hard work @Andale-online, much appreciated!

Issue #134, opened 2019-11-07
MasstTransit WorkflowConsumer cannot consume scoped service IWorkflowInvoker


petedavis:
Issue running Sample08 project

Not too sure how to fix this immediately as I'm sure the workflow invoker needs to be scoped to web requests. However this then makes it an issue trying to consume it form the WorkflowConsumer in MassTransit activity.

```
System.AggregateException
  HResult=0x80131500
  Message=Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.CreateOrder] Lifetime: Singleton ImplementationType: Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.CreateOrder]': Cannot consume scoped service 'Elsa.Services.IWorkflowInvoker' from singleton 'Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.CreateOrder]'.) (Error while validating the service descriptor 'ServiceType: Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.OrderShipped] Lifetime: Singleton ImplementationType: Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.OrderShipped]': Cannot consume scoped service 'Elsa.Services.IWorkflowInvoker' from singleton 'Elsa.Activities.MassTransit.Consumers.WorkflowConsumer`1[Sample08.Messages.OrderShipped]'.)
  Source=Microsoft.Extensions.DependencyInjection
  StackTrace:
   at Microsoft.Extensions.DependencyInjection.ServiceProvider..ctor(IEnumerable`1 serviceDescriptors, ServiceProviderOptions options)
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider(IServiceCollection services, ServiceProviderOptions options)
   at Microsoft.Extensions.DependencyInjection.DefaultServiceProviderFactory.CreateServiceProvider(IServiceCollection containerBuilder)
   at Microsoft.Extensions.Hosting.Internal.ServiceFactoryAdapter`1.CreateServiceProvider(Object containerBuilder)
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
[...]

PR #148, opened 2019-11-14
MassTransit Fixes

petedavis:
This fixes issue #134 and a few others.

I have a TODO comment in there where I think behaviour has changed. Can you look at the ReceiveMassTransitMessage class. Im not too sure if OnCanExecute can be validated in the current state. Maybe the message type needs to be inserted at another input key?

### DIFF
--- a/src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
+++ b/src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
@@ -32,14 +32,11 @@ namespace Elsa.Activities.MassTransit.Activities
 
         protected override bool OnCanExecute(WorkflowExecutionContext context)
         {
-            var message = context.Workflow.Input[Constants.MessageInputKey];
+            var messageTypeName = context.Workflow.Input[Constants.MessageTypeNameInputKey];
+            var messageInputType = System.Type.GetType(messageTypeName.ToString());
             var messageType = MessageType;
-
-            // TODO: Workout what needs to be done here as the message input is now a JObject,
-            // and not the original message received from the bus.
-
-            // return message != null && messageType != null && message.GetType() == messageType;
-            return true;
+            
+            return messageInputType != null && messageType != null && messageInputType == messageType;
         }
 
         protected override ActivityExecutionResult OnExecute(WorkflowExecutionContext context)
--- a/src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
+++ b/src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
@@ -32,10 +32,14 @@ namespace Elsa.Activities.MassTransit.Activities
 
         protected override bool OnCanExecute(WorkflowExecutionContext context)
         {
-            var message = context.Workflow.Input["message"];
+            var message = context.Workflow.Input[Constants.MessageInputKey];
             var messageType = MessageType;
 
-            return message != null && messageType != null && message.GetType() == messageType;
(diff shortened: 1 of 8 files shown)


src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
```diff
@@ -32,10 +32,14 @@ public Type MessageType
 
         protected override bool OnCanExecute(WorkflowExecutionContext context)
         {
-            var message = context.Workflow.Input["message"];
+            var message = context.Workflow.Input[Constants.MessageInputKey];
             var messageType = MessageType;
 
-            return message != null && messageType != null && message.GetType() == messageType;
+            // TODO: Workout what needs to be done here as the message input is now a JObject,
```
sfmskywalker: Good catch. I'm thinking we might just update `WorkflowConsumer<T>` to include the assembly-qualified message type name as part of the workflow input in addition to the message itself, like so:

```csharp
input.SetVariable("MessageType", typeof(T).AssemblyQualifiedName);
```

Then the `OnCanExecute` method can read the message type from the workflow input and compare that.

src/activities/Elsa.Activities.MassTransit/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -19,38 +19,43 @@ public static IServiceCollection AddRabbitMqActivities(this IServiceCollection s
...
+
             return Bus.Factory.CreateUsingRabbitMq(
                 bus =>
                 {
-                    var options = sp.GetRequiredService<IOptions<RabbitMqOptions>>();
-                    var host = bus.Host(new Uri(options.Value.Host), _ => { });
+                    var options = sp.GetRequiredService<IOptions<RabbitMqOptions>>().Value;
+                    var host = bus.Host(new Uri(options.Host), h =>
+                    {
+                        h.Username(options.Username);
```
sfmskywalker: Would it perhaps be better if we allow the user to either include the credentials as part of the connection string, or separately? Or will this code allow for that already (and it still works if option.Username / option.Password are empty)?

petedavis: Default behaviour inside MassTransit is to use guest/guest if nothing is provided. Which is what was happening before.

Most MassTransit samples have a username password setting in the app.config, or hadcode guest/guest to be explicit about the password, so thought it would be best to do so here.

src/activities/Elsa.Activities.MassTransit/Activities/ReceiveMassTransitMessage.cs
petedavis: Yeah sounds good. I'll hand over the type in the input from the consumer.

src/activities/Elsa.Activities.MassTransit/Extensions/ServiceCollectionExtensions.cs
sfmskywalker: I'm fine with that, just as long as it still works when the user uses the credentials as part of the AMQP connection string.

sfmskywalker:
Looking good! Let me know when it's ready to be merged.

src/activities/Elsa.Activities.MassTransit/Extensions/ServiceCollectionExtensions.cs
petedavis: I didnt even know you could set the username and password in the connection string.

I have now tested both use cases with explicit username and password, and just a host URI with username and password as defined here: https://www.rabbitmq.com/uri-spec.html

petedavis:
Sample08 up and running. Ready for merge.

PR #155, opened 2019-11-14
GetVariable(string name, Type type)

Andale-online:


src/core/Elsa.Abstractions/Models/Variables.cs
```diff
@@ -29,10 +30,15 @@ public JToken GetVariable(string name)
             return ContainsKey(name) ? this[name] : default;
         }
 
-        public T GetVariable<T>(string name)
+        public object GetVariable(string name, string typeName)
```
sfmskywalker: Why pass in a typeName and not a `Type` itself?

Andale-online: Err.. ye... Err.... You're right 😃 I'll fix it

src/core/Elsa.Abstractions/Models/WorkflowExecutionScope.cs
```diff
@@ -18,6 +18,7 @@ public void SetVariable(string variableName, object value)
         }
 
         public T GetVariable<T>(string name) => Variables.GetVariable<T>(name);
+        public object GetVariable(string name, string typeName) => Variables.GetVariable(name, typeName);
```
sfmskywalker: Do we still need this, or can it be changed to accept a `Type` as well?

Issue #152, opened 2019-11-14
Problem: Please help... Response has started


Andale-online:
I get an Fault "Response has already started" while sitting at the XMLHandler...

Tried with ExecuteAsync and Execute .... Same result

```
            var response = httpContextAccessor.HttpContext.Response;

            if (response.HasStarted)
                return Fault("Response has already started");
```

What the...?

sfmskywalker:
Do you have a link for me to the source code? I tried to find it in your fork, but I gave up 🤷‍♂

Andale-online:
Damned, had to reorganize /rebase my fork and local git - and exactly in these 20mins you accessed... 8-)

sfmskywalker:
Well, at least that gave me a giggle 😄

Andale-online:
:-D

I marked the position in the PR

Andale-online:
This can be closed and is followed up by: #160

PR #156, opened 2019-11-14
WIP: Feature xml activities

Andale-online:
Better PR for checking #152

### DIFF
--- a/src/activities/Elsa.Activities.Http/Activities/WriteXmlHttpResponse.cs
+++ b/src/activities/Elsa.Activities.Http/Activities/WriteXmlHttpResponse.cs
@@ -1,0 +1,170 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Xml.Serialization;
+using Elsa.Attributes;
+using Elsa.Design;
+using Elsa.Expressions;
+using Elsa.Extensions;
+using Elsa.Results;
+using Elsa.Services;
+using Elsa.Services.Models;
+using Microsoft.AspNetCore.Http;
+
+namespace Elsa.Activities.Http.Activities
+{
+    [ActivityDefinition(
+        Category = "HTTP",
+        DisplayName = "Write HTTP XML Response (type-safe!)",
+        Description = "Write an HTTP XML response. (type-safe!)",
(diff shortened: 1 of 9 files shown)


src/activities/Elsa.Activities.Http/Activities/WriteXmlHttpResponse.cs
```diff
@@ -0,0 +1,170 @@
...
+
+
+        protected override async Task<ActivityExecutionResult> OnExecuteAsync(
+            WorkflowExecutionContext workflowContext,
+            CancellationToken cancellationToken)
+        {
+            var response = httpContextAccessor.HttpContext.Response;
+
+            if (response.HasStarted)
+                return Fault("Response has already started");
```
Andale-online: This fails, because response.HasStarted == true

sfmskywalker:
I'm not at all convinced we need an almost-duplicate version of the general-purpose `ReceiveHttpRequest` just to handle XML.

The need to be able to match a root XML element name against a .NET type seems too application-specific.

IMO it would be better if the `ReceiveHttpRequest` itself would be able to receive the XML payload, just like it can receive JSON payloads and Form payloads.

Other activities can then access these payloads in their native forms:

* JSON as JObject object (converted to ExpandoObject to allow dot notation)
* XML as XElement
* Form as key/value pairs (dictionary of StringValues)

I do see one complimentary activity type that will meet your need:

* BindModel

`BindModel` would have one `ModelType:Type` property` and one `Expression:WorkflowExpression` property evaluating to the value to be bound against.

It would evaluate the expression and inspect the value type returned. If its JSON, it will deserialize it into the specified type. If its XML, same thing. And of course same story for the Form values. In fact, I think we might be able to reuse actual MVC model binding for this.

Regarding the `AssemblyName` property and the need to actually load the DLL: why is this necessary? Aren't all the types you use already part of the app domain if you reference these assemblies from your application project (directly or indirectly)?

Alternatively, you could consider changing the `ReceiveXmlHttpRequest` into a `BindHttpRequest` activity that does the model binding I described.

Let me know if you would like to discuss this further over Skype or anything. Alternatively, I'm happy to implement what I proposed myself. Meanwhile you can of course use the proposed activity as-is.

Andale-online:
I think we'll discuss this later. Do you have an idea about the response already started thingy?

Andale-online:
Stop thinking about it... I think i got it...

Andale-online:
We can close this... I'll submit a "bug report"

PR #161, opened 2019-11-15
Changed Variables.cs to full IDictionary with ConcurrentDictionary<string, object> backing-store

Andale-online:
Changed Variables.cs to full IDictionary with ConcurrentDictionary<string, object> backing-store
This should be a 1:1 drop in and work even in resumed flows.
Now i get objects (the originals) and you get JTokens *CoughSweat*

BTW i think it may perform better because convert to jtoken is only done when needed. (Dont kill me if not)

This should even be a lit more thread-safe - (there should be some locks somewhere i think to be 100% safe....)

Lotsa shoulds... :-D Could you give it a look please? This would straighten the GetVariable mess in my flows and reduces converts a lot...

### DIFF
--- a/src/core/Elsa.Abstractions/Models/Variables.cs
+++ b/src/core/Elsa.Abstractions/Models/Variables.cs
@@ -1,12 +1,32 @@
-﻿using System.Collections.Generic;
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Elsa.Models
 {
-    public class Variables : Dictionary<string, JToken>
+    public class Variables : IDictionary<string, JToken>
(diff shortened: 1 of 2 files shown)


sfmskywalker:
Did you test it also with long running workflows where values are restored from persistence upon resuming workflows? Just want to make sure that the expectation of being able to store an object of a given type is also then returned in that type when rehydrates.

Andale-online:
Err... Got a sample to test?

sfmskywalker:
Sure, try this: https://github.com/elsa-workflows/elsa-samples/blob/master/src/Elsa.Samples.UserRegistration/user-registration-workflow.json

Perhaps easiest to clone the entire example and replace the Elsa package reference with full Elsa source code from your branch.

Andale-online:
Got it running and did registration.

sfmskywalker:
Allright, then we'll give it  go. Thanks!

PR #163, opened 2019-11-16
Fix response message workflow instance correlation

petedavis:
Something I missed on the last pull request


sfmskywalker:
Great catch 👍🏻

PR #168, opened 2019-11-17
Refactorings and fixes

sfmskywalker:
* Variables are now stored including their type name for automatic
   deserialization into their actual type.
   This is important for use in expressions.

* **SplitObject** and **ExecuteMethod** now use workflow expressions
   instead of variable names. This allows for more flexibility.

* Improved **Join** activity to allow proper function within looping
   constructs.


sfmskywalker:
@Andale-online I refactored a few things here and there. The most important change is the way variables are now stored by including their type name. Which means that upon deserialization, the variables are automatically deserialized into their original types. I provided `Sample20` to see how it works.

Notice I removed the `AssemblyPath` property. My rationale being that it seems a bit arbitrary to me to have to load a DLL for this activity, and not any others. As you can see with Sample20, it's not necessary for it to work. If for some reason you do need to dynamically load an assembly, then you probably need to do so at the application level in case you're doing dynamic plugin loading.

In any case, I hope you'll be able to find a moment and have a look before I merge it in. If you have any concerns, let me know. I'm also on Gitter: [Gitter](https://gitter.im/elsa-workflows/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)

Andale-online:
Variables are now stored including their type name for automatic
    deserialization into their actual type.
    This is important for use in expressions.
 
          /me likes this one!
____________________________________________________________________________

    SplitObject and ExecuteMethod now use workflow expressions
    instead of variable names. This allows for more flexibility.

           needs some fixing on my side - but thats okay. but it in the branch and i will love to check

There's only one point :-D Big one...

I give u a use case: 

 - We have a XML / Json Service (Doesnt matter)
 - 3 Version of schema definition (1.0, 1.1, 1.2) in different DLLs - not created / editable by myself. POCOs are named the same way but have difference between versions.
 - Trying to deserialize a POCO of a specific version will fail / not be possible.

This (IMHO) is not possible without directing to the right assembly... There are a lot of possibilities to handle this "in one of the correct ways" (or any nicer way), but honestly "talking about a flexible and dynamic" system... You can't expect a system like Elsa to be recompiled just to add some functionality.

If you agree to me - let's rather talk about some plugin concept :-) Even the activities could register themselves without the need to put some code in your project and recompile. In this manner there could be a Plugin that just loads libraries by configuration to a given name. Would even probably be faster and safer than loading the library on demand.

Greetings... Thanks for the nice work!

Andale-online:
Aaaaaand:

With the loaded assemblies under control we could pass them to a combobox :-D

Aaaaaaaaaaand:

With some reflection we could give assistance in configuration by loading assembly's methods and display them :-D

Or a tree of methods.....

Or....

I'll stop now :-D

PR #173, opened 2019-11-18
Fix issue with property expressions throwing invalid cast exceptions

petedavis:
Expression was returning the JValue of the property and was unable to convert to a string.

This was occurring in Sample08 in the `SendEmail` activity. The expression `order.customer.email` was returning a `JValue` from the `JavaScriptExpressionEvaluator`, which then throws in the `WorkflowExpressionEvaluatorExtensions.EvaluateAsync` method trying to cast the `JValue` to a `string`.

### DIFF
--- a/src/scripting/Elsa.Scripting.JavaScript/Services/JavaScriptExpressionEvaluator.cs
+++ b/src/scripting/Elsa.Scripting.JavaScript/Services/JavaScriptExpressionEvaluator.cs
@@ -13,6 +13,7 @@ using Jint;
 using Jint.Native;
 using MediatR;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using NodaTime;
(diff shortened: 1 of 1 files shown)


src/scripting/Elsa.Scripting.JavaScript/Services/JavaScriptExpressionEvaluator.cs
```diff
@@ -117,13 +118,18 @@ private object ConvertValue(JsValue value, Type targetType)
...
+                        return JsonConvert.DeserializeObject(json, targetType, serializerSettings);
+                    }
+                    case JValue jValue:
+                        return jValue.Value;
+                    default:
+                        return obj;
                 }
-
-                return obj;
             }
```
sfmskywalker: Thanks for all of these fixes! Much appreciated.

PR #175, opened 2019-11-18
Scheduled MassTransit message delivery

petedavis:
Currently working on trying to schedule messages with a sample. Creating a replication of the MassTransit cart expiry sample.

Not completely functional yet, but thought I would leave a pull request open.

### DIFF
--- a/Samples.sln
+++ b/Samples.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28010.2036
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29509.3
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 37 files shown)


sfmskywalker:
This is awesome 🔥

petedavis:
OK @sfmskywalker I think this is getting close. Just have to inject the schedule message endpoint and provide configuration for that.

I have had to make changes to message correlation as when a scheduled message is published, the correlation id on the context was the scheduled message token id. That is the ID that you use to cancel the scheduled message. Ultimately I dont think that value can be used within the context of ELSA, so I have adapted the conventions used in MassTransit to get properties to use as the correlation id.

This then allows the workflow creator to control correlation by always using the CorrelateBy<Guid> interface, or by adding a CorrelationId, CommandId, or EventId property to the messages.

Let me know if you spot any issues.

sfmskywalker:
I think this is looking great. Nothing jumps out to me as a potential issue, so I think we just need to take it for a spin and see if anything gets thrown at us.

Let me know if you think it's ready to be merged.

petedavis:
Ready to be merged!

sfmskywalker:
Maybe we can use DateTime and NodaTime directly in JS - I never tried, but should work now that I think of it. For NodaTime we’d just have to add an IClock instance to the context.

But that’s something I’ll look into separately - I’m fine with the added JS methods.

Issue #180, opened 2019-11-19
Add lastExpressionResult() javascript function


jamesdoran:
When activities are transitioning based on an expression, it's difficult to debug what's happened. In the case of the switch activity, there's a "default" case providing a catch-all. 

I'd like to be able to have access to the evaluated expression result for logging/debugging.

PR to follow/discuss

sfmskywalker:
@jamesdoran I think I can close this, but please do reopen if there's still something there. Thanks!

PR #181, opened 2019-11-19
Added lastExpressionResult() javascript function #180

jamesdoran:
Not sure if there's a nicer way to get the Evaluators to set the last expression result other than copying in to each evaluate function.

Tested all existing evaluators with following workflow snippet:
```
.Then<WriteLine>(x => x.TextExpression = new LiteralExpression("literal"))
.Then<WriteLine>(x => x.TextExpression = new JavaScriptExpression<string>("'Last expression: ' + lastExpressionResult()"))

.Then<WriteLine>(x => x.TextExpression = new LiquidExpression<string>("liquid"))
.Then<WriteLine>(x => x.TextExpression = new JavaScriptExpression<string>("'Last expression: ' + lastExpressionResult()"))

.Then<WriteLine>(x => x.TextExpression = new JavaScriptExpression<string>("'javascript'"))
.Then<WriteLine>(x => x.TextExpression = new JavaScriptExpression<string>("'Last expression: ' + lastExpressionResult()"))
```

Output:
```
literal
Last expression: literal
liquid
Last expression: liquid
javascript
Last expression: javascript
```

### DIFF
--- a/src/core/Elsa.Abstractions/Models/WorkflowExecutionScope.cs
+++ b/src/core/Elsa.Abstractions/Models/WorkflowExecutionScope.cs
@@ -8,6 +8,7 @@ namespace Elsa.Models
         }
 
         public Variable LastResult { get; set; }
+        public Variable LastExpressionResult { get; set; }
         public Variables Variables { get; }
 
         public void SetVariable(string variableName, object value) => Variables.SetVariable(variableName, value);
--- a/src/core/Elsa.Abstractions/Services/Models/WorkflowExecutionContext.cs
+++ b/src/core/Elsa.Abstractions/Services/Models/WorkflowExecutionContext.cs
@@ -83,6 +83,8 @@ namespace Elsa.Services.Models
         public void SetLastResult(object value) => SetLastResult(new Variable(value));
         public void SetLastResult(Variable value) => CurrentScope.LastResult = value;
 
+        public void SetLastExpressionResult(object value) => CurrentScope.LastExpressionResult = new Variable(value);
+
         public void Start()
         {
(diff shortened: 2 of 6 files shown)


sfmskywalker:
What’s the reason we need this? Can you describe a scenario?

jamesdoran:
Sure!

I'm working on a project where I'm building an Elsa workflow in code. Some of the workflow is user-driven and other parts are automated. As the workflow progresses, I use a custom activity to build up a history of what happened and store it outside of Elsa.

For the automated steps, I'm making use of the switch activity. An expression will be run and an outcome selected. If there's a problem with the expression (if an outcome doesn't match the expression result), I'll fall back and handle with user-input, but I'd like to add the reason for this to the history e.g. 

`"Requested user input because {expressionResult} was not found in valid outcomes: {outcomesCsv}"`

I realise this is a non-standard use-case. Is there another way I could achieve this? The expression evaluation uses context information, so it's difficult to get hold of the value once it's passed.

sfmskywalker:
I see. I agree this is a valid use case and should be enabled. I'd like to dig a little bit into this myself to see if this is the best way to go, or if there might be a more elegant approach to expose expression results. TBC

jamesdoran:
Ok great, thanks for that. I agree this isn’t the most elegant solution. I’ll look forward to seeing what you come up with 🙂

sfmskywalker:
I'm wondering if it would suffice to just store the expression result of the `Switch` activity in the activity's output? That way you can reference it from other activities using expressions, e.g. `MySwitch.Result`.

jamesdoran:
Great suggestion and so much easier to implement. I've created new PR #193

PR #187, opened 2019-11-20
Add the ability to publish events via MassTransit

petedavis:
I have also made a small change to try and maintain using the consume context which is available in the current scope when consuming a message from the bus. 

If this consume context is not available in the current scope, then the bus is used.

This allows for maintaining a continuous conversation ID if available.

### DIFF
--- a/src/activities/Elsa.Activities.MassTransit/Activities/PublishMassTransitMessage.cs
+++ b/src/activities/Elsa.Activities.MassTransit/Activities/PublishMassTransitMessage.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Elsa.Attributes;
+using Elsa.Expressions;
+using Elsa.Results;
+using Elsa.Services;
+using Elsa.Services.Models;
+using MassTransit;
+
+namespace Elsa.Activities.MassTransit.Activities
+{
+    [ActivityDefinition(
+        Category = "MassTransit",
+        DisplayName = "Publish MassTransit Message",
+        Description = "Publish an event via MassTransit."
+    )]
+    public class PublishMassTransitMessage : Activity
+    {
+        private readonly ConsumeContext consumeContext;
+        private readonly IBus bus;
+        private readonly IWorkflowExpressionEvaluator evaluator;
+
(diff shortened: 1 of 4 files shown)


src/activities/Elsa.Activities.MassTransit/Activities/PublishMassTransitMessage.cs
```diff
@@ -0,0 +1,76 @@
...
+        DisplayName = "Publish MassTransit Message",
+        Description = "Publish an event via MassTransit."
+    )]
+    public class PublishMassTransitMessage : Activity
+    {
+        private readonly ConsumeContext consumeContext;
+        private readonly IBus bus;
+        private readonly IWorkflowExpressionEvaluator evaluator;
+
+        public PublishMassTransitMessage(ConsumeContext consumeContext, IBus bus, IWorkflowExpressionEvaluator evaluator)
```
sfmskywalker: Just being curious: how does the service provider know what `ConsumeContext` to inject?

sfmskywalker:
Very slick.

src/activities/Elsa.Activities.MassTransit/Activities/PublishMassTransitMessage.cs
petedavis: After this comment I now have reservations that this might not work with all container implementations, so I might test a little further.

Basically MassTransit registers `ConsumeContext` interface as scoped. So when you are within the scope of a consumer, the container will contain the `ConsumeContext instance.

Here I have modified Sample08 to publish a message before the timer starts. You can see that I have a handle to the consume context, and the callstack starts from the consume of the `CreateOrder` command.
image

However after the timer, there is no consume context within the current scope to publish the `OrderShipped` event.
image

So this event will be published using the bus.

petedavis: Well if Unity container works, then I would assume that all others should also work.

image

sfmskywalker: Great answer. I didn’t know this about MT.

Issue #179, opened 2019-11-19
Allow turning off Jint's AllowClr function


jamesdoran:
Would be great to have the ability to turn this off. 
Am I right in thinking this feature opens up some potential security issues?

sfmskywalker:
Sure, we can expose this as a setting to the application host. It can certainly open up security issues, depending on who's using the application in what context.

jamesdoran:
Can now be configured as follows:

```
new ServiceCollection()
    .AddElsa()
    .WithJavaScriptOptions(o => o.Configure(x => x.AllowClr = true))
```

PR #190, opened 2019-11-21
Add script options configuration (#179)

jamesdoran:
I've given this a crack using the options builder pattern, but I'm interested to know if you think there's a better/tidier way of doing it.

I tried adding another optional parameter to `AddElsa` extension method, but that didn't look great, as I then needed to provided the param name with both the existing param and the new one being optional:

`services.AddElsa(scriptOptions: options => options.Configure(x => x.AllowClr = false)`

What I've commited is a separate extension method. 

```
.AddElsa()
.WithJavaScriptOptions(x => x.Configure(o => o.AllowClr = false))
```

If it's not called, it will default to allowing CLR use as per the existing version.

### DIFF
--- a/src/scripting/Elsa.Scripting.JavaScript/Extensions/JavaScriptServiceCollectionExtensions.cs
+++ b/src/scripting/Elsa.Scripting.JavaScript/Extensions/JavaScriptServiceCollectionExtensions.cs
@@ -17,11 +17,11 @@ namespace Elsa.Scripting.JavaScript.Extensions
                 .AddNotificationHandlers(typeof(JavaScriptServiceCollectionExtensions));
         }
 
-        public static IServiceCollection ConfigureJavaScriptOptions(this IServiceCollection services, Action<OptionsBuilder<ScriptOptions>> options)
+        public static IServiceCollection WithJavaScriptOptions(this IServiceCollection services, Action<OptionsBuilder<ScriptOptions>> options)
         {
             var scriptOptions = services.AddOptions<ScriptOptions>();
             options(scriptOptions);
-            
+
             return services;
         }
     }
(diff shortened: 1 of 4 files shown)


sfmskywalker:
That looks very elegant to me. Let's bring it in 👍

src/scripting/Elsa.Scripting.JavaScript/Extensions/JavaScriptServiceCollectionExtensions.cs
```diff
@@ -13,5 +16,13 @@ public static IServiceCollection AddJavaScriptExpressionEvaluator(this IServiceC
                 .TryAddProvider<IExpressionEvaluator, JavaScriptExpressionEvaluator>(ServiceLifetime.Scoped)
                 .AddNotificationHandlers(typeof(JavaScriptServiceCollectionExtensions));
         }
+
+        public static IServiceCollection ConfigureJavaScriptOptions(this IServiceCollection services, Action<OptionsBuilder<ScriptOptions>> options)
```
sfmskywalker: What's the difference between this method and the `WithJavaScriptOptions` in `ElsaServiceCollectionExtensions`? The method signature and implementation look the same.

jamesdoran: Sorry, I should have only included the one in the JavaScript service collection extensions. There's now only `WithJavaScriptOptions` in `Elsa.Scripting.JavaScript`

jamesdoran:
Apologies @sfmskywalker, @petedavis just informed me I should have done a squash commit merge

sfmskywalker:
No worries at all @jamesdoran.

jamesdoran:
@sfmskywalker Did you want me to move the extension method into `Microsoft.Extensions.DependencyInjection` namespace like AddElsa() so it's more easily accessible?

sfmskywalker:
@jamesdoran That would be good, thanks. I always forget about that one :)

PR #195, opened 2019-11-22
Fix link on Orchard Core doc

agriffard:
The new url is https://orchardcore.readthedocs.io/en/dev/docs/reference/modules/Workflows/


sfmskywalker:
Thanks Antoine!

Issue #203, opened 2019-11-29
Startup Activity


sfmskywalker:
The Startup activity automatically executes workflows when the workflow host starts up. This would be equivalent to using a TimerEvent activity with 0 seconds.

Andale-online:
Made a PR for this. Think you have to include the Act. project to elsa.sln (missing in pr)
Greetings

sfmskywalker:
Thanks!

PR #206, opened 2019-12-02
Feature: Startup Activities

Andale-online:
Hey dude...
Had some troubles with me fork... Had to delete a lot.

Here's the Startup Activity #203 

Greetings

### DIFF
--- a/src/activities/Elsa.Activities.Startup/Activities/Startup.cs
+++ b/src/activities/Elsa.Activities.Startup/Activities/Startup.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Elsa.Attributes;
+using Elsa.Expressions;
+using Elsa.Extensions;
+using Elsa.Results;
+using Elsa.Services;
+using Elsa.Services.Models;
+using NodaTime;
+
+namespace Elsa.Activities.Startup.Activities
+{
+    [ActivityDefinition(
+        Category = "Startup",
+        Description = "Triggers at startup."
+    )]
+    public class Startup : Activity
+    {
+
+        public Startup()
+        {
+        }
+
+
+        protected override ActivityExecutionResult OnExecute(WorkflowExecutionContext context)
+        {
+            return Halt();
+        }
+
+        protected override async Task<ActivityExecutionResult> OnResumeAsync(WorkflowExecutionContext context, CancellationToken cancellationToken)
+        {
+            return Done();
+        }
+    }
+}
--- a/src/activities/Elsa.Activities.Startup/Activities/Startup.cs
+++ b/src/activities/Elsa.Activities.Startup/Activities/Startup.cs
@@ -17,18 +17,7 @@ namespace Elsa.Activities.Startup.Activities
     )]
     public class Startup : Activity
     {
-
-        public Startup()
-        {
-        }
-
-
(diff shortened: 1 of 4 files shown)


src/activities/Elsa.Activities.Startup/Elsa.Activities.Startup.csproj
```diff
@@ -0,0 +1,30 @@
...
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <PackageVersion>1.0.0</PackageVersion>
+        <LangVersion>latest</LangVersion>
+        <Authors>Elsa Contributors</Authors>
+        <Description>
+            Elsa is a set of workflow libraries and tools that enable super-fast workflowing capabilities in any .NET Core application.
+            This package provides the following Console activities:
+            
+            * ReadLine
+            * WriteLine
```
sfmskywalker: Maybe replace these with:

```markdown
* Startup
```

src/activities/Elsa.Activities.Startup/Activities/Startup.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+
+        public Startup()
+        {
+        }
+
+
+        protected override ActivityExecutionResult OnExecute(WorkflowExecutionContext context)
+        {
+            return Halt();
```
sfmskywalker: I'm not sure I understand how this works. It looks like that the activity will halt, but I don't see at which point the activity will be resumed. The `StartupHostedService` will trigger workflows starting with the `Startup` activity, which will cause the `OnExecute` method to be invoked. But then the workflow becomes suspended (the new term for "halted" in Eldsa 2.0).

Unless I'm missing something?

src/activities/Elsa.Activities.Startup/Elsa.Activities.Startup.csproj
Andale-online: Sorry, of course

src/activities/Elsa.Activities.Startup/Activities/Startup.cs
Andale-online: Dunno, copied from timer activities. It works :-D
Shall I do onexecute instead?

sfmskywalker: I see. So here's the thing: the Timer activities work because they are repeatedly being triggered. Internally, thee activities compare the current time with the first time they got triggered. If the difference exceeds a given threshold, the activity executes and continues the workflow.

The Startup background task on the other hand only triggers the Startup activity once. Since the Startup activity returns `HaltResult`, it will cause the workflow to be suspended. Only when you kill the ost and restart it, will the workflow continue. 

At least, this is the behavior I'd expect to see by looking at the code.

To summarize: please do return `Done()` from `OnExecute` :)

src/activities/Elsa.Activities.Startup/Elsa.Activities.Startup.csproj
Andale-online: Fixed

src/activities/Elsa.Activities.Startup/Activities/Startup.cs
Andale-online: Fixed :-)

Andale-online:
Give it a try

PR #211, opened 2019-12-12
Variable fixes

jamesdoran:
* SetVariables no longer re-wraps variable in another variable
* GetVariable<T> now uses Convert.ChangeType
Fixes issue where saving an Int32, serialising and deserialising, then accessing with GetVariable<int>(...) no longer causes type cast exception.
The saved value is a JValue which treats Int32 as Int64. Convert.ChangeType has a better chance of performing conversion without error.
Also returns default of T if the variable isn't found - **is this safe to do?**

### DIFF
--- a/src/core/Elsa.Abstractions/Models/Variables.cs
+++ b/src/core/Elsa.Abstractions/Models/Variables.cs
@@ -42,13 +42,18 @@ namespace Elsa.Models
             return this[name] = new Variable(value);
         }
 
+        public Variable SetVariable(string name, Variable variable)
+        {
+            return SetVariable(name, variable.Value);
+        }
+
         public void SetVariables(Variables variables) =>
             SetVariables((IEnumerable<KeyValuePair<string, Variable>>) variables);
 
         public void SetVariables(IEnumerable<KeyValuePair<string, Variable>> variables)
         {
             foreach (var variable in variables)
-                SetVariable(variable.Key, variable.Value.Value);
+                SetVariable(variable.Key, variable.Value);
         }
 
         public bool HasVariable(string name)
--- a/src/core/Elsa.Abstractions/Models/Variables.cs
+++ b/src/core/Elsa.Abstractions/Models/Variables.cs
@@ -33,8 +33,10 @@ namespace Elsa.Models
 
         public T GetVariable<T>(string name)
         {
-            object value = GetVariable(name) ?? default(T);
-            return (T)System.Convert.ChangeType(value, typeof(T));
+            object value = GetVariable(name);
+            return (value != default) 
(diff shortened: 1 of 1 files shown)


src/core/Elsa.Abstractions/Models/Variables.cs
```diff
@@ -44,7 +48,7 @@ public Variable SetVariable(string name, object value)
         public void SetVariables(IEnumerable<KeyValuePair<string, Variable>> variables)
         {
             foreach (var variable in variables)
-                SetVariable(variable.Key, variable.Value);
+                SetVariable(variable.Key, variable.Value.Value);
```
petedavis: I feel like the `SetVariable` method needs to have an overload that takes a Variable object. Then that can redirect the `variable.Value` to `SetVariable` 

That should ensure no one accidentally sets a variable to a variable like this.

Im sure the compiler would like to the more specific type if there were two overloads:
```csharp
public Variable SetVariable(string name, object value){ ... }

public Variable SetVariable(string name, Variable variable)
{
    return SetVariable(name, variable.Value);
}
```

petedavis: or maybe just add an `is Variable` test in the current `SetVariable` method?

Andale-online:
>   Also returns default of T if the variable isn't found - **is this safe to do?**

I think the way you did it it may fail... The Changetype may get a default(null) value.
How about defaulting in the return. Maybe then do the convert before the return.

src/core/Elsa.Abstractions/Models/Variables.cs
sfmskywalker: I think the overload version makes sense. The compiler will pick the right one when a Variable is specified.

jamesdoran:
> > Also returns default of T if the variable isn't found - **is this safe to do?**
> 
> I think the way you did it it may fail... The Changetype may get a default(null) value.
> How about defaulting in the return. Maybe then do the convert before the return.

Thanks, let me know if you think the latest update is better

src/core/Elsa.Abstractions/Models/Variables.cs
jamesdoran: Nice suggestion @petedavis. I've just pushed an update. Nice to know this won't accidentally happen again with this overload in there.

@sfmskywalker What are your thoughts on unit/integration tests? Do you think they'd provide value in this project? If so, do you have any preferences on frameworks/project structure etc?

sfmskywalker: @jamesdoran Yes, I was thinking not too long ago that we need both kinds of tests. I really like xUnit, but I'm open to suggestions.

PR #229, opened 2020-01-14
Add web farm hosting support

### LINKED ISSUE #224: Multi Node hosting environment support
When hosted in a multi-node environment (Kubernetes, web farms, etc.), the same workflow will run simultaneously on each instance. For example, when a background task triggers a TimerEvent, a workflow definition starting with a TimerEvent will execute on every node. This is undesirable.

Hosting workflows in a multi-node environment should allow for additional resources processing 
**different** workflows that need to be executed, and not the **same** workflow.

To solve for this, the first node to trigger workflows should first acquire a **lease**. When a lease is successfully acquired, any other node will fail to do so until the lease is released. The node that acquired the lease is now the **master** node for as long as it has the lease.

The **master** node will select any and all workflows that need to be executed, and **enqueue** them for execution. Each node in the cluster will consume this queue, allowing the work to be distributed equally across any number of nodes.

The above design requires the following services:

* Lease Provider
* Service Bus

These services will be provided as abstractions, and default implementations will be provided.

sfmskywalker:
This PR changes the way workflows are executed by sending them to a queue (in-memory by default), and adds support for distributed locking.

Currently only two locking providers are provided:

* In-memory provider for single-node hosting environments (and for local development)
* Sql Server locking provider, taking advantage of SQL Server's locking mechanism. Taken & adapted from Workflow Core.

Fixes #224


CleverAtBen:
Pleased to see this feature is in the next release.  
Is there a planned date for release as yet? 
Is there a neat solution to force all workflows to run on a single node in the mean time?

sfmskywalker:
Yes, I hope to release a beta version of Elsa 2.0 around April. Then depending on what issues pop up, a RC around May, and then the final bits around June.

Unfortunately I don't have a solution currently for running workflows on a single node without back-porting the workflow locking services from the 2.0 branch. If your workflows don't use any activities that are triggered from background tasks such as timers and messaging, but rather based on HTTP requests, then it's fine (since these workflows will intrinsically execute on one node). But I suspect most workflows will involve some (AMQP) messaging and timer activities.

CleverAtBen:
Thanks for the update - unfortunately all of my workflows have timers in them. 

Is there any way I can stop the background services running except on a node I target?

Ideally would want to start to be able to start workflows on any node - but could live with the background tasks running on a single node, until this feature became available.

sfmskywalker:
If there is a way, I'm unaware of it I'm afraid. The only way I can think of without down-scaling to just one node is to implement a lease mechanism. For example using [SqlServer](https://github.com/elsa-workflows/elsa-core/blob/feature/elsa-2.0/src/providers/Elsa.DistributedLocking.SqlServer/SqlLockProvider.cs). You could create a copy of the Timers Hosted Service that then uses this lock provider, as shown [here](https://github.com/elsa-workflows/elsa-core/blob/feature/elsa-2.0/src/activities/Elsa.Activities.Timers/HostedServices/TimersHostedService.cs).

Other than that, I don't know how else to do it.

PR #232, opened 2020-01-15
Fix an issue with some DB Context factories

sfmskywalker:
Fix an issue where some DB Context factories use the base ElsaContext type instead of the derived type, causing a "Unable to resolve service for type 'Microsoft.EntityFrameworkCore.DbContextOptions`1[]' while attempting to activate" error.

### DIFF
--- a/src/persistence/Elsa.Persistence.EntityFrameworkCore/DbContexts/DesignTimeDbContextFactoryBase.cs
+++ b/src/persistence/Elsa.Persistence.EntityFrameworkCore/DbContexts/DesignTimeDbContextFactoryBase.cs
@@ -1,0 +1,29 @@
+using System;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Design;
+
+namespace Elsa.Persistence.EntityFrameworkCore.DbContexts
+{
+    public abstract class DesignTimeDbContextFactoryBase<TDbContext> : IDesignTimeDbContextFactory<TDbContext> where TDbContext : DbContext
+    {
+        public TDbContext CreateDbContext(string[] args)
+        {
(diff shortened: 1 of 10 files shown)


src/persistence/Elsa.Persistence.EntityFrameworkCore/DbContexts/DesignTimeDbContextFactoryBase.cs
```diff
@@ -0,0 +1,29 @@
...
+
+            if(connectionString == null)
+            {
+                var providerName = typeof(TDbContext).Name.Replace("Context", "");
+                throw new InvalidOperationException($"Set the EF_CONNECTIONSTRING environment variable to a valid {providerName} connection string.");
+            }
+
+            optionsBuilder.UseSqlite(
+                connectionString,
+                x => x.MigrationsAssembly(migrationAssembly)
```
terry-delph: Because this only uses UseSqlite then no other migrations can be created for the other providers.
We could make this work by allowing the implementations to configure the options as an abstract method that must be implemented, that can then set the provider.

terry-delph: Ignore!
I can see you reverted back with the correct options builder applied in each factory in another PR :)

sfmskywalker: Exactly. I realized my mistake right after I merged it 😅

sfmskywalker: Thanks for looking though!

PR #283, opened 2020-04-03
Add Exception property to WorkflowFault

aburada:
Expose the underlying System.Exception that caused the Workflow to return a fault. This will allow callers to execute error handling based on the runtime Exception (if present) instead of error text.

Right now, Exception can be null. If Nullable Reference Types will be enabled in this project in the future, perhaps we can change this implementation a little and initialize the Exception object when only an errorMessage is specified.

### DIFF
--- a/src/core/Elsa.Abstractions/Models/WorkflowFault.cs
+++ b/src/core/Elsa.Abstractions/Models/WorkflowFault.cs
@@ -1,8 +1,11 @@
-﻿namespace Elsa.Models
+using System;
+
+namespace Elsa.Models
 {
     public class WorkflowFault
     {
         public string FaultedActivityId { get; set; }
         public string Message { get; set; }
+        public Exception Exception { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


aburada:
Hmm, I see your point. I haven't considered that aspect. My scenarios are mostly short running workflow that don't need to be serialized.  Most .NET Exceptions should be serializable, but it could pose a problem with custom Exceptions. I also see System.Exception also has an IDictionary Data property where anything could be added.

We could mark it as NonSerialized. Or create a wrapper that can serialize additional information from the exception (like Type, StackTrace?). Let me know what you think.

sfmskywalker:
@aburada I did a little research, and it appears to me that exception types are intended to be serializable. So even if some exception type is not-serializable (like custom ones), I would consider that an omission on the part of the exception implementation, which should then be fixed.

So let's go ahead with your PR as-is.

Thank you for the contribution, it is much appreciated.

syntaxonline-rewhitley:
Has anyone experienced issues the exception missing properties and not deserializing when throwing faults. Currently I cannot list workflows if any of my workflows contain faults

Issue #262, opened 2020-03-11
Update HTTP Activity with support for accepting Correlation ID as a header


sfmskywalker:
See also: #261

flew2bits:
Would this have to happen in TriggerRequestHandler? If so, it doesn't look like it would be too terribly difficult to implement.

However, what's the behavior in the case that `X-Correlation-Id` isn't a header? Do you include all matching workflow instances, or just those without a correlation id?

sfmskywalker:
@flew2bits 

> Would this have to happen in TriggerRequestHandler? If so, it doesn't look like it would be too terribly difficult to implement.
Indeed, this would have to happen there. I also don't suspect this would be too hard to do.

> However, what's the behavior in the case that X-Correlation-Id isn't a header? Do you include all matching workflow instances, or just those without a correlation id?
Good question, I hadn't thought of that. I think it might make sense to include just those without a correlation ID.

sfmskywalker:
Fixed in Elsa 2

PR #289, opened 2020-04-10
TriggerRequestHandler workflows to resume filtered by X-Correlation-ID header(s)

flew2bits:
The `TriggerRequestHandler.HandleRequestAsync` checks for the existence of the `X-Correlation-Id` header. If present, it will filter the workflows to resume by the correlation id(s) provided in the header(s). If the header is not present, no additional filtering will occur; i.e. all matching workflow instances will be included whether they have a correlation ID. This addresses issue #262.


sfmskywalker:
Thanks for this!

PR #330, opened 2020-06-14
Fix#228: Add Redis distributed lock provider 

anktsrkr:
This PR will add new feature of distributed lock provider for ELSA 2.0. User will be able to use redis as a lock provider. 

This implementation is based on this [link ](https://redis.io/topics/distlock)and uses popular open source project [RedLock.net](https://github.com/samcook/RedLock.net) as dependency. 

PR do not have any breaking changes however `IDistributedLockProvider` api now has more functions to setup initial state and dispose after use it. 

Also none of the `LockProvider `will implement `IDistributedLockProvider `api directly instead new `abstract DistributedLockProvider` is introduced to ease the custom lock provider implementation. 

User will be able to use this provider as - 
image


src/activities/Elsa.Activities.Timers/HostedServices/TimersHostedService.cs
```diff
@@ -29,6 +30,12 @@ public class TimersHostedService : BackgroundService
             this.logger = logger;
         }
 
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            await distributedLockProvider.SetupAsync(cancellationToken);
```
sfmskywalker, line 35: What would you say if we did the following:

1. Remove the `SetupAsync`.
1. Implement a `SetupRedisLockProvider: IHostedService` class that takes care of the setup autimatically.

Couple of advantages with that:

1. No need for each background task that needs distributred locking (like this `TimersHostedService`) to imperatively call `SetupAsync` - which might be inefficient if it is called multiple times, plus if someone implementing their own activity services forgets about this, it becomes inconsistent and prone to error.
1. No need anymore to "force" a `SetupAsync` on other distributed locking providers.

Conceptually, it would be akin to having a hosted servie that takes care of setting up database indices, running migrations etc - an async setup of the system before ASP.NET Core starts serving requests.

Thoughts?

anktsrkr: That makes sense! however i will have to look into it. 

If user wants to develop custom provider,how can ELSA enforces to implement that any LockProvider(in our cases) should also implement IHostedService? 

Also, If we do that connection will stay open a long as application is not stopped -- Happy Path 
In case of transient issue b/w ELSA and Redis, if connection broke, how it will up again? 

am i missing something?

anktsrkr: @sfmskywalker  I guess `SetupAsync `name is misleading, this method only connects to redis, you might need to take a look on that. 

[https://github.com/anktsrkr/elsa-core/blob/feature/elsa-2.0redisdistributedlockprovider/src/providers/Elsa.DistributedLocking.Redis/RedisLockProvider.cs#L73](https://github.com/anktsrkr/elsa-core/blob/feature/elsa-2.0redisdistributedlockprovider/src/providers/Elsa.DistributedLocking.Redis/RedisLockProvider.cs#L73)

My suggestion will be keep the method there but rename it. 

Thoughts?

sfmskywalker: @anktsrkr When a user wants to develop a custom provider that needs some hosted service to do some setup, I imagine this provider will ship with an extension method (on IServiceCollection), like the other modules do. In this extension method, all the necessary services would be registered: the provider as well as the hosted service. This is what I do for example in Elsa.Timers, where not only new activities are registered, but also a hosted service.

sfmskywalker: > Also, If we do that connection will stay open a long as application is not stopped -- Happy Path
In case of transient issue b/w ELSA and Redis, if connection broke, how it will up again?

This depends on how you set things up. For example, if all you need to do is setup something on the service one time, you can do this work in the Start method of the hosted service (just implement IHostedService directly instead of BackgroundService).

Re-establishing lost Redis connections could be done in the background (using a hosted service), or perhaps on the fly, when e.g. you try to read something from Redis. Depending on the library you're using, reconnecting should be handled for you according to [this post](https://stackoverflow.com/questions/24922782/recovering-from-redis-connectivity-loss).

anktsrkr: Indeed we do not need `SetupAsync`. Even we don't need `IHostedService` , StackExchange.Redis is capable enough to handle Transient error and ELSA will wait for 60 seconds to acquire it. In case of permanent server shutdown, ELSA will be unable to acquire lock and it will return false which is expected.  

I am still checking for possible enhancement, but hopefully by tommrow I can commit my changes.

sfmskywalker: Sounds good. I checked your changes and it looks great. Thanks!

sfmskywalker:
Hi @anktsrkr , this PR looks good to be merged, except that there are some merge conflicts. Could you try resolve them first? Let me know if you need help with that. Thanks!

anktsrkr:
@sfmskywalker  i'm not exactly sure why this merge conflict, I tried to remove all the projects from my repo and add it back but seems not able resolve it. 
I guess you need to do a small merge to resolve the conflict in `elsa-workflows:feature/elsa-2.0` directly as i do not have write access.

sfmskywalker:
@anktsrkr There are 3 small conflicts in Sample.sln. You can use the GitHub [Resolve conflicts button](https://github.com/elsa-workflows/elsa-core/pull/330/conflicts) you see below, or simply add the Elsa repo as a remote to the clone of your fork. You can then pull from that remote and merge the tip into your feature branch. You'll get a merge conflict which you should resolve and then push to your fork.

Give it a try and let me know.

anktsrkr:
Oops! Just tried today and did it work this time! Not sure why it did not work last time. Now it is ready to merge finally :)  @sfmskywalker

sfmskywalker:
Sorry @anktsrkr , I merged in two other PRs before this one, so now there are more merge conflicts 😅 But I'm guessing they will be easy for you to resolve as well, especially since they relate to your other PR.

anktsrkr:
Sure, I just resolved all conflicts, all seems passing, However do not merge now, I will confirm once I test it properly. @sfmskywalker

anktsrkr:
@sfmskywalker  it looks good to me after conflict resolution, it is ready to merge. Thanks!

Issue #228, opened 2020-01-14
Add Redis distributed lock provider


sfmskywalker:

anktsrkr:
PR for this enhancement #330 - Discussion in progress

anktsrkr:
PR #330 is approved, this needs to be closed once merged! @sfmskywalker

sfmskywalker:
Thanks for this awesome contribution!

PR #332, opened 2020-06-19
fix#227: Add Azure Blob Storage distributed lock provider

anktsrkr:
This PR will add new feature of distributed lock provider for ELSA 2.0. User will be able to use azure blob as a lock provider.

PR do not have any breaking changes however IDistributedLockProvider api now has more functions to setup initial state and dispose after use it.

Also none of the LockProvider will implement IDistributedLockProvider api directly instead new abstract DistributedLockProvider is introduced to ease the custom lock provider implementation.(This implementation may get change. Discussion still in progress)


anktsrkr:
@sfmskywalker please review once you have time, This PR contains updated version same as #228

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+        private Timer _renewTimer;
+
+        public AzureBlobLockProvider(string connectionString, TimeSpan leaseTime, TimeSpan renewInterval,
+                                        ILogger<AzureBlobLockProvider> logger)
+        {
+            _logger = logger;
+            _connectionString = connectionString;
+            if (leaseTime > TimeSpan.FromSeconds(60) || leaseTime < TimeSpan.FromSeconds(15))
+            {
+                _logger.LogInformation($"Leasetime must be between 15 Seconds and 60 seconds, Found {leaseTime.TotalSeconds} seconds. " +
```
sfmskywalker, line 36: When creating log strings, you shouldn't use `$` for string interpolation. Instead, you should allow the logging system to perform the string interpolation itself. For example:

`_logger.LogInformation("Leasetime must be between 15 Seconds and 60 seconds, found {LeaseTimeTotalSeconds} seconds. Setting default value of 60 seconds", LeaseTime.TotalSeconds);`

See also: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#log-message-template

Please update the other logging statements as well. Thank you!

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+                }
+                finally
+                {
+                    _leaseSemaphore.Set();
+                    _renewTimer?.Dispose();
+                }
+            }
+        }
+
+        private CloudBlobContainer CloudBlobContainer
```
sfmskywalker, line 140: Any chance we can make this an async method so we can leverage the non-blocking methods of the blob container?

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+        private CloudBlobContainer _cloudBlobContainer;
+        private readonly TimeSpan _renewInterval;
+        private Timer _renewTimer;
+
+        public AzureBlobLockProvider(string connectionString, TimeSpan leaseTime, TimeSpan renewInterval,
+                                        ILogger<AzureBlobLockProvider> logger)
+        {
+            _logger = logger;
+            _connectionString = connectionString;
+            if (leaseTime > TimeSpan.FromSeconds(60) || leaseTime < TimeSpan.FromSeconds(15))
```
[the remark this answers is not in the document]
anktsrkr: As per this [documentation](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob), the lock duration can be 15 to 60 seconds, or can be infinite and hence the Guard. 
If it is taking more than 60 seconds to process the next block it will automatically try to re-acquire a lock.

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
anktsrkr: Hmm, but dont think it will give you much benefit, because  -There are only two non-blocking methods `_cloudBlobContainer.Exists()` and `_cloudBlobContainer.CreateIfNotExists` and theoretically it is very rare that code execution will flow till that, so what I mean is it is not the HotPath.

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
anktsrkr: Updated and checked-in.

src/providers/Elsa.DistributedLocking.AzureBlob/AzureBlobLockProvider.cs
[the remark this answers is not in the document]
sfmskywalker: I see. I didn't know. Would you mind adding a brief comment referencing this documentation? And also move the literal values into a constant field? Thanks!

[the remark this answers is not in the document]
anktsrkr: As suggested I have added the reference link and moved literal values to constant field, Now this PR should be ready to merge! @sfmskywalker

PR #413, opened 2020-10-21
Minor code cleanup

hankovich:
As I mentioned I found several places that could be improved. These are mostly very minor / picky cleanups that make the code more readable.

This PR fixes them. Note that I did not touch Core projects in order not to make this PR huge.


src/activities/Elsa.Activities.Dropbox/Handlers/DropboxApiHandler.cs
```diff
@@ -1,18 +0,0 @@
-using System.Net.Http;
```
hankovich, line 1: I'm not sure I can delete it since it's public, but it looks like it's not used to have no value

src/scripting/Elsa.Scripting.JavaScript/Converters/TruncatingNumberJsonConverter.cs
```diff
@@ -27,7 +23,7 @@ private static bool IsWholeValue(object value)
 
             if (value is float || value is double)
             {
-                var doubleValue = (double) value;
```
hankovich, line 30: Note that
```
object value = (float) 42;
var number = (double) value;
```

will throw an InvalidCastException, but Convert.ToDouble will not

sfmskywalker (approved):
Very nice improvements, thanks! I wonder if you could apply these also onto the `feature/elsa-2.0` branch?

src/scripting/Elsa.Scripting.JavaScript/Converters/TruncatingNumberJsonConverter.cs
sfmskywalker: Good point.

src/activities/Elsa.Activities.Dropbox/Handlers/DropboxApiHandler.cs
sfmskywalker: I don't think you can delete that namespace import, since `HttpRequestMessage` comes from that namespace.

PeterLaComb-Availity (approved):

src/activities/Elsa.Activities.Dropbox/Handlers/DropboxApiHandler.cs
hankovich: My comment was related to the entire file :)

hankovich:
Ah, my bad, I was not aware of the `feature/elsa-2.0` branch.
Do you want me to open another PR (to the `feature/elsa-2.0`) instead or it should be an additional one (in addition to the current PR)?

Should I just cherry-pick the commit or maybe just apply the same changes to the code in the `feature/elsa-2.0`?

sfmskywalker:
Not at all, I should have mentioned it.
Please feel free to open another PR in addition to this one - the bug fixes are still relevant for V1.

Cherry-picking is fine, whatever is easier for you.

src/activities/Elsa.Activities.Dropbox/Handlers/DropboxApiHandler.cs
sfmskywalker: Ah. Yes, this entire file can be removed (and `ServiceCollectionExtensions` needs to be updated accordingly).

hankovich:
Ah got it, thanks.
I think the current PR should be merged. I'll open another one, will apply the same cleanups and probably will find a new ones specific to v2. Does it make sense?

PR #421, opened 2020-10-23
Minor cleanup of elsa 2.0

hankovich:
I am opening this PR according to https://github.com/elsa-workflows/elsa-core/pull/413#pullrequestreview-513985910
It also does not contain changes for the projects in the `core` folder so as not to make this PR huge.

--------------------

Did you get rid of it Reflection Activity btw? I liked it and even had a plan to improve it.

--------------------

I also want to ask a question (probably it's not the best place, but anyway).
What is the benefit of taking `Action<OptionsBuilder<TOptions>>`?  like

```
public static IServiceCollection AddHttpServices(this IServiceCollection services, Action<OptionsBuilder<HttpActivityOptions>> options)
{
    options.Invoke(services.AddOptions<HttpActivityOptions>());
    // other stuff
}
```

and a caller should do
```
services.AddHttpActivities(options => options.Bind(elsaSection.GetSection("Http")))
```

Why is it better than

```
public static IServiceCollection AddHttpServices(this IServiceCollection services, Action<HttpActivityOptions> options)
{
    services.Configure(options);
    // other stuff
}
```

and 

```
services.AddHttpActivities(elsaSection.GetSection("Http").Bind)
```

?

I think `Action<OptionsBuilder<TOptions>>` adds extra complexity, but I don't see any value in it. Could you please share your opinion on this (I'm probably missing something important)?

### DIFF
--- a/src/activities/Elsa.Activities.Console/Activities/ReadLine/ReadLine.cs
+++ b/src/activities/Elsa.Activities.Console/Activities/ReadLine/ReadLine.cs
@@ -1,4 +1,4 @@
-﻿using System.IO;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Elsa.ActivityResults;
@@ -16,15 +16,13 @@ namespace Elsa.Activities.Console
         Category = "Console",
         Description = "Read text from standard in.",
         Icon = "fas fa-terminal",
-        RuntimeDescription =
-            "a => !!a.state.variableName ? `Read text from standard in and store into <strong>${ a.state.variableName }</strong>.` : 'Read text from standard in.'",
         Outcomes = new[] { OutcomeNames.Done }
     )]
     public class ReadLine : Activity
     {
         private readonly TextReader? _input;
 
-        public ReadLine()
+        public ReadLine() : this(System.Console.In)
         {
         }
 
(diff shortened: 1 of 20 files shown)


sfmskywalker (commented):
Nice cleanup, thanks!

sfmskywalker (approved):

sfmskywalker:
> Did you get rid of it Reflection Activity btw? I liked it and even had a plan to improve it.

Indeed I did, for a couple of reasons:

1. It seemed a little bit too technical / low-level.
2. I couldn't come up with a real world use-case.

But perhaps you can change my mind. What improvements did you have in mind?

> What is the benefit of taking Action<OptionsBuilder<TOptions>>

Zero benefits, just downsides as you mentioned. The reason it's there because at the time I thought this was the best way to do it. Please feel free to change it for Elsa 2, it would be greatly appreciated! :)

hankovich:
>Please feel free to change it for Elsa 2

Cool, will do!

>What improvements did you have in mind?

That was more like an idea. And now I'm sure you would not like it 😊
I think a lot of methods nowadays are async and basically I wanted to add an await if method returned awaitable type. So it's even more technical point than reflection 😊

PR #423, opened 2020-10-26
CurrentScope is incorrect

hyzx86:
In the first iteration, a new scope will be created and pushed onto the top of the stack. CurrentScope is now a scope at the top of the stack, but in the second iteration, the newly added scope will change to the bottom. CurrentScope becomes the original Scope at the bottom of the stack.
My code only solved my problem, I didn't find the reason for stack inversion.


sfmskywalker:
I'm considering dropping the multiple scope thing entirely. It's buggy and incomplete, and not really used AFAIK. Would you be OK with making this change (deleting `Workflow.Scopes` and updating `Workflow.CurrentScope` to return a single instance)? If not I will do it soon, just let me know. Thanks!

hyzx86:
Yes, I also think Scope is useless

PR #457, opened 2020-11-12
Enhancements for using azure service bus with MassTransit activities

cvijayak:
Exposed the MassTransit setup to the caller. So that caller can configure RabbitMq or AzureServiceBus for their needs.

With this change, caller can use AzureServiceBus with MassTransit activities.


sfmskywalker (approved):
Amazing contribution. Thank you!

PR #472, opened 2020-11-24
DocumentDB persistence layer unable to deserialize the Exception class instances

cvijayak:
The **JsonSerializerSettings** which is attached with **DocumentClient** do not have proper converters for **Exception** that leads to the JSON serialization error.

The changes to fix the issue available as part of the PR. Please have a look and approve the same.


sfmskywalker (approved):

sfmskywalker:
Nice find!

PR #494, opened 2020-12-05
Added partition key for heavy read queries and also to avoid cross partition queries 

cvijayak:
Since dashboard get crashed or hanged for longtime due to cross partition queries. I just add a partition key to resolve the same.

After adding the partition key, I tested the changes with CosmoDB. Now I am not seeing any crash or hang. Dashboard loads without delay. 

Without the fix, I have to wait for longtime for dashboard to load. Most of time, it will crash or hang. Now it is loading quickly without much delay.


sfmskywalker (approved):

sfmskywalker:
@cvijayak Your other PR was just merged, but now this PR has a small merge conflict. I'll merge this one in too once you updated it. Thanks!

PR #516, opened 2020-12-14
Separate quartz

Rafael94:
Add Hangfire support
Add CancelTimer Activity


Rafael94:
@sfmskywalker 
I think I have found a bug or am missing something. Namely, there are repetitive triggers (cron, timer) on timers. But these are not deleted or I can not find the place.

`WithSimpleSchedule(x => x.WithInterval(interval.ToTimeSpan()).RepeatForever()`

`WithCronSchedule(cronExpression)`

My suggestion:
- Within the job, the status is checked, when the workflow instance has run through, the background task is deleted.
- Add new activity: `UnscheduleJob`

Rafael94:
There is still a problem with the Recurring Job. The name can be a maximum of 20 characters long, which is much too short.

Either open an issue at Hangfire or Recuring Jobs as scheduled jobs. I find the latter option better because it makes it easier to remove workflow instances that no longer exist or have been terminated.

Rafael94:
Hangfire does not set up standing orders, only one-time orders. In the job it is checked whether it is a standing order (timer, cron) and then a one-time order is created again. This way I have no problem with the ID length.

sfmskywalker:
I think you're right; for blocking activities such as `Timer` and `Cron`, the trigger should delete the background task when done. I'll look into that.

As for the `UnscheduleJob`, let's not do that; the user creating the workflow should not have to worry about unscheduling a job that is triggered by a Timer/Cron activity.

Rafael94:
I did it with `UnscheduleJob` for a different reason. 

It could be that the activity needs to be cancelled before running. For example. 

For a ticket, the owner should be changed once within 5 days. When the owner has been changed, I want to cancel the timer (5 days) so that no escalation email is sent.

sfmskywalker:
Oh I see. But then what should happen with the workflow's execution when the timer is cancelled? Should it keep being blocked on the timer activity? Obviously it shouldn't continue execution.

If, on the other hand, you **forked** execution into e.g. 2 branches, where **branch 1** is blocked on the timer, and **branch 2** is blocked on the "owner changed" event, then you wouldn't need to manually un-schedule the timer activity. Instead, the `Join` activity, while clearing incoming blocking activities (the timer on branch 1), would then also un-schedule the task (which I just realized as I'm writing this, does not currently happen, but should be fixed).

Rafael94:
> Oh I see. But then what should happen with the workflow's execution when the timer is cancelled? Should it keep being blocked on the timer activity? Obviously it shouldn't continue execution.
> 
> If, on the other hand, you **forked** execution into e.g. 2 branches, where **branch 1** is blocked on the timer, and **branch 2** is blocked on the "owner changed" event, then you wouldn't need to manually un-schedule the timer activity. Instead, the `Join` activity, while clearing incoming blocking activities (the timer on branch 1), would then also un-schedule the task (which I just realized as I'm writing this, does not currently happen, but should be fixed).

I am not sure if there is always a join. Only as an example

Added -> Timer 5 Days -> EMail
            -> Owner Changed -> Cancel Timer
            -> State Changed -> EMail ->Finish

If the email is sent, no join is needed. If only certain people are using the Workflow Designer, you can still require it, but as soon as end users start using the Designer, it becomes difficult to require them to use a join for every fork.

I also just noticed that with CancelTimer (UnscheduleJob) you need to remove the activity from BlockingActivities and ScheduledActivities as well, if I'm not mistaken.

I'm not that familiar with Elsa in depth, unfortunately. Could you please take a look at it. I would want to help but have no approach.

sfmskywalker:
Ok let me make sure I got this straight.

In your example workflow, you want the workflow to finish from either branch. Since the workflow would remain blocked on the timer event, you are proposing a `UnscheduleJob` activity to cancel the timer and allow the workflow to complete.

Although this would work for this particular case, what would you propose we do in case the workflow wasn't blocked on a timer, but a `SignalReceived` event? Or an `HttpRequestReceived` event. Or perhaps even a `ServiceBusMessageReceived`?

My point of course being: the `UnscheduleJob` is too specific; we need an activity that can unblock any blocking activity. This is the role of the `Join` activity; it would not only clear the Timer blocking activity, but also any other type.

Similarly, the `Finish` activity fulfills the same role: to clear out any blocking activity (which means this could be used instead of the Join activity like in your example).

That being said, we do currently have the issue you identified, namely the fact that none of the triggers/jobs are removed from Quartz (and possible from Hangfire, although I haven't looked at those details yet).

This is something I'll fix.

src/activities/Elsa.Activities.Timers.Hangfire/Elsa.Activities.Timers.Hangfire.csproj
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <PackageVersion>1.0.0</PackageVersion>
+        <LangVersion>latest</LangVersion>
+        <Authors>Elsa Contributors</Authors>
```
sfmskywalker, line 7: Perhaps update this with common.props?

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,19 @@
+namespace System
+{
+    public static class TimeSpanExtensions
+    {
+        public static string ToCronExpression(this TimeSpan timeSpan)
```
sfmskywalker, line 5: Is there a specific reason not to do this for `Duration`?

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimersOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Add Hangfire for background processing
+        /// </summary>
+        /// <remarks>
+        /// Only if Hangfire is not already registered in DI
+        /// </remarks>
+        /// <param name="timersOptions"></param>
+        /// <param name="timersOptions">Hangfire settings</param>
+        public static void UseHangfireProvider(this TimersOptions timersOptions, Action<IGlobalConfiguration> configure)
```
sfmskywalker, line 34: Perhaps merge this method with `UseHangfireProviderCore` and make `configure` nullable and optional? At the very least, both methods should have the same name.

src/activities/Elsa.Activities.Timers.Hangfire/Jobs/RunHangfireWorkflowJob.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (data.WorkflowInstanceId == null)
+            {
+                if (workflowBlueprint.IsSingleton == false || await _workflowInstanceManager.GetWorkflowIsAlreadyExecutingAsync(data.TenantId, data.WorkflowDefinitionId) == false)
+                {
+                    await _workflowRunner.RunWorkflowAsync(workflowBlueprint, data.ActivityId);
+                }
+            }
+            else
+            {
+                for (var i = 0; i < 3 && workflowInstance == null; i++)
```
sfmskywalker, line 52: Perhaps extract this retry mechanism to a private method? Makes the `ExecuteAsync` a little bit easier to read. It's OK if you don't; I can do it later.

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimersOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Elsa.Activities.Timers.Hangfire.Extensions
+{
+    public static class TimersOptionsExtensions
+    {
+        /// <summary>
+        /// Add Hangfire for background processing
+        /// </summary>
+        /// <param name="timersOptions"></param>
+        public static void UseHangfireProviderCore(this TimersOptions timersOptions)
```
sfmskywalker, line 19: Unless there's a reason to use "Core" in the name that I'm unaware of, I think we can just go with `UseHangfireProvider` or even `UseHangfire` (since it's an extension on `TimersOptions`, i.e. won't conflict with any Hangfire extensions).

src/activities/Elsa.Activities.Timers/Options/TimersOptions.cs
```diff
@@ -1,11 +1,16 @@
...
 {
     public class TimersOptions
     {
-        public TimersOptions()
+        public IServiceCollection Services { get; }
+
+        public TimersOptions(IServiceCollection services)
         {
+            Services = services;
             SweepInterval = Duration.FromMinutes(1);
```
sfmskywalker, line 14: This reminds me: the `SweepInterval` isn't used anymore. It was used before I added Quartz as the background task scheduler.

sfmskywalker (approved):
Awesome work, thank you!

Rafael94:
> Ok, let me make sure I got this straight.
> 
> In your example workflow, you want the workflow to finish from either branch. Since the workflow would remain blocked on the timer event, you are proposing an `UnscheduleJob` or `CancelTimer` activity to cancel the timer and allow the workflow to complete.
> 
> Although this would work for this particular case, what would you propose we do in case the workflow wasn't blocked on a timer, but a `SignalReceived` event? Or an `HttpRequestReceived` event. Or perhaps even a `ServiceBusMessageReceived`?
> 
> My point of course being: the `UnscheduleJob` / 'CancelTimer' activity is too specific; we need an activity that can unblock any blocking activity. This is the role of the `Join` activity; it would not only clear the Timer blocking activity, but also any other type.
> 
> The `Finish` activity fulfills the same role: to clear out any blocking activity (which means this could be used instead of the Join activity like in your example).
> 
> That being said, we do currently have the issue you identified, namely the fact that none of the triggers/jobs are removed from Quartz (and possible from Hangfire, although I haven't looked at those details yet).
> 
> This is something I'll fix.

You're right. I did not think of the other trigger, but only in my case with the timers.

Where I'm not sure though is if there is always a join. What if the branches have different ends and never join? Actually you would need a combination of switch and fork. As I understand it, the switch only reacts to values, but not to triggers. 

So if a trigger in the switch triggers activity, all other triggers would have to be removed.


And about the timers and join, I have another question. If the timer triggers and then ends in a join, can I reset it without any problems?

For example. The ticket reaches the follow up date and an email is sent.The employee then changes the follow up date and the timer should be reset so that an email is sent again when the follow up date is reached.The employee then changes the follow up date and the timer should be reset so that an email is sent again when the follow up date is reached.

src/activities/Elsa.Activities.Timers.Hangfire/Elsa.Activities.Timers.Hangfire.csproj
Rafael94: Yes. The base was at the level before common.props

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimeSpanExtensions.cs
Rafael94: Has no reason. It was created during the development. I had it differently in the beginning.

src/activities/Elsa.Activities.Timers.Hangfire/Jobs/RunHangfireWorkflowJob.cs
Rafael94: Can do. I wasn't sure you wanted it that way. Because with Quartz it's an infinite loop if the instance doesn't exist.

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimersOptionsExtensions.cs
Rafael94: I did it this way in case someone would prefer to add Hangfire directly to their application instead of through Elsa. I for example use it directly in my web application and configure it in the startup.

I didn't necessarily want to add it again via Elsa. Or let's say if `configure` is null, then only the services are added.

sfmskywalker:
> What if the branches have different ends and never join?

Then the branches blocked on a given trigger remain blocked. If the workflow developer wants to ensure this doesn't happen, they need to ensure to either join the forked execution, or add the Finish activity which will then clean up all blocking activities (and todo: also clear scheduled jobs). So it doesn't really matter whether the developer does this using a ClearTimer, Join or Finish activity. 

The ClearTimer however has two disadvantages: it only works for Timers, and you have to explicitly reference the timer by name that you want to clear. On the other hand, it might be a nice activity to have in case you want to clear a timer from a completely different branch of execution. Perhaps this is reason enough to keep it (but we should still make sure Join and Finish clear up scheduled jobs as well, which is on my list of things to do).

> And about the timers and join, I have another question. If the timer triggers and then ends in a join, can I reset it without any problems?

Yes, this will not be a problem. Right now, the Quartz trigger will simply be removed and then added again.
The only thing still missing is to delete the trigger immediately after the blocking timer has executed (right now it keeps "ticking", although it won't invoke the workflow anymore since the timer won't be part of the BlockingActivities list anymore).

src/activities/Elsa.Activities.Timers.Hangfire/Jobs/RunHangfireWorkflowJob.cs
sfmskywalker: > Because with Quartz it's an infinite loop if the instance doesn't exist.

Indeed. Which is why I like your solution here 👍

src/activities/Elsa.Activities.Timers.Hangfire/Extensions/TimersOptionsExtensions.cs
sfmskywalker: I see. That makes sense. Leave it as-is.

Rafael94:
I have adjusted things now. I just can't compile it on my end because I don't have the common.props locally anymore. I also don't want to add this now before more conflicts arise.

sfmskywalker:
Sure, no problem. We can fix it after merging in.

Rafael94:
> Sure, no problem. We can fix it after merging in.

Ok. If there are no more suggestions for changes, I'm done.

sfmskywalker:
Awesome!

sfmskywalker:
Only thing left to do is resolve the merge conflicts.

sfmskywalker:
Looks like I managed to push the conflict resolution myself.

Rafael94:
> Looks like I managed to push the conflict resolution myself.

Thank you. I have hardly had time in the last few days. I hope it will be better now.

sfmskywalker:
No worries!

PR #526, opened 2020-12-16
Adding Activities, Workflows, TriggerProviders and WorkflowContextProviders by Reflection

Rafael94:
The new enhancements eliminate the need to call the appropriate function for each Activity, Workflow, TriggerProvider and WorkflowContextProvider.

I have often forgotten to call `AddActivity<T>()`. It's just a small thing, but it speeds up the development.


Rafael94:
Something just went wrong. The pull request should actually only contain this commit: https://github.com/elsa-workflows/elsa-core/pull/526/commits/598245cfb4d2c6065063d49332782fae24b132a7. @sfmskywalker any idea how to fix it?

src/activities/Elsa.Activities.AzureServiceBus/FodyWeavers.xsd
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sfmskywalker, line 1: If these files are generated, can't we exclude the from git? Assuming they'll be generated upon build.

src/activities/Elsa.Activities.Email/Activities/SendEmail/SendEmail.cs
```diff
@@ -23,8 +25,14 @@ public SendEmail(ISmtpService smtpService)
         [ActivityProperty(Hint = "The sender's email address.")]
         public string From { get; set; }
 
-        [ActivityProperty(Hint = "The recipient's email address.")]
-        public string To { get; set; }
+        [ActivityProperty(Hint = "The recipients email addresses.")]
```
sfmskywalker, line 28: Didn't we already merge the send email PR?

src/activities/Elsa.Activities.Email/Activities/SendEmail/SendEmailExtensions.cs
```diff
@@ -7,17 +8,38 @@ namespace Elsa.Activities.Email
 {
     public static class SendEmailExtensions
     {
+        #region sendner
```
sfmskywalker, line 11: Small typo (should be "sender")

sfmskywalker, line 11: I prefer if we'd use single line comments to group these methods instead of #regions.

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,6 +10,45 @@ public static class ServiceCollectionExtensions
...
         public static IServiceCollection AddWorkflowProvider<T>(this IServiceCollection services) where T : class, IWorkflowProvider => services.AddTransient<IWorkflowProvider, T>();
         public static IServiceCollection AddTriggerProvider<T>(this IServiceCollection services) where T : class, ITriggerProvider => services.AddTransient<ITriggerProvider, T>();
+
+        /// <summary>
+        /// Add all TriggerProviders (<see cref="ITriggerProvider"/>) that are in the assembly
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="assembly"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddTriggerProvider(this IServiceCollection services, Assembly assembly)
```
sfmskywalker, line 20: Since Elsa already depends on Scrutor, can we use that package to register services instead?

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,6 +10,45 @@ public static class ServiceCollectionExtensions
...
+        {
+            var ITriggerProviderType = typeof(ITriggerProvider);
```
sfmskywalker, line 22: Rename to `triggerProviderType`

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,6 +10,45 @@ public static class ServiceCollectionExtensions
...
+
+        /// <summary>
+        /// Add all WorkflowContextProviders (<see cref="IWorkflowContextProvider"/>) that are in the assembly
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="assembly"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddWorkflowContextProvider(this IServiceCollection services, Assembly assembly)
+        {
+            var iWorkflowContextProviderType = typeof(IWorkflowContextProvider);
```
sfmskywalker, line 43: Let's drop the `i` (and start with lowercase `w`)

sfmskywalker:
Did you update your fork with the latest version before creating this PR?

src/activities/Elsa.Activities.AzureServiceBus/FodyWeavers.xsd
Rafael94: I didn't assume that these would be constantly recreated.

src/activities/Elsa.Activities.Email/Activities/SendEmail/SendEmail.cs
Rafael94: Yes, that should also not be in the pull request

src/activities/Elsa.Activities.Email/Activities/SendEmail/SendEmailExtensions.cs
Rafael94: Okay, it's not a problem here either. But sections that contain many more lines, this method is better. Then you can collapse the individual sections.

Rafael94:
I updated my fork and I think I did something wrong.

Pull request Elsa-> to my fork and then rebase, instead of merge.

I'm thinking of finishing Hangfie/Quartz and then deleting the fork and recreating it. There were not many changes for this pull-request.

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
Rafael94: I would like to avoid external libraries if possible. Especially if only one line is needed. I already had the problem that with external libraries both had the same dependency, but different versions. This caused conflicts and cost me days.

In addition, Scrutor is not needed. It is only used in one place AutoMapperServiceCollectionExtensions. But it makes no sense here because AutoMapper already includes a scanner https://docs.automapper.org/en/stable/Configuration.html . I would prefer the scanner from Automapper.

src/activities/Elsa.Activities.AzureServiceBus/FodyWeavers.xsd
sfmskywalker: Do you know at which point they are created? Looking at this [.gitignore template](https://raw.githubusercontent.com/github/gitignore/master/VisualStudio.gitignore), I take it is fine to exclude them.

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
sfmskywalker: Actually, Scrutor is also used by Elsa.Core for decorating services, so we might as well take advantage of it. If it causes versioning issues with consuming applications and libraries, we can always consider making it a private dependency. If it really were just a couple of lines of code, it would be OK to remove Scrutor as a dependency, but I think it;s a little bit more than that, in which case I prefer to maintain as little code as I can.

src/activities/Elsa.Activities.Email/Activities/SendEmail/SendEmailExtensions.cs
sfmskywalker: I think that's debatable, but let's cross that bridge when we get there :)

sfmskywalker:
> I'm thinking of finishing Hangfie/Quartz and then deleting the fork and recreating it. There were not many changes for this pull-request.

OK

src/activities/Elsa.Activities.AzureServiceBus/FodyWeavers.xsd
Rafael94: I also think it's better. I didn't think of it at first. 

The .gitignore contains more entries than the .gitignore from this repository. Maybe you should take the entries to solve future problems.

If you could please take a look at my other pull request, then I could recreate the fork and make the change this week.

src/core/Elsa.Abstractions/Extensions/ServiceCollectionExtensions.cs
Rafael94: I could take a look at why it is needed. Possibly can be replaced by a small change. If it is more, I agree with you.

sfmskywalker: Sounds good, I'll leave this to you until you let me know otherwise. Thank you!

src/activities/Elsa.Activities.AzureServiceBus/FodyWeavers.xsd
sfmskywalker: That's fine, the visual studio ignore is there as a convenience, but as you see it also contains entries of things that are very unlikely to be ever used in this project, such as xamarin and Visual C++. It's fine to add new entries as needed.

I'll check out your other PR!

Issue #468, opened 2020-11-19
Elsa Ideas


sfmskywalker:
This issue is a collection of pins for ideas to get feedback on, refine, prioritize & plan.

- [ ] Expression editor with intellisense
- [ ] Validation rules (e.g. mandatory properties) (see also #452)
- [ ] Test workflows from the designer using test inputs (including breakpoints)
- [ ] Define automated workflow tests from the dashboard & run them
- [ ] Workflow ACL rules: workflows can only be edited if one has the required permissions. Similar to Orchard Core content permissions.
- [ ] Restrict activities that are available from the designer. Perhaps configurable per workflow, or associated with a "whitelist profile" that itself has a set of required permissions a user must have (see also #315)
- [ ] See detailed diagnostic data during workflow development/testing (see also #444).
- [ ] Have telemetry data output during production workflow execution (e.g. to Application Insights).
- [ ] Modular Dashboard application. Could use (parts of) Blazing Orchard for dynamic component composition or even Oqtane. Allows developers to enrich the dashboard unobtrusively with application specific functionality.
- [ ] Pluggable workflow designer: configure per workflow what editor to use. Allows for optimized editing experience depending on the type of workflow you're editing. Alternatively, or perhaps ultimately, there would be a single designer that exposes knobs and bobs to configure it (also perhaps on a per workflow basis)
- [ ] Re-hostable workflow designer. Right now the designer is implemented using Blazor. Should be able to rehost it in e.g. Orchard Core, ABP Framework, Oqtane, etc. This allows develoeprs to embed the designer into their own backoffice applications.
- [ ] Conductor: a set of activities that allow for simple orchestration of tasks to be executed, similar to the way Netflix Conductor works. This allows Elsa to be used easily with any other type of application through a messaging transport system (HTTP, gRPC, AMQP) (#466)
- [ ] Workflow Templates: a way to design "master" workflows with "seams" into which "child" workflows can be inserted (#463)
- [ ] Workflows DSL / Code editor: the designer should have a toggle view that displays a code editor containing the DSL for the workflow being edited (#462)
[...]

PR #536, opened 2020-12-23
Redesign activities and workflows management

Rafael94:
I took a point from the list #468.

Activities are now stored in `ElsaOptions`. I renamed the old class `ElsaOptions` to `ElsaConfigurationsOptions`. Among other things because `public IServiceCollection Services { get; }` is not needed at runtime.

Also, activities and workflows can be added via Assembly. It is no longer necessary to add each activity and workflow individually.

### DIFF
--- a/src/core/Elsa.Core/ElsaConfigurationsOptions.cs
+++ b/src/core/Elsa.Core/ElsaConfigurationsOptions.cs
@@ -17,9 +17,9 @@ using Storage.Net.Blobs;
 
 namespace Elsa
 {
-    public class ElsaOptions
+    public class ElsaConfigurationsOptions
     {
-        public ElsaOptions(IServiceCollection services)
+        public ElsaConfigurationsOptions(IServiceCollection services)
         {
             Services = services;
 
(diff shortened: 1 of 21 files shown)


sfmskywalker:
Awesome, I will have a look.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
```diff
@@ -59,9 +62,28 @@ public static class ElsaServiceCollectionExtensions
         }
 
         public static IServiceCollection AddActivity<T>(this IServiceCollection services) where T : class, IActivity =>
-            services
-                .AddTransient<T>()
-                .AddTransient<IActivity>(sp => sp.GetRequiredService<T>());
+            services.AddActivity(typeof(T));
```
sfmskywalker, line 65: I'm wondering if we can't skip this registration altogether. After all, we can still instantiate activity types using `ActivatorUtilities` using `IServiceProvider` to resolve an activity's constructor arguments. Then we could even potentially register additional activity types at runtime.

src/core/Elsa.Core/ElsaConfigurationsOptions.cs
```diff
@@ -17,9 +17,9 @@
 
 namespace Elsa
 {
-    public class ElsaOptions
+    public class ElsaConfigurationsOptions
```
sfmskywalker, line 20: Would `ElsaConfiguration` be a better name? Combining "configuration" and "options" seems a bit much.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
```diff
@@ -59,9 +62,28 @@ public static class ElsaServiceCollectionExtensions
...
+
+        public static IServiceCollection AddActivity(this IServiceCollection services, Type activityType) =>
+           services.Configure<ElsaOptions>(options => options.RegisterActivity(activityType));
+
+        /// <summary>
+        /// Add all activities (<see cref="IActivity"/>) that are in the assembly
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="assembly"></param>
+        /// <returns></returns>
```
sfmskywalker, line 75: You can remove the empty XML comments; they are just noise.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
```diff
@@ -75,9 +97,29 @@ public static class ElsaServiceCollectionExtensions
 
         public static IServiceCollection StartWorkflow<T>(this IServiceCollection services) where T : class, IWorkflow => services.AddHostedService<StartWorkflow<T>>();
 
-        private static IServiceCollection AddMediatR(this ElsaOptions options) => options.Services.AddMediatR(mediatr => mediatr.AsScoped(), typeof(IActivity));
+        /// <summary>
+        /// Add all workflows (<see cref="IWorkflow"/>) that are in the assembly and have no constructor with parameters
```
sfmskywalker, line 101: We have to be able to add `IWorkflow` implementations that do receive constructor arguments.

Rafael94: I'd like it that way too, but it doesn't work for one example. In one example Instant is passed as a parameter. This is not registered in the DI, which gives an error.

If you don't mind, I would change the example to also add workflows with parameters.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
Rafael94: It is already possible now. I have only left the registration type as it is. The activities are no longer registered to DI, but to `ElsaOptions`. `ElsaOptions` is registered as an option and can be accessed anywhere. Thus, new activities can be added at any time.

Rafael94:
Is `AddSingleton<IWorkflow>(sp => sp.GetRequiredService<T>())` actually needed? Because I didn't find any places.

sfmskywalker:
It is needed by the `ProgrammaticWorkflowProvider` workflow provider; it injects an `IEnumerable<IWorkflow>`. But with your work done here, this is no longer necessary since the `ProgrammaticWorkflowProvider` can now simply get the types from `ElsaOptions`.

Rafael94:
> ProgrammaticWorkflowProvider

Now I have seen it. Didn't know that the DI also works with lists. I'll just do that then.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
sfmskywalker: In that case, the user should provide their own workflow factory method (as is also shown in that example).
But I see why it wouldn't work with `IWorkflow` types now being part of `ElsaOptions`, because all you do there is register workflow types. Perhaps we ought to consider workflows not to be part of ElsaOptions, but regular services instead (as is the situation before the changes introduced here).

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
sfmskywalker: Great. Then there's no need to register the activity type as a service, right?

Rafael94: Exactly.

src/core/Elsa.Core/Extensions/ElsaServiceCollectionExtensions.cs
Rafael94: We still have the problem. In `ProgrammaticWorkflowProvider` I need the instance of the workflow. All existing workflows are instantiated.  So here we have the problem again, as soon as a constructor has a parameter, which is not in the DI.

sfmskywalker: Not if the user registers their workflow like this:

```csharp
services.AddWorkflow(new SomeWorkflow("Foo"))
```

The above registers `SomeWorkflow` as a singleton.
In fact, we could simply add workflow instances to `ElsaOptions` instead of just their types. Or at the very least, we could register workflow factory methods, allowing the caller to provide a workflow instance as they see fit.

Rafael94: That would work, but then concrete instantiations would have to be made before calling `public static IServiceCollection AddWorkflow(this IServiceCollection services, Assembly assembly)`.

I have not yet dealt with factory methods. I would have to think about that.

sfmskywalker: Not necessarily; the `IServiceCollection` API allows you to specify a factory method which gets invoked whenever an object instance is requested (and only executed once when life time scope is Singleton).

Anyway, see what you can do. If you want we can always revisit this later after we merged in this PR.

Rafael94: Good idea. I was thinking too complicated.

Rafael94: Is added it now. Is this what you had in mind?

Unfortunately I don't know how to resolve conflicts with Github and a pull request.

sfmskywalker: Almost. Right now you are allowing the user to add an instance of a workflow, but then you only register its type and discard the object itself.

But that doesn't have to block this PR as far as I'm concerned - we can work on this after merge.
I'll work on the merge conflicts, but let me know if you want me to show you.

Rafael94: Sorry. I was a little careless there. When I wrote that, I was already thinking about the next method.

I have corrected it now.

sfmskywalker: No worries, that looks good. Although I am still not sure what the point is of registering the type or even the instance with the service container *and also* with `ElsaOptions`. I think we only need to do one *or* the other.

If this is OK with you, I'd like to get your PR merged in regardless. and make changes afterwards.

Rafael94: I did it just in case someone wants to add a concrete instance. 

With `workflows.Add((IWorkflow)ActivatorUtilities.CreateInstance(serviceProvider, workflow));` in `ProgrammaticWorkflowProvider` this instance should be retrieved.

However, if you would like it to be different, feel free to change it.

Issue #514, opened 2020-12-14
Elsa 2.0: Context (ContextId, ContextType)


Rafael94:
Hi,

I work a lot with entities in my projects (tickets, devices, people, etc.)

Users can create their own workflows and therefore there can be multiple instances for one entity.

CorrelationId must exist only once ( `Task<WorkflowInstance?> GetByCorrelationIdAsync(string correlationId, WorkflowStatus status,CancellationToken cancellationToken = default)`). That's why I would like to work with the `ContextId`. But at the moment I can't specify a type (ticket, device, ...) for the instance. I could only help myself out by applying such a construct `ContextId = $"Ticket_{TicketId}"`.

My suggestion:
- In the WorkflowDefinition and in the WorkflowInstance store the optional `ContextType` (ticket, device). For the entity trigger, I introduced `EntityNameAttribute`. It could be moved to `Elsa.Core` to determine the `ContextType`.
- `Task<IEnumerable<WorkflowInstance>> ListByContextAsync(string contextType, string? contextId, string? tenantId,CancellationToken cancellationToken = default);` add to `IWorkflowInstanceStore`.

For later:
- Filter activities and triggers by `ContextType` so that only general and context specific activities and triggers are shown.

sfmskywalker:
Your scenario makes sense, but I'm wondering if we could maybe  do something with custom "tags", allowing you to associate workflow instances with zero or more tags? The advantage of that is that other applications can associate more application-specific information with their workflows.

On the other hand, the `ContextId` is clearly intended for use with the subject/context of the workflow, so it might make sense to include its type as well. In fact, this would be a hard-requirement if we are to implement intellisense in expressions editing UIs.

Rafael94:
Do you mean tags or properties? Tags is just a list of keywords. With properties you can store values in addition to the keywords.

Node: Node-1
SystemWorkflow: true
...

As you say, we already have `ContextId` for contextual workflows, so it would make sense to store the type as well. 

The UI is another issue. For now, I'll try to improve the backend based on your excellent preliminary work according to my wishes and share them.  As a step by step improve the backend and then the UI.

Rafael94:
Can be closed, has been implemented.

PR #539, opened 2020-12-25
Add ContextType to WorkflowInstance

Rafael94:
Issue: #514 

The data type `Type` was not serialized correctly, which made a comparison from `ContextType` impossible. Therefore I decided to use string. I find it more convenient this way than writing a separate serializer for each database.

`ContextType` in `WorkflowContextOptions` could also be stored as `string` later.

After PR #536 and this PR. `WorkflowContextProvider` could also be stored in `ElsaOptions`. When registering a `WorkflowContextProvider`, the `ContextType` is also registered. This way all available `ContextTypes` for the designer are automatically available.

### DIFF
--- a/src/core/Elsa.Abstractions/Attributes/ContextTypeNameAttribute.cs
+++ b/src/core/Elsa.Abstractions/Attributes/ContextTypeNameAttribute.cs
@@ -1,0 +1,15 @@
+using System;
+
+namespace Elsa.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class ContextTypeNameAttribute : Attribute
+    {
+        public string Name { get; }
+
+        public ContextTypeNameAttribute(string name)
+        {
+            Name = name;
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


sfmskywalker:
I'm curious about the reason for moving context type to workflow instance, rather than keeping it part of the workflow definition?

Rafael94:
> I'm curious about the reason for moving context type to workflow instance, rather than keeping it part of the workflow definition?

I wanted to search directly in the instances and not take a detour via WorkflowDefinition. If I go via WorkflowDefinition, I have to query two different stores.

sfmskywalker:
I see. But given that the workflow definition still has a WorkflowContextId, would it not make sense for it to also store the type of the workflow context still? The designer for example will need to know the context type if it is to provide any form of intellisense.

Rafael94:
I don't know if I'm possibly misunderstanding you right now.

The context is still stored in the WorkflowDefinition (`WorkflowContextOptions`).

sfmskywalker:
Sorry you are right, I am seeing things that are not there 😅 Forget I said anything.

sfmskywalker (approved):

PR #550, opened 2021-01-04
Cosmos DB persistence layer refactoring

cvijayak:
1. Cleaned up the duplicate code
2. Enhanced the cosmos DB persistence assembly to inject the custom logic for handling tenants
3. Enhanced to inject the custom logic in Cosmos DB CURD operations


sfmskywalker (approved):
Looks good!

PR #685, opened 2021-02-26
merge workflow input and instance variables  instead of overwriting

stevedig:
Update Workflow.cs

merge input and instance variables  instead of overwriting


sfmskywalker (approved):

sfmskywalker:
Nice catch 👍🏻

PR #689, opened 2021-02-28
Resolves #564 temporal activities backend

### LINKED ISSUE #564: To avoid confusion, developers should not consume temporal (time-based) activities without an appropriate back-end
Right now Quartz is available as a separate package, but without either Quartz or Hangfire, the Timer activities are useless. One option could be to implement a simplistic in-memory scheduler, but Quartz already provides this, so we might use that as the default instead. We will keep the current abstractions so that Hangfire or other schedulers can be used instead of Quartz.

craigfowler:
Temporal activities now use the implementation-specific packages as their entry-point.

Actually some of the logic noted in the original issue had already been implemented.  To avoid developers using it by mistake, the method which adds temporal activities is no longer an extension method.  But it is still used by the two implementation methods.  I didn't fully move it out of the common package though, in order to avoid duplication.

Also I renamed a lot of namespaces Timers → Temporal.  I'm hoping that everything I changed is good to go.  The only thing I'd like @sfmskywalker to check-on particularly is a change in the Blazor app.  I haven't used Blazor before but I presume it infers namespace from directory structure?  A `.razor` file wasn't found on build until I changed its intermediate directory structure.


sfmskywalker (approved):
With tests and everything 🎉

sfmskywalker:
@craigfowler 
>  I presume it infers namespace from directory structure?

Yep, that's exactly what it does.

LGTM :shipit:

Issue #664, opened 2021-02-25
To simplify the config, AppVeyor YML should be managed per-branch


craigfowler:
Currently, the `appveyor.yml` file that's checked into source control (across all active long-lived branches) has the build configuration for _all of the branches_.  For example in the v2 branch the file has build configs for 1.4.x and 1.5.x as well as the 2.x code.  Which part of the yml build config is then chosen by an almost switch/case-like syntax.

To simplify the build configs, the copy of this file in each branch should only contain the build config for the code in that branch.  This is OK because when AppVeyor runs a build, it uses the `appveyor.yml` file _from the same commit as it is building_.  The extra build configs in the version of the file in each branch are actually never used anyway.

## Acceptance criteria
In this ticket, we will 'tidy up' the YML file in the v2 branch.  The master & 1.4 branches can be left as they are for now.  When Elsa 2 is finalised and merged into master, it will overwrite and 'fix itself' anyway.

_This ticket was created following a brief discussion started by @craigfowler in https://github.com/elsa-workflows/elsa-core/discussions/660_

craigfowler:
PR https://github.com/elsa-workflows/elsa-core/pull/678 _should resolve this_ once merged.

Only things to note are:
* We can't test that the MyGet deployment works until we see a build on a v2 branch (and not a feature branch as it's on now)
* We can't test that the NuGet deployment works until we see a build from a tagged commit

On that basis, _I'd suggest leaving this issue open_ until we have seen those two scenarios working OK.  Otherwise we can't be 100% certain that the change has been completely successful.

craigfowler:
[As of build 2.0.0-preview6.1504], I can see that MyGet deployment is working as-intended.

[As of build 2.0.0-preview6.1504]: https://ci.appveyor.com/project/sfmskywalker/elsa/builds/37955695

craigfowler:
@sfmskywalker at your next convenience/when appropriate (no rush on my account), if you create a new v2 preview tag on the repo then we can test that the other half of this works: Publishing to full NuGet.

I saw in that build that [it detected the "publish to full NuGet" deployment step and skipped it](https://ci.appveyor.com/project/sfmskywalker/elsa/builds/37955695#L938) because it didn't meet the criteria, so I do expect it to work when it's a tagged build.

Thanks

sfmskywalker:
[...]

PR #697, opened 2021-03-02
Stale Bot

jwillmer:
- You should got a request to enable stale bot in this project. 
- This pull will add the config file for the stale bot to the project. 

It does not matter if you merge first or after you approved the bot since stale bot will wait 24h before the first run.

https://github.com/probot/stale

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -11,7 +11,20 @@ daysUntilClose: 7
 onlyLabels: []
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
-exemptLabels: []
+exemptLabels: 
+  - activity
+  - bug
+  - core
+  - dashboard
+  - devops
+  - documentation
+  - enhancement
+  - good first issue
+  - help wanted
+  - persistence
+  - priority high
+  - priority low
+  - workflow tooling  
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: true
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,58 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels: []
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: true
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+
+# Comment to post when removing the stale label.
+# unmarkComment: >
+#   Your comment here.
+
+# Comment to post when closing a stale Issue or Pull Request.
+# closeComment: >
+#   Your comment here.
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+only: issues
+
+# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
+# pulls:
+#   daysUntilStale: 30
+#   markComment: >
+#     This pull request has been automatically marked as stale because it has not had
+#     recent activity. It will be closed if no further activity occurs. Thank you
+#     for your contributions.
+
+# issues:
+#   exemptLabels:
+#     - confirmed
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -17,13 +17,13 @@ exemptLabels:
   - core
   - dashboard
   - devops
+  - do later
+  - do now
+  - do soon
   - documentation
   - enhancement
   - good first issue
-  - help wanted
   - persistence
-  - priority high
-  - priority low
   - workflow tooling  
 
 # Set to true to ignore issues in a project (defaults to false)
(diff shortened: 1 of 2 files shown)


craigfowler:
There's already a discussion which touches on this here https://github.com/elsa-workflows/elsa-core/discussions/688

We might not need the Stale Bot when we get around to implementing that process.  My only concern about Stale bot is that it could end up closing issues in our backlog, just because they're not going to be implemented any time soon.  I can foresee myself 'fighting' a tool like this and re-opening a lot of what it closes, because the issue contains valuable information about a bug or wishlist feature which should be in the backlog.

Triage and issue grooming is something we need to get on top of, but I'm not sure that "close any issues which have no activity since X time" is the right solution yet.  I think that (eg: using the process suggested in the linked discussion) we could actually get out in front of our pile of issues, without discarding work/contributions which have already been made.

craigfowler:
Also, I notice this breaks the build but that's really my fault more than anything.  If you cherry-pick (or just copy the same change) made in https://github.com/elsa-workflows/elsa-core/commit/850b01b2d25372c1060c67d9d1095b0621c9bec6 then it'll fix the build.  Because of the secure variable (which is not evaluated on a PR build), the SonarScanner tries to run but fails on an invalid API token.  The only fix I could find was not to run SonarScanner for PRs.

jwillmer:
@craigfowler I configured the bot in a way that it does not close any issue that has already a label assigned (except `wont-fix` and `duplicate`) and that it does not touch any issue that is assigned to a user, project or milestone.

jwillmer:
I don't know if I should add the new `appveyor.yml` to this branch since it get's merged into master. And the master `appveyor.yml` says: 

```
# Note that this YML has significantly diverged from the v2 YML
# when v2 branch comes to be merged to master, there will be
# serious conflicts.  When that happens, you should probably just
# take the YML for v2 and discard what was in master.
```

craigfowler:
Agreed, but it implies that "If we have any old issues without labels", then if we haven't managed to triage & label them before the cut-off point (whatever is chosen), then they're going to get auto-closed.

This is why I started the discussion the other day, because I could we we have a lot of unprioritized issues needing triage and figured we should have a strategy for getting out in front of them.  Using stale bot is a different strategy and whilst one is not more "correct" than the other, IMO we should collectively choose which is right for the project.

I for one don't think it would take a massive amount of resource to triage & prioritize all of the issues we have at the moment into a backlog.  Heck, I'd volunteer to do it.  Bulk-closing large numbers of them just because they weren't spoken about recently could easily be a case of "throwing the baby out with the bath water".

jwillmer:
Another note about the config: After 60 days of inactivity the issue get's a tag `stale`. Then after some time (don't know exact how much) the issue get's closed. We could run the bot and then list all issues with the label `stale` and have a quick look that we don't loose valuable tasks.

craigfowler:
Ah, as for the `appveyor.yml` changes yes.  The v2 branch has appveyor build config set up that's good for v2.  The master branch has slightly different build steps and due to the history of the files - see https://github.com/elsa-workflows/elsa-core/issues/664 - I already know it's going to be a car-crash merge conflict when it comes down to merging the v2 branch into master.

That comment is basically saying "when that car crash happens, just discard everything that was in the master version of this file and take the content from the v2 branch".

So rather than merging the v2 appveyor into master (which would break it), just replay that one change from the commit I noted above into the master version.  You might not be able to cherry-pick but the diff should make it clear what needs to be changed.  That YML fix is the only change in the commit.

craigfowler:
> Another note about the config: After 60 days of inactivity the issue get's a tag stale. Then after some time (don't know exact how much) the issue get's closed. We could run the bot and then list all issues with the label stale and have a quick look that we don't loose valuable tasks.

That's certainly interesting.  That eases some of my concerns.
I'm going to have to run off and get on with my day job work for now though.  I'll pick this up again in ~7 hours.

jwillmer:
@craigfowler can you fix the appveyor script? I merged the changes you requested but it did not work.

craigfowler:
@jwillmer - you missed off a close-brace at the end of the first block of Powershell ;)

craigfowler:
So, now I've got a bit more time to stop and think this through, I do agree that it could be useful, but mainly for issues which are requests for support/help.  I personally don't think it's so useful for bug reports (ones which we have confirmed are real, at least) and feature enhancement requests.

Because I was wearing _"my day-job head"_ when I looked this morning, I was thinking of an issue tracker as _only being a product backlog_ containing bugs & enhancements.  At the day-job we have a totally separate helpdesk who deal with support requests, so I tend to forget they exist.  In the Github & open source world, yes, the issue tracker ends up containing a lot of "please help" requests too.  _For those_, I totally see the value of something like stale-bot.

## Stale support requests could be auto-closed
For a support-request, if someone asks for help and nobody was able to answer it after a few weeks (or perhaps even a month) then actually there's a really high chance that nobody will ever be able to answer it.

This is totally worthwhile.  It stops the issue tracker from filling up with unanswered questions which we're ignoring anyway.  If the person who asked the question urgently needs an answer they can bump with a comment I guess.  It's reasonable to accept that sometimes people will ask questions that we collectively can't answer, especially because we're all volunteers.

## Stale change-requests are "do later"
On the other hand, change requests (confirmed bugs & enhancements) which aren't high-priority enough to be scheduled now can be given the label **do later**.  They become _our product backlog_ and we don't want to close them unless we think that they should be **wontfix**.

As we develop, we're going to be finishing releases.  As that happens, we take a section away from our roadmap and turn it into a changelog.  When we do that, we can add a new section to the end of the roadmap with the next batch of things we would like to implement.  That's when we can skim through the **do later** issues to see if there's any that we'd like.  We can take the ones which are best/highest-priority, discuss & agree any details we deferred & promote them to **do soon**, ready for scheduling & implementation.

Where it comes to looking for issues to work on, we can totally use a filter `-label:"do later"` so that we are not encumbered by them.

## So, actually, I think I'm trying to agree with you!
So, having thought about all of this, what I'm really saying is that I think it's a good idea.  The only refinement I'd suggest is that we have stale-bot work on:

* Issues with no labels at all
* Issues with the **help wanted** label (support requests)

@sfmskywalker ^^ Do you have any thoughts on any of this (including the earlier discussion)?

jwillmer:
Just one note: Any new label we create we need to exclude for the bot to process like I did in [this commit](https://github.com/elsa-workflows/elsa-core/pull/697/commits/9f3eb1ad366ce96b982f717c9a53d76cd190db5a) - there is no option for "only process no label and labels in this list"

craigfowler:
Ah, OK.  That's not the end of the world for now though ;)

I'll push in a commit to deal with the new/edited labels I recently changed.

jwillmer:
@craigfowler prio high and low where meant to be used in the projects to define the priority of tasks. I know you can just sort them but often you like to mark high prio tasks so that you have an easy way of seeing that all prio tasks are at the top of the list.

craigfowler:
@jwillmer ah, for the priority labelling thing I've started #702 . Again, there was a parallel discussion about that which you might not have seen. That happened as #688.

In short what I did was swap them for do now/do soon/do later. High & low priority were edited to become soon & later. "do now" is a new label which currently has no issues (and ideally it's very rarely used).

But - I shouldn't have just edited the labels. I had a feeling that maybe you hadn't seen the discussion at #688 when you set the priority labels up.

craigfowler:
Generally-speaking btw, if I've got something which I'm suggesting project-wide, I'll post it in that design & tech category first.

sfmskywalker (approved):

sfmskywalker:
Awesome stuff guys. Let's start with this and see if it works for us.

PR #714, opened 2021-03-04
Resolve #548 - improperly connected activity

### LINKED ISSUE #548: The last WriteLine activity of sample25 is missing one connection
https://github.com/elsa-workflows/elsa-core/blob/3717c7b9509de89e3df2c8578311a69a5d91a326/src/samples/Sample25/DataProcessingWorkflow.cs#L54

This line of code is not executed

craigfowler:
The activity I removed would never have been executed, because it is
not connected from either of the IfElse branches.
Since this is Elsa 1.x, I've fixed it using the cheapest/quickest possible
solution which is to just remove the misleading part of the sample.

### DIFF
--- a/src/samples/Sample25/DataProcessingWorkflow.cs
+++ b/src/samples/Sample25/DataProcessingWorkflow.cs
@@ -50,8 +50,7 @@ namespace Sample25
                         ifElse
                             .When(OutcomeNames.True)
                             .Then<WriteLine>(x => x.TextExpression = new LiteralExpression("Data exceeds threshold (TRUE)"));
-                    })
-                .Then<WriteLine>(x => x.TextExpression = new JavaScriptExpression<string>("(`Finished data processing. Result: ${Absolute.Result}`)"));
+                    });
         }
     }
 }


sfmskywalker (approved):

sfmskywalker:
Interestingly enough, this example would have worked properly in Elsa 2. In Elsa 2, the `If` activity always returns a `"Done"` outcome _after_ either the `"True"` or `"False"` branch complete.

craigfowler:
Sipke, did you install the time machine extension to Visual Studio?

_When this codebase gets up to 88mph, you're gonna see some serious schtick_ :rofl:

craigfowler:
Also, _hopefully_ now we've got a new commit in master (since I reconfigured its recognition of long-lived branches), SonarCloud will recognise master as one.  That _should mean_ that it stops failing PRs for the wrong reasons :eyes:

sfmskywalker:
Ahhh that explains the failing build notifications I've been seeing (and blissfully ignoring 🙈 ).

PR #732, opened 2021-03-08
Indicis for EF Providers

DavidPx:
#709 

Added a number of indices to the EF context configuration.    

* multi-column: I tried to target the calls that the background services make
* single-column: these will help UIs filter/sort by the various fields

I've deployed/migrated all three providers and ran the sample app.

Note that the nvarchar SQL Server columns affected by this PR have been changed from `nvarchar(max)` to `nvarchar(450)` because:
A) SQL Server cannot index `varchar(max)` fields
B) Older verions have a 900 byte size limit to indices


sfmskywalker (approved):

sfmskywalker:
Great stuff. Thanks for the PR!

Issue #552, opened 2021-01-04
Bug: Elsa 2.0 MongoDb Deserializing


Rafael94:
When loading workflow instances I get the following error. 

`An error occurred while deserializing the Activities property of class Elsa.Models.WorkflowInstance: An error occurred while deserializing the Output property of class Elsa.Models.ActivityInstance: Unknown discriminator value 'FinishOutput'.`

I already know the cause of this: https://stackoverflow.com/a/43312423

`BsonClassMap.RegisterClassMap<FinishOutput>();` will fix the problem.

But it will probably happen more often. It would be best to register the classes centrally, so that in Elsa.Persistence.MongoDb only the central place has to be read and thus no dependencies to other projects exist.

I just don't know yet where it is nicest and how to name it best.

sfmskywalker:
That's a problem. I think I should be storing Output values as JTokens instead. The consuming code needs to know the targt type anyway, so they may just as well deserialize JTokens themselves (using convenience helpers).

Rafael94:
That is also a good idea.  Could you see if you can do it with `System.Text.Json` (https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/ `JsonElement`) instead of Newtonsoft? `System.Text.Json` comes directly from Microsoft and might be faster.

craigfowler:
@sfmskywalker - would you be opposed to splitting this into two or even three separate issues, so they can be prioritised independently?  We've got:

* This bug - reading from MongoDb crashes
* Revisiting what is stored in the db: A model class or a JSON token object
* Evaluating a switch to `System.Text.Json` from Json.NET

I think that these could be three separate pieces of work and might not _all_ need to go in 2.0.0.  The only concerns about _not doing one or more of these in 2.0.0_ is that I suspect changing what's stored in the DB would force a breaking change (pushing it to 3.0.0).

I'm not against any of these suggestions though; indeed, I think @Rafael94 's idea is a good one.  It does look suspiciously like it would become a solution-wide job though.  I see 89 matches for `using Newtonsoft.Json` at the moment, so it probably wouldn't be a quick swap.

sfmskywalker:
@craigfowler Not at all, I think you're spot-on with those suggestions.

[...]

PR #735, opened 2021-03-09
Resolve #728 workflow instances weren't persisting

### LINKED ISSUE #728: Workflow instances are not persisted when they should be
As demonstrated in the code at https://github.com/craigfowler/elsa-core/commit/672e32a7caae61355e928bc4c299e3687f468d17 (the new test added in this commit), the workflow instance which is created in this test never makes it to the MongoDb persistence store.  This is despite the workflow including a `UserTask` activity which would suspend the workflow and preventing it from completing immediately.

This issue is to diagnose and fix this issue with workflow persistence.  It has been proven to affect MongoDb and also anecdotally affects Entity Framework.  **Update**: This is confirmed to affect all persistence, _not just MongoDb_.

## Acceptance criteria
* A workflow should be persisted at at least one point after it has begun executing
* Test coverage must be added to prove that this works with at least an in-memory persistence store (ideally add a MongoDb test case as well)
* The default persistence behaviour for a workflow _should be altered to be_ `WorkflowPersistenceBehavior.WorkflowBurst` (currently it is implicitly `Suspended`)

craigfowler:
In short (there's A LOT MORE info in the comments to #728 itself):

* This bug wasn't (anywhere near) as serious as we first thought.  The test case I was using was triggering some behaviour which I didn't know about, which meant workflow instances didn't persist due to intended functionality.
* There was a (real) bug with the `ActivityExecuted` persistence behaviour - it wasn't triggering persistence of the workflow instance.  That's fixed now.
* I have changed the default persistence behaviour to `WorkflowBurst` as suggested
* This PR is based on the initial work I was doing towards #552.
  * Once this PR is merged to the v2 branch, I'll pick up #552 again by back-merging v2 into that feature branch and resuming the original investigation.

I left a TODO note on the persistence behaviour enum.  I'm totally willing to do that, and I half-encourage you to fail the PR and ask me to do it.  I could use a little help pointing out where I would need to consume that service though.  I don't yet know all of the places which could set up a persistence behaviour (or is it _only_ the workflow builder?).

Alternatively I could just delete the TODO comment and leave it working the same as it is already.  I'm only anxious about it because it's implicit behaviour and not explicit.  Change the order of items in the enum and whoops!


src/core/Elsa.Abstractions/Models/WorkflowPersistenceBehavior.cs
```diff
@@ -6,15 +6,21 @@ namespace Elsa.Models
...
-        Suspended,
-        
+        WorkflowBurst = 0,
+
         /// <summary>
-        /// Workflow instances are persisted after the workflow completed a burst of execution.
+        /// Workflow instances are persisted only when being suspended. 
         /// </summary>
-        WorkflowBurst,
+        Suspended,
```
sfmskywalker, line 23: Indeed, let's not change the order like this, but instead explicitly set the default behavior when constructing a workflow blueprint.

When building workflows using the builder API, this means setting the default behavior on `WorkflowBuilder`.
When building workflows using the designer, this means updating the designer to an empty workflow instance with the default behavior.

The reason I prefer not to change the order of the enum values is because they go from "persists lazily" --> "persist eagerly"

PS: You made a typo in your comment ;)

src/core/Elsa.Core/Handlers/PersistWorkflow.cs
```diff
@@ -28,12 +29,16 @@ public PersistWorkflow(IWorkflowInstanceStore workflowInstanceStore, ILogger<Per
 
         public async Task Handle(WorkflowSuspended notification, CancellationToken cancellationToken)
         {
+            _logger.LogTrace("Received notification {notification}", notification);
```
sfmskywalker, line 32: I can see how logging is helpful. So how about we introduce a logging behavior for Mediatr? Something like what is described in [this blog post](https://dev.to/raulismasiukas/logging-pipeline-with-mediatr-5d4f). Then we can remove the repetitive logging statements, while automatically getting logging for all notification handlers.

sfmskywalker (changes requested):
Just a couple of small things to consider.
Beyond that, I agree with the fix. 

PS. Digging the integrations tests!

src/core/Elsa.Abstractions/Models/WorkflowPersistenceBehavior.cs
craigfowler: I'm glad you asked about this, because it's the exact same as I was thinking.

> When building workflows using the designer, this means updating the designer to an empty workflow instance with the default behavior.

Do you have any hints as to where the code change needs to be made in order to the bit where you suggest _"updating the designer to an empty workflow instance"_?  I've not even looked at the designer/dashboard/UI stuff at all yet and I know it's currently in flux so I've not been rushing to look into it yet either.  So any pointer to the right code welcome thanks :)

Otherwise, totally, I can make the change in the builder class.  Are there any other places you can think of which might initialise a persistence behaviour?

PS:  It actually took me a few re-reads to remember that US English doesn't include a U in the word "behaviour" 😆   Goes all the way back to things I was taught in childhood to remember the spelling:  "It's not **my** behaviour, it's **our** behaviour".  Thankfully I'm deleting that comment and reverting that part of the change anyway.

src/core/Elsa.Core/Handlers/PersistWorkflow.cs
craigfowler: That seems pretty cool and I did wonder at the time whether I could do that logging better.  It wasn't my chief concern though because I only wanted some proof that the persistence methods were actually being used.

The generalised solution in the blog looks good.  The only thing I think which might be useful to add to that would be to record which handler method(s) are receiving the notifications.  I put the logging there to try and figure out "did this even get called?" (I'd have used a breakpoint but Omnisharp's debugger kept crashing on me when executing this code).  So, just "mediatr received a notification of type X and relayed it to zero or more handlers" doesn't tell me "Handler X was notified & Handler Y was notified but Handler Z was not" etc.

I'll see if I can adapt the example in the blog though, to a degree I'd forgotten I even left the logging logic in the PR.  But now it's in there I may as well do it properly 😄

src/core/Elsa.Abstractions/Models/WorkflowPersistenceBehavior.cs
sfmskywalker: Sure, it's right here on [this line](https://github.com/elsa-workflows/elsa-core/blob/feature/elsa-2.0/src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx#L18).

But feel free to skip it, I can just as easily add & test it.

PS: Hahaha sorry, I just couldn't help myself 😄

src/core/Elsa.Core/Handlers/PersistWorkflow.cs
sfmskywalker: > I'll see if I can adapt the example in the blog though, to a degree I'd forgotten I even left the logging logic in the PR. But now it's in there I may as well do it properly 😄

Sounds good! But feel free to do it as a separate PR if you wish.

src/core/Elsa.Abstractions/Models/WorkflowPersistenceBehavior.cs
craigfowler: Done

src/core/Elsa.Core/Handlers/PersistWorkflow.cs
craigfowler: I backed-out the repeated logging.  I had a look into the technique linked in the blog post but [it is no good for notification handlers, only request handlers](https://github.com/jbogard/MediatR/wiki/Behaviors#nb).

I went on to look into [the technique described here, adding custom publish behaviour](https://github.com/jbogard/MediatR/wiki#publish-strategies) - overriding `PublishCore` to inject logging.  It worked to a degree but clearly the instances of `Func<INotification, CancellationToken, Task>` which are received by this method do not point to _the real/original methods_ on the notification handler classes, they are lambda functions which wrap those original methods.  That means that [the `Target`  and `Method` properties](https://docs.microsoft.com/en-us/dotnet/api/system.delegate?view=net-5.0#properties) of that delegate provide only meaningless names, and don't point to the actual concrete class which was going to handle the notification.

Since this is an aside from the overall aim of this PR, I've just backed it all out.  The net effect is that I think there's not much (if any) change now.

sfmskywalker (approved):
Great stuff. Thank you!

PR #741, opened 2021-03-10
Refactor options provider with DI support

sfmskywalker:


craigfowler (approved):
That's certainly the kind of solution I was describing in Discord.

A couple of things to consider (but 100% up to you to ignore or not):
* The act of "getting the options" could totally be refactored out of `ActivityDescriber` into a service (interface & impl class) of its own.
  * That would help follow [SRP] and also make the code easier to test in a self-contained manner (more on that below).
  * It also reduces the cognitive complexity of each overall class.  Someone reading `ActivityDescriber` only needs know that a (say) `IGetsOptionsForActivityProperty` gets the options somehow.  Someone reading the impl class for that interface need not worry about the other contextual aspects of describing an activity, since getting the options is a stand-alone responsibility.
* You could totally create a test for this at the unit-testing level.  Take an `ActivityDescriber`, with a sample activity (decorated with attribs), then test that the describe method returns an object which was created using the logic you declared.  It'd serve as both an example which can be followed, and help prevent regressions.
  * The "cost of entry" to writing that test would probably be lower if the functionality were refactored out as above.  The unit test wouldn't be broken if something related to the rest of describing an activity "got in the way" of the test logic.

[SRP]: https://en.wikipedia.org/wiki/Single-responsibility_principle

But yes, these are low-priority suggestions rather than my saying _"you shouldn't merge this without doing this stuff"_.

sfmskywalker:
Very good suggestions, I agree with all of them. What I'll do as part of this PR is extract the options provider resolver thing. Although you won't find me writing tests until anytime soon until after the dust has settled, at least it would make that path a little bit easier ;)

PR #749, opened 2021-03-11
Add Azure Servicebus Topic Activities (Sender & Receiver)

jdevillard:
This pull request is to add Service Bus Topic Activities to the Elsa.Activities.AzureServiceBus library.

Don't want to create a Breaking Change but maybe : 
- AzureServiceBusMessageReceived could be rename to AzureServiceBusQueueMessageReceived
- SendAzureServiceBusMessage could be rename to SendAzureServiceBusQueueMessage

On the MessageBusFactory class, Queue and Topic Sender rely on the same type of Sender Object (using an entityPath), but we've to manage Queue or Topic creation with different methods.

Question : 

- I'm not a big fan of mixing Service Bus Entity Management and Runtime. Connection String that give only Send/Receive permissions can't be use due to use of Ensure**Entity**Exist and Create**Entity**. Maybe an other PR could separate this? any idea?

- with current impl of Sender (Topic or Queue) and the use of the defaultContentSerializer, we can't send a simple string to the Queue/Topic. It could be interesting to send all type of data.
- Could be interesting to add a DictionaryProperty to MessageModel to get the Properties of the BrockeredMessage. (Which could be used by the user in the Workflow)

- Last point, could be also interesting to check if the last Azure.Messaging.ServiceBus would not be better (for perf, protocol usage and to better manage MessageBrocker, I've seen that a lot of improvement has been done) 

Let me know how I can help on this.

Thanks.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=749) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=749) before we can accept your contribution.<br/><br/>:x: jdevillard [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=749)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=749) it.</sub>

sfmskywalker:
Hi @jdevillard,

Thanks for the PR, it looks quite good!

Regarding your renaming suggestions, I think we should do it. It's OK to introduce breaking changes at this stage. Let me know if you'd like to do it as part of this PR or a separate one.

Regarding your questions:

1. You make a very good point. Yes, we should separate this as part of a separate PR. Do you want to do it, or perhaps at least create a separate issue for it with some thoughts on how we might approach this? What I think is important is that by default, the queues get created automatically. Or at least throw a clear exception when the user starts Elsa and didn't create the necessary queues. In any case, it would be great if the user can opt-in or out of automatically creating the queues.
2. We should fix that. I do want the user to be able to send simple strings. If you have an idea what changes would have to be done, then please feel free to open an issue or PR directly for this if you want.
3. I'm not at all against that. 
4. Same as #3.

As per the above, if you'd like to take on any of this then that would be nothing short of fantastic!

src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/SendAzureServiceBusTopicMessage.cs
```diff
@@ -0,0 +1,42 @@
...
+            _serializer = serializer;
+        }
+
+        [ActivityProperty] public string TopicName { get; set; } = default!;
+        [ActivityProperty] public object Message { get; set; } = default!;
+
+        protected override async ValueTask<IActivityExecutionResult> OnExecuteAsync(ActivityExecutionContext context)
+        {
+            var sender = await _messageSenderFactory.GetTopicSenderAsync(TopicName, context.CancellationToken);
+            var json = _serializer.Serialize(Message);
```
sfmskywalker, line 31: As per your suggestion, perhaps here we might first perform a check if the `Message` is a string value. If anything else, it gets serialized, otherwise sent as-is.

sfmskywalker (approved):

src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/SendAzureServiceBusTopicMessage.cs
jdevillard: yep , I'll change this part to be sure that "simple" work (need to do on Sender And Receiver.

sfmskywalker:
@jdevillard Just say the word when you want this merged in.

jdevillard:
I think I ll add a last commit with the renaming for the queue and it ll be ok for this pr. I'll open others with the new issue.

jdevillard:
@sfmskywalker ok for me!

sfmskywalker (approved):

PR #758, opened 2021-03-14
Resolve #683 - Tests to prove no exception

### LINKED ISSUE #683: As a developer, to receive appropriate guidance, I would like to see a more helpful exception than an Automapper one
When a developer making use of Elsa APIs causes a problem whereby an Activity is added to a Workflow Instance more than once, they receive an exception similar to the one pictured below.

image

This scenario _is always an error_ and I think that there is not much choice but to throw an exception in this scenario.  The reason that this should be improved is that the exception above isn't particularly helpful to the developer who is working with Elsa.  This is because they are not themselves consuming Automapper and should not need to have to look through Elsa's internals to determine what is actually wrong.

In this case, Elsa should catch the Automapper exception and then rethrow a specific Elsa exception, which has a more helpful message.  It should tell them roughly what is wrong, conceptually how to fix it and perhaps provide some information which will help identity which Activity has caused the problem (not just a GUID identity).

## Acceptance criteria
* Find some repro steps to reliably reproduce this error
* Catch exceptions thrown in this specific scenario (or perhaps pre-emptively test for the scenario)
* When the scenario occurs, Elsa should throw a more useful exception which:
  * Explicitly states that the same activity is being added to a workflow instance more than once
  * That this is not allowed
  * Provides some useful identifying information about the offending activity and perhaps the offending workflow instance (beware of including potentially sensitive data though)
* Test that the new exception is thrown when reproducing the error scenario

craigfowler:
As noted in the discussion for #683 - the issue is no longer relevant as-reported.  There might be some further consideration to be had as to _whether or not we truly wish to permit duplicate activity IDs_, but that is a conversation which does not necessarily need to happen right now.

### DIFF
--- a/test/integration/Elsa.Core.IntegrationTests/Autofixture/HostBuilderWithDuplicateActivitiesWorkflowAttributes.cs
+++ b/test/integration/Elsa.Core.IntegrationTests/Autofixture/HostBuilderWithDuplicateActivitiesWorkflowAttributes.cs
@@ -1,0 +1,80 @@
+using System.Reflection;
+using AutoFixture;
+using AutoFixture.Xunit2;
+using Elsa.Core.IntegrationTests.Workflows;
+using Elsa.Persistence.MongoDb.Extensions;
+using Elsa.Testing.Shared.AutoFixture.Customizations;
+using Microsoft.Extensions.DependencyInjection;
+using Elsa.Persistence.EntityFramework.Core.Extensions;
+using Microsoft.EntityFrameworkCore;
+using Elsa.Persistence.EntityFramework.Sqlite;
+using Elsa.Persistence.YesSql;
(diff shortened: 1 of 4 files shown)


craigfowler:
You might not want to merge this @sfmskywalker .  Given that these are not tests for actually-desired functionality.
The fact that you can see the tests pass once shows that #683 does not need any code changes.  But I'm not convinced about the long-term benefit of keeping these tests around.

I leave it to you to either merge or "look and see that it proves what it says it proves" then delete the PR without merging.

sfmskywalker (approved):
It's OK to have these tests, especially with your clear explanation written in the comments. If we change the behavior, the tests will fail, and it will be immediately clear what we'd need to update.

PR #805, opened 2021-03-25
Resolve #785 - improve API for StringValuesModel

### LINKED ISSUE #785: Use of an HTTP request QueryString in a Javascript expression can raise InvalidCastException unless toString() is used
I stumbled upon a problem, better to say I'm confused......
I have HTTP endpoint immediately followed by HTTP response.

If I place: **"" +input.QueryString["foo"]** for content I get correct response (text/plain) used.
However if I place: **input.QueryString["foo"]** nothing is displayed and console reports warning:
warn: Elsa.Expressions.ExpressionEvaluator[0]
      Failed to evaluate expression input.QueryString["foo"] using syntax JavaScript
      System.InvalidCastException: Object must implement IConvertible.

Can you please explain why one works and other not and if is there any bit more verbose documentation which could solve my confusion. Thanks.

craigfowler:
Mostly this is non-controversial (details in the commit message but it's going to be real boring stuff).

_The only thing I wanted to flag up_ is that implementing `IConvertible` makes this class [non-CLS-compliant].  Right now, the assembly is not marked with the `CLSCompliantAttribute` and so we are not claiming to be CLS-compliant.  I am not sure how important CLS-compliance is to us though, generally speaking CLS-compliant libraries are more reusable across a wider variety of platforms.

* If we don't care about CLS compliance then _"nothing to see here, move along"_
* If we think we might care about CLS compliance then when we add `[assembly: CLSCompliant(true)]`, we need to add `[CLSCompliant(false)]` to this class

[non-CLS-compliant]: https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components

### DIFF
--- a/src/activities/Elsa.Activities.Http/Models/StringValuesModel.cs
+++ b/src/activities/Elsa.Activities.Http/Models/StringValuesModel.cs
@@ -1,33 +1,84 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Extensions.Primitives;
 
 namespace Elsa.Activities.Http.Models
 {
-    public class StringValuesModel
+    /// <summary>
+    /// Represents a collection of zero or more string values.  It is used for values such as
+    /// HTTP header/query string values.  These are often thought-of as key/value pairs but might
+    /// have more than one value.
+    /// </summary>
+    public class StringValuesModel : IConvertible
     {
-        public string? Value { get; set; }
-        public string[]? Values { get; set; }
+        /// <summary>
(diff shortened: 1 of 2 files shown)


sfmskywalker (approved):

sfmskywalker:
I wasn’t aware of the CLS thing. I suppose we can always revisit this if we care in the future, right?

craigfowler:
Sure. If I recall correctly (it's been a while), the compiler will scream at you if you include non-CLS compliant APIs in an assembly marked as compliant.

Mainly it's the usage/presence of the unsigned numeric/integral types in the API.

PR #810, opened 2021-03-26
Method to skip adding core activites automatically.

tomy2105:
As discussed in https://github.com/elsa-workflows/elsa-core/discussions/802 added method to options called `NoCoreActivities()` , this enables user of Elsa to easily control which of the core activities will be available. 

For example, to have only SetVariable activity:
`
            services

                .AddElsa(options => options

                    .NoCoreActivities()

                    .UseEntityFrameworkPersistence(ef => ef.UseSqlite())

                    .AddActivity<Elsa.Activities.Primitives.SetVariable>()

                );
`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,10 @@ _ReSharper*/
 [Tt]est[Rr]esult*
 .vs/
 
+# VS Code/Omnisharp crash dumps
+mono_crash.mem*.blob
+
+
 #Nuget packages folder
 packages/
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -121,6 +121,8 @@ deploy:
     on:
       APPVEYOR_REPO_TAG: true
 
-deploy_script:
-  - docker login -u="$env:docker_user" -p="$env:docker_pass"
-  - docker push elsaworkflows/elsadashboard-samples-monolith
+# Docker deployments temporarily disabled due to authentication issue
+# this is causing failing builds.
+#deploy_script:
+#  - docker login -u="$env:docker_user" -p="$env:docker_pass"
+#  - docker push elsaworkflows/elsadashboard-samples-monolith
--- a/src/core/Elsa.Abstractions/Exceptions/CannotSetActivityPropertyValueException.cs
+++ b/src/core/Elsa.Abstractions/Exceptions/CannotSetActivityPropertyValueException.cs
@@ -1,0 +1,16 @@
+namespace Elsa.Exceptions
+{
+    /// <summary>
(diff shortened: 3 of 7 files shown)


craigfowler:
### The build failure's not your fault
You'll have probably guessed, the reason for the build failure on this PR has nothing to do you with you.  The 2.0 feature branch was broken at that point in time.  I've since fixed it.  If you merge latest `feature/elsa-2.0` into your PR and re-publish then you should be good to go.

### Is it a good idea to allow de-registration of core activities?
I had a little look around and whilst a few of the core activities seem to have minor inter-dependencies they're not ones which would cause those activities to be resolved.  So, from my view not-registering them would not cause an actual problem.

I did notice though, that @sfmskywalker added some (very) recent work where an extra activity library makes use of some of the core activities.  For example [this Telnyx activity makes use of `Finish`] and also `Fork`.  If you tried using that Telnyx activity and you had excluded the core activities, then _you're gonna' have a bad time_.

I suppose the question really is "is it reasonable for these core activities to be unavailable?".  If we're writing other activities which take the assumption that the core ones are registered and present then maybe that's not so great.

Since the ink is barely dry on Telnyx activities, we might have an opportunity to change our minds, if that's appropriate?  

[this Telnyx activity makes use of `Finish`]: https://github.com/elsa-workflows/elsa-core/blob/3ba1582b54d167f46b9e935627c2fbc349cc7325/src/activities/Elsa.Activities.Telnyx/Activities/CallRingGroup.cs#L160

sfmskywalker:
What we could do is to require activity modules (such as Telnyx) to register dependency activities themselves. So for instance, if a user calls `.AddTelnyxActivities`, then that extension method will also call e.g. `AddCoreActivities` to make sure all the primitive activities are there.

We would just have to make sure that these "add activities" extensions are safe to be called more than once.

tomy2105:
If a user calls .NoCoreActivites and calls .AddTelnyxActivities he would need to manually add activities that telenyx requires and it is up to you to document which are needed.

I wouldn't make .AddTelnyxActivities, call AddCoreActivities but only the ones that are required by telenyx, for example add fork activity only.

sfmskywalker:
> I wouldn't make .AddTelnyxActivities, call AddCoreActivities but only the ones that are required by telenyx, for example add fork activity only.

I agree - only the dependency activities should be added.

sfmskywalker:
In that case I'm good to accept this PR - it makes sense for a user to control Elsa behavior via the options as is done here. We can work on registering dependencies separately.

sfmskywalker (approved):

PR #814, opened 2021-03-28
Temporarily disable docker deployment

craigfowler:
It looks like the docker credentials are incorrect, because
builds are failing with an auth problem.  For example build 2.0.0-preview7.1704:
  https://ci.appveyor.com/project/sfmskywalker/elsa/builds/38438996#L6715

This can be re-enabled once the auth issue is fixed.  There should be
no real harm in disabling it, since it's not working at the moment anyway.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -121,6 +121,8 @@ deploy:
     on:
       APPVEYOR_REPO_TAG: true
 
-deploy_script:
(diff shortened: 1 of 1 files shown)


sfmskywalker (approved):
I agree. I need a bit more time to figure out why my credentials aren't accepted.

Issue #812, opened 2021-03-26
Problem with URL of Send HTTP Request Activity


tomy2105:
Last Elsa 2.0 built from code.
I cannot make Send HTTP Request Activity work with URL set to `https://tryphp.w3schools.com/demo/demo_global_post.php`.
It gives me following error:
`
      Failed to evaluate expression https://tryphp.w3schools.com/demo/demo_global_post.php using syntax Literal

      System.ArgumentException: The path in 'value' must start with '/'. (Parameter 'value')
`

I believe that the problem lies in usage of `PathString` as a type for `Url` in `SendHttpRequest.cs`.
Does anyone know why it `PathString` is used here?
I think `Uri` would be better suited.
I can post this simple Pull request on Monday unless there is a specific reason to use `PathString`.

sfmskywalker:
PathString is a mistake. It should be Uri as you suggested. A PR would be perfectly splendid!

Issue #811, opened 2021-03-26
Support for application/x-www-form-urlencoded content type in Send HTTP Request Activity


tomy2105:
Would be good to have this supported.
Even at basic level where Content needs to be specified in "raw" form like `name1=value1&name2=value2`. Or if using Javascript expression they could use it to encode values properly `"name1=" + encodeURIComponent("value1") + "&name2=" + encodeURIComponent("value2")`

PR #819, opened 2021-03-29
SendHttpRequest enhancement for form post as well as URL bug resolution

tomy2105:
This adds another content type to Send HTTP Requests activity to enable form posting as described in https://github.com/elsa-workflows/elsa-core/issues/811.
In addition it resolves Send HTTP Requests activity  Url bug described in https://github.com/elsa-workflows/elsa-core/issues/812.


craigfowler:
Thanks for this @tomy2105 !

PR #834, opened 2021-04-02
Add composable autofixture attributes

craigfowler:
This is a refactor of some testing logic, to counteract the bloat that was beginning to emerge.

The big deal is that a number of autofixture attributes may now be composed on the same AutoMoqData parameter, rather than having to choose just one.  This uses a different Autofixture mechanism: **Behaviours** (specifically `Postprocessor`) instead of only relying on specimen builders.

There is one example in commit 7044e692315d51b2da876eff9f3c6fc283e7c06a, which is a rework of the test at `TriggersForActivityBlueprintAndWorkflowProviderTests` so that the `IServiceProvider` used by that test _has both_ `FrozenAttribute` and `WithAutofixtureResolutionAttribute`.  Before this PR, only one of these two attributes would have been effective (the other would have been silently ignored).

The benefit is that it allowed the objects `workflowExecutionContext` & `activityExecutionContext` to be moved into the test parameters, because they no longer needed to be constructed manually.


test/integration/Elsa.Core.IntegrationTests/Persistence/EntityFramework/EntityFrameworkIntegrationTests.cs
```diff
@@ -12,17 +12,17 @@
...
 // ReSharper disable EF1001
 namespace Elsa.Core.IntegrationTests.Persistence.EntityFramework
 {
     public class EntityFrameworkIntegrationTests
     {
         [Theory(DisplayName = "A persistable workflow instance with default persistence behaviour should be persisted-to and readable-from an Entity Framework store after being run"), AutoMoqData]
-        public async Task APersistableWorkflowInstanceWithDefaultPersistenceBehaviourShouldBeRoundTrippable(
-            [HostBuilderWithPersistableWorkflowAndEfSqlite]
-            IHostBuilder hostBuilder)
+        public async Task APersistableWorkflowInstanceWithDefaultPersistanceBehaviourShouldBeRoundTrippable([WithPersistableWorkflow,WithEntityFramework] ElsaHostBuilderBuilder hostBuilderBuilder)
```
sfmskywalker, line 23: Should it be "Persistence" instead of "Persistance"?

sfmskywalker (approved):
Amazing work 🙏🏻

test/integration/Elsa.Core.IntegrationTests/Persistence/EntityFramework/EntityFrameworkIntegrationTests.cs
craigfowler: It should!  And now it is!

PR #840, opened 2021-04-07
Feature/blob storage

lucapisano:
Dear team,
I'd like to contribute to the project by adding BlobStorage activities to manipulate blob objects.
Kind regards,
Luca


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=840) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=840) before we can accept your contribution.<br/><br/>:x: lucapisano [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=840)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=840) it.</sub>

src/activities/Elsa.Activities.BlobStorage/Activities/BlobExists.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class BlobExists : Activity
+    {
+        public BlobExists(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The ID of the blob")]
+        public string BlobID { get; set; }
```
sfmskywalker, line 28: Please rename to `BlobId`.

src/activities/Elsa.Activities.BlobStorage/Activities/DeleteBlob.cs
```diff
@@ -0,0 +1,42 @@
...
+        public DeleteBlob(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The ID of the blob")]
+        public string BlobID { get; set; }
+        [ActivityProperty(Hint = "The IDs of the blob")]
+        public List<string> BlobIDs { get; set; } = new List<string>();
```
sfmskywalker, line 30: Please change the type to `IList<string>` - in this project's code base, we tend to expose interfaces (when available) rather than concrete types.

src/activities/Elsa.Activities.BlobStorage/Activities/DeleteBlob.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class DeleteBlob : Activity
+    {
+        public DeleteBlob(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The ID of the blob")]
+        public string BlobID { get; set; }
```
sfmskywalker, line 28: From the designer's point of view, it would be better to remove the `BlobId` property and instead have just the `BlobIds` one. When configured with `UIHint = ActivityPropertyUIHints.MultiText`, `DefaultSyntax = SyntaxNames.Json` and `SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid }`, the user by default will get a nice tagged input control.

src/activities/Elsa.Activities.BlobStorage/Activities/ReadBlob.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string FilePath { get; set; }
+
+        protected override async ValueTask<IActivityExecutionResult> OnExecuteAsync(ActivityExecutionContext context)
+        {
+            if (string.IsNullOrWhiteSpace(BlobID))
+                throw new System.Exception($"BlobID must have a value");
+            if (!string.IsNullOrWhiteSpace(FilePath))
+                await _storage.ReadToFileAsync(BlobID, FilePath, context.CancellationToken);
+            else
+                Bytes = await _storage.ReadBytesAsync(BlobID);
```
sfmskywalker, line 42: Setting `Bytes` like this won't do much.  What you might want to do instead is return the bytes as output, e.g.:

```csharp
var bytes = await _storage.ReadBytesAsync(BlobId);
return Done(bytes);
```

Then the blob bytes will be available to the workflow as this activity's output.

src/activities/Elsa.Activities.BlobStorage/Activities/ReadBlob.cs
```diff
@@ -0,0 +1,46 @@
...
+        public ReadBlob(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The ID assigned to the blob.")]
+        [Required]
+        public string BlobID { get; set; }
+        [ActivityProperty(Hint = "The bytes")]
```
sfmskywalker, line 30: Please add line breaks between multiline-properties.

sfmskywalker (changes requested):
Thank you for the PR!

I had only some minor feedback. If you could take a look at that then that would be great.

src/activities/Elsa.Activities.BlobStorage/Activities/ReadBlob.cs
lucapisano: @sfmskywalker 
Thank for your feedback. Just a question here: from DB serialization perspective, it is safe to pass a byte[] as Activity result?
I mean, Bytes should be available for the next activity to consume, but it shouldn't be persisted in the DB for the following reasons:
-) low serialization performance
-) big db records/documents
-) conflict in serialization mappings
Ref: https://elsa-workflows.github.io/elsa-core/docs/extensibility-custom-activities#state

sfmskywalker: Good point. Right now the user can configure whether a given activity's output should be persisted or not, so perhaps that's enough. But perhaps we need to consider allowing the activity itself to control whether or not the output gets persisted or not.

The simplest solution for now would be to:

1. Return the bytes as output.
2. Update Elsa to *not* store output by default, but instead require the user to explicitly configure an activity's "Persist Output" setting.

lucapisano:
@sfmskywalker Thank you again for the feedback, please find revisions in the commit 8f82354

src/activities/Elsa.Activities.BlobStorage/Activities/DeleteBlob.cs
```diff
@@ -0,0 +1,43 @@
...
+        [ActivityProperty(Hint = "The Ids of the blobs", 
+            UIHint = ActivityPropertyUIHints.MultiText, 
+            DefaultSyntax = SyntaxNames.Json,
+            SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid })]
+        public IList<string> BlobIds { get; set; } = new List<string>();
+
+        protected override async ValueTask<IActivityExecutionResult> OnExecuteAsync(ActivityExecutionContext context)
+        {
+            if (BlobIds==default || !BlobIds.Any())
+                throw new System.Exception($"BlobID or BlobIds must have a value");
```
sfmskywalker, line 38: Let's change this to `"BlobIds must have a value"`

src/activities/Elsa.Activities.BlobStorage/Activities/DeleteBlob.cs
```diff
@@ -0,0 +1,43 @@
...
+    )]
+    public class DeleteBlob : Activity
+    {
+        public DeleteBlob(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The Ids of the blobs",
```
sfmskywalker, line 29: Although we use `Id` when naming variables, when writing plain English descriptions, comments etc. we should use "ID". Using `Id` is just a C# naming convention where we use _PascalCasing_ for abbreviations. Unless it concerns two-letter acronyms such as IO (Input Output). Oftentimes, people confuse ID to be an acronym, although it's not: it's simply an abbreviation of "Identity", which is why the naming convention we use dictates it to be Id instead of ID. But when it comes to regular English, different rules apply.

src/activities/Elsa.Activities.BlobStorage/Activities/ReadBlob.cs
```diff
@@ -0,0 +1,47 @@
...
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The Id assigned to the blob.")]
+        [Required]
+        public string BlobId { get; set; }
+
+        [ActivityProperty(Hint = "If set, the output of this activity is written into the specified file, alternatively the outcome contains the bytes of the blob")]
+        public string FilePath { get; set; }
```
sfmskywalker, line 32: Great idea!

src/activities/Elsa.Activities.BlobStorage/Activities/WriteBlob.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        public WriteBlob(IBlobStorage storage)
+        {
+            _storage = storage;
+        }
+        private readonly IBlobStorage _storage;
+
+        [ActivityProperty(Hint = "The ID to be assigned to the blob. It's needed to retrieve the blob")]
+        [Required]
+        public string BlobID { get; set; }
```
sfmskywalker, line 29: Let's rename this to `BlobId`.

sfmskywalker (changes requested):
Only a couple more small naming feedback (which means this PR is in a great shape since I have nothing else to complain about 😄 )

Thanks!

jdevillard:
Hello, 

what about the capability to configure a connection string that can be use for the different activity? (Maybe for another PR)

sfmskywalker (approved):

Issue #781, opened 2021-03-22
For readability, users should see connections from a multiple-output activity more intuitively placed/styled


tomy2105:
Activities with more than one output seem to order subsequent activities in first added leftmost fashion.
Which tends to be confusing and hard to follow if you add activity for right output first (lines are crossed and hard to follow).
The more outputs, the more it can become unreadable :(.

In the image below iterate is connected to set var activity and done to http response activity.
![OutputOrder](https://user-images.githubusercontent.com/7658839/112009421-a4830480-8b26-11eb-8d21-9c0de36f4394.gif)

craigfowler:
Agreed, I can see how what you have there is pretty confusing!

tomy2105:
There is a second issue related to this where there is no cross overlap of lines (left output goes to right child activity and right output to left child activity) but lines simple overlap when more than 3 outputs even in case child activites are sorted same as outputs.

![Image 2](https://user-images.githubusercontent.com/7658839/114876851-06215080-9dff-11eb-88e5-c09bfe2b472e.png)

sfmskywalker:
Fixed:

![image](https://user-images.githubusercontent.com/938393/116413684-24903e80-a838-11eb-9ded-3bd998353ea4.png)

![image](https://user-images.githubusercontent.com/938393/116414063-746f0580-a838-11eb-9cc2-51380613daa4.png)

PR #867, opened 2021-04-15
Sort Child Activities Based On Output Order

tomy2105:
This is my attempt on solving first part of the problem described in https://github.com/elsa-workflows/elsa-core/issues/781.
Sorting child activities based on output order should prevent lines from crossing left to right and vice versa.

However second part of the problem remains so https://github.com/elsa-workflows/elsa-core/issues/781 should remain open.

### DIFF
--- a/src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx
+++ b/src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx
@@ -1,7 +1,7 @@
 import {Component, Host, h, Prop, State, Event, EventEmitter, Listen, Watch, Method} from '@stencil/core';
 import {Map, addConnection, findActivity, getChildActivities, getInboundConnections, getOutboundConnections, removeActivity} from '../../../../utils/utils';
 import {cleanup, destroy, updateConnections} from '../../../../utils/jsplumb-helper';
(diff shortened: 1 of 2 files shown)


sfmskywalker (approved):
Thanks for that!

Don't worry about the other issue; we are working on replacing the entire designer canvas with a new implementation that will solve the overlapping connection issue.

tomy2105:
> Don't worry about the other issue; we are working on replacing the entire designer canvas with a new implementation that will solve the overlapping connection issue.

OK, thanks for the info.
Does this mean designer with visually look different or just the implementation will be different?

sfmskywalker:
The designer will look noticeably different, but the principles remain the same: each activity will have outcomes from which you can connect to other activities. A crucial difference will be the fact that the new implementation draws the nodes directly to the canvas, while the current implementation only draws the connections using SVG if I'm not mistaken. The activities themselves are HTML-based, but in dagre-d3, they will be rendered using d3 objects.

tomy2105:
OK, thanks for the info.

I guess the only remaining question I have is when can we expect this "new" designer to be available/finished?

sfmskywalker:
No problem.

We are optimistic that the new designer will be available somewhere between the end of next week and the end of the current month.

If that happens, odds are we can create an Elsa 2.0 release candidate release where we only work on bugs & docs. It won't have the full set of features that was originally planned, but at the very least we can continuously add them as they become available.

PR #1052, opened 2021-05-27
BREAKING: Update MongoDB provider with support for object derivatives

### LINKED ISSUE #552: Bug: Elsa 2.0 MongoDb Deserializing

sfmskywalker:
Fixes #552

### DIFF
--- a/src/persistence/Elsa.Persistence.MongoDb/Serializers/JObjectSerializer.cs
+++ b/src/persistence/Elsa.Persistence.MongoDb/Serializers/JObjectSerializer.cs
@@ -1,0 +1,37 @@
+using System;
+using MongoDB.Bson.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NodaTime;
+using NodaTime.Serialization.JsonNet;
+
+namespace Elsa.Persistence.MongoDb.Serializers
+{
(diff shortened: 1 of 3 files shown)


Juandavi1:
Hi guys, 

i have the same issue 

`
[15:57:10 ERR] flow-engine 06/25/2021 20:57:10 500 Internal Server Error System.FormatException: An error occurred while deserializing the Output property of class Elsa.Models.WorkflowInstance: Unknown discriminator value 'FinishOutput'.
 ---> MongoDB.Bson.BsonSerializationException: Unknown discriminator value 'FinishOutput'.
`

sfmskywalker:
Can you create a separate issue for this so we can track? Thank you.

PR #1067, opened 2021-05-31
Conductor

sfmskywalker:
This PR adds a new module in the form of `Elsa.Activities.Conductor`, which offers the following features:

1. The ability to send commands to and receive events from external applications.
2. Three new activities that enable communication between Elsa and your application: **SendCommand**, **EventReceived** and **RunTask**.
3. The ability to declare supported commands, events and tasks via configuration. These commands, events and tasks then become available as preconfigured activities.

This is an initial version - further enhancements might follow later, such as:

* The ability to define commands, events and tasks from the Elsa Dashboard.
* Define multiple applications.
* Customize the "broker" between Elsa and your application. Right now it's always using HTTP REST, but new strategies will be added such as in-memory communication (in case your Elsa Host needs to handle instructions), gRPC and SignalR to communicate directly with browser applications.


jamshally:
A Rebus "broker" would be good also. It seems it is already used within Elsa, allows for a number of different transports to be easily configured, and provides very easy hosting options.

sfmskywalker:
Yep, agreed.

Issue #753, opened 2021-03-12
Enhancement for Azure Service Bus Activities


jdevillard:
_Originally posted by @sfmskywalker in https://github.com/elsa-workflows/elsa-core/issues/749#issuecomment-797483211_

I propose the following step/idea to update the Azure Service Bus Activities : 

- [ ] Implement opt-in/out to check / create service bus entity and give a clear Exception Message
- [ ] Enhance Message Model
- [ ] Allow ability to select custom Serializer
- [ ] Add capacity to configure Azure Storage Fallback for large message
- [ ] Check new Azure.Messaging.ServiceBus library

feel free to propose other ideas , I can take the dev on theses differents points, submit PR and update this issue

PR #1084, opened 2021-06-02
Azure Service Bus Enhancement - Add Fire And Forget Option

jdevillard:
This PR is a the first of Azure Service Bus Enhancement #753  , that allow to use Fire and Forget for the Topic Sender.

This allow to wait before sending the Message to the Service Bus. 
With this, we can wait for the next suspend and use a Service Bus Received Activity to receive the response. If not setting FireAndForget to false, we can have side effect and receive the message using the Topic Worker before the BookMark with Correlation Id was correctly set.

The default value is True to avoid breaking change (sent directly)

### DIFF
--- a/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
+++ b/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
@@ -78,14 +78,12 @@ namespace Elsa.Activities.AzureServiceBus
                 message.CorrelationId = context.WorkflowExecutionContext.CorrelationId;
 
             var sender = await GetSenderAsync();
+            
             if(SendMessageOnSuspend)
                 return Combine(Done(), new ServiceBusActionResult(sender, message));
-            else
-            {
-                await sender.SendAsync(message);
-                return Done();
-            }
-
+            
+            await sender.SendAsync(message);
+            return Done();
         }
 
         protected Message CreateMessage(Object input)
--- a/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
+++ b/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
@@ -69,6 +69,9 @@ namespace Elsa.Activities.AzureServiceBus
         [ActivityInput(Category = PropertyCategories.Advanced, DefaultSyntax = SyntaxNames.Json, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid, SyntaxNames.Json }, UIHint = ActivityInputUIHints.MultiLine)]
         public IDictionary<string, object>? UserProperties { get; set; } = new Dictionary<string, object>();
 
+        [ActivityInput(Category = PropertyCategories.Advanced,Hint = "If set, the message will be send to the Service Bus only after Workflow Suspend, usefull for Request/Response pattern", DefaultValue = false)]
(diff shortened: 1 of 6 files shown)


sfmskywalker:
Interestingly enough, I just pushed a change that prevents the race condition you are describing. Perhaps you can try and see if that works? Because it would be better if the user doesn't even have to be concerned with setting "fire and forget" to false to prevent the race condition.

Please let me know. Thanks!

sfmskywalker:
It's this code that prevents the race condition: https://github.com/elsa-workflows/elsa-core/commit/cdaa105f3e75e255a1331b9c0817a0f926fd87bf#diff-327721ff436b8f80f7444da4c470c3f9259d2ffb37836fe6e1258f5fccdfac11R54-R56

jdevillard:
ok, so using this simple test : 
```
    public class SbFlow : IWorkflow

    {
        {
            builder
                    .StartWith<HttpEndpoint>(httpEndpoint => httpEndpoint
                    .WithPath("/testSb")
                    .WithMethod("POST")
                    .WithReadContent(true)
                    .WithTargetType<String>()
                ).WithName("HttpRequest")

                .WriteLine(context => { 
                    context.CorrelationId = Guid.NewGuid().ToString("n");
                    return $"Received: {JsonConvert.SerializeObject(context.Input)}";
                    })
                .SendTopicMessage("testtopic1", Guid.NewGuid().ToString())
                .TopicMessageReceived<string>("testtopic1","testsub")
                .WriteLine("Receive Message from ")
                .WriteHttpResponse(HttpStatusCode.NotFound, "Hello World!", "text/html");
        }
    }
```
this doesn't seems to fix the issue
in fact when I receive a message in the Topic Message Received : I can have no pending workflows as shown here :
image

and here is the log :

image

we already discuss about this before, I don't already check the usage of the new class (Launchpad, InstanceExecutor and LockingWorkflow*) but the fact is : There is no link between the Azure Service Bus Sender .SendMessage() and the Workflow Engine
so the message can even be sent before the SendTopicActivity was finished.
That why you talk me about the ActivityExecutionResult class

src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/SendAzureServiceBusTopicMessage.cs
```diff
@@ -22,6 +23,9 @@ public SendAzureServiceBusTopicMessage(ITopicMessageSenderFactory messageSenderF
 
         [ActivityInput] public string TopicName { get; set; } = default!;
         [ActivityInput] public object Message { get; set; } = default!;
+        
+        [ActivityInput(DefaultValue = true, Hint = "Allow the activity to send immediatly the message if true or wait for the first suspend. Usefull for RequestResponse pattern using Topic Message Received " )] 
+        public bool FireAndForget { get; set; } = true;
```
sfmskywalker, line 28: As [you suggested on Discord](https://discord.com/channels/814605913783795763/814605914212270101/849946099367018517), let's move this to the "Advanced" category.

jdevillard:
Side effect is resolve with this commit 97df3b7bc7f951eb94965df6d53adee18679f053

So PR can be close

jdevillard:
Reopen this issue as after new test, change doesn't seem to fix.

jdevillard:
This PR is old and need to be updated to latest master or do you prefer to create a new PR?

sfmskywalker:
Up to you - whichever you find easier.

jdevillard:
(I've force push to be on the latest master) I've add a unit test to check. If the test is good, we have to change from **suspend workflow -> execute scheduled tasks -> index bookmarks** to **suspend workflow -> index bookmarks -> execute scheduled tasks**. Then I'll take the old commit to add the property (fire&forget) in the activity to allow the delay or not.

jdevillard:
PR is fine now. 

@sfmskywalker  you can merge when possible for you.

sfmskywalker (approved):

PR #1085, opened 2021-06-03
add capability to use MessageModel as the input of the Azure Service Bus Sender

jdevillard:
This PR give the ability for the user to specify a MessageModel object as the input of the Activity.

This allow the user to customize a lot of property of the Brocker Service Message like SystemProps and User Props.

Today this can be use in Code and we can think about new Activity like CreateServiceBusMessage to have UI to create this kind of object using the Designer

### DIFF
--- a/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
+++ b/src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
@@ -28,33 +28,33 @@ namespace Elsa.Activities.AzureServiceBus
         [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid, SyntaxNames.Json })]
         public object Message { get; set; } = default!;
 
-        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        [ActivityInput(Category = PropertyCategories.Advanced, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
         public string? CorrelationId { get; set; }
-        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        [ActivityInput(Category = PropertyCategories.Advanced, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
         public string ContentType { get; set; } = default!;
-        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        [ActivityInput(Category = PropertyCategories.Advanced, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
         public string? Label { get; set; }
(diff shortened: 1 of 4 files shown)


sfmskywalker:
I'm wondering if it wouldn't be better to simply expose the message properties directly from the UI (under the Advanced tab)? Although it's clever what you did, it kinda mixes the "payload" input (currently named "Message" on the activity) with the metadata.

So perhaps you can simply model the `SendAzureServiceBusQueueMessage` and `SendAzureServiceBusTopicMessage` activities after the Azure Service Bus `Message` model by replicating each property.

Then each property is available from the UI and of course from the fluent API.

And to prevent too much copy & paste of these properties, maybe you can add a base class activity called e.g. `AzureServiceBusActivity`.

Thoughts?

jdevillard:
ok, let's go for this.

Just, about the Key/Value pair in user props , what UI is used for this?

jdevillard:
Base class was added with metadata mapping, SendAzureServiceBusQueueMessage and SendAzureServiceBusTopicMessage was simplified.

src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
```diff
@@ -0,0 +1,100 @@
...
+        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        public DateTime ExpiresAtUtc { get; set; }
+        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        public TimeSpan TimeToLive { get; set; }
+        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        public string? ReplyToSessionId { get; set; }
+        [ActivityInput(SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })] 
+        public DateTime ScheduledEnqueueTimeUtc { get; set; }
+        [ActivityInput(DefaultSyntax = SyntaxNames.Json, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid,SyntaxNames.Json },UIHint = ActivityInputUIHints.MultiLine)] 
+        public IDictionary<string,Object> UserProperties { get; set; } = new Dictionary<string, object>();
```
sfmskywalker, line 58: Should we move these to the Advanced category?

sfmskywalker (approved):

src/activities/Elsa.Activities.AzureServiceBus/Activities/SendAzureServiceBusMessage/AzureServiceBusSendActivity.cs
jdevillard: yep 👍 

done!

sfmskywalker:
> Just, about the Key/Value pair in user props , what UI is used for this?

We don't have an appropriate UI for that, but what will work for the time being is using the `ActivityInputUIHints.MultiLine` UI with a `DefaultSyntax = SyntaxNames.Json` combination (and `SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid }`) - exactly like what we do for the `RequestHeaders` property on `SendHttpRequest`.

The user can then enter key/value pairs with JSON syntax like this:

```json
{
   "key1": "value1",
   "key2": "value2",
   //...etc.
}
```

Issue #1146, opened 2021-06-17
Enhancement to ISetupActivity and SetupActivity


tomy2105:
Related to #1122.

I wrote a custom activity using IActivityTypeProvider that has variable number of properties (depending on some external input obtained during startup). This activity has custom name and what is important its properties are not directly mapped to C# class properties.

Now I wanted to invoke this activity from CompositeActivity but I could not because ISetupActivity  only provides methods that allow you to set activity properties that are mapped to C# class properties.

I managed to get around it by casting ISetupActivity   to SetupActivity  and directly manipulating SetupActivity ValueProviders dictionary. 

        public override void Build(ICompositeActivityBuilder builder)
        {
            builder
                .ThenTypeNamed(typeof(SetFlag).FullName, setup =>
                {
                    var trueSetup = (SetupActivity<IActivity>)setup;
                    trueSetup.ValueProviders[FlagActivityTypeProvider.INTERNAL_ID_NAME] = context =>
                    {
                        return new ValueTask<object?>(Flags[FlagActivityTypeProvider.INTERNAL_ID_NAME]);
                    };
                }, null);
        }

This is definitely not good programming, so I suggest for an enhancement to SetupActivity and ISetupActivity with method with following signature/implementation:

        public ISetupActivity Set<TProperty>(string propertyName, Func<ActivityExecutionContext, ValueTask<TProperty?>> valueFactory)
        {
            ValueProviders[propertyName] = async context => await valueFactory(context);
            return this;
        }

In fact current Set method in SetupActivity could be moved to extensions and implemented using this one

        public ISetupActivity<T> Set<TProperty>(this ISetupActivity<T> builder,Expression<Func<T, TProperty?>> propertyAccessor, Func<ActivityExecutionContext, ValueTask<TProperty?>> valueFactory) => builder.Set(propertyAccessor.GetProperty(),Name, valueFactory);

sfmskywalker:
Not bad at all. Would you mind sending a PR with this?

PR #1166, opened 2021-06-23
Enabling directly named properties used using SetupActivity (not just through C# property name)

tomy2105:
This solves #1146.

I needed to leave existing methods in ISetupActivity and SetupActivity because when I tried to move them to SetupActivityExtensions it started complaining that `propertyAccessor.GetProperty()` is not accessible/defined.

So I'll leave that to someone more clever than I am :)......


sfmskywalker (approved):
Thanks for this :)

PR #1186, opened 2021-06-28
Workflow priority queues

sfmskywalker:
This PR adds the ability to associate a workflow definition with a "channel". A channel can be used by workflow dispatcher implementations to select an appropriate queue to send a message to for the workflow to execute, effectively enabling **prioritization of workflow definitions**. 

By default, all workflows are associated with the "Default" channel, but you can configure your own set of channels using the new `ConfigureWorkflowChannels` method on `ElsaOptionsBuilder`.

See the following example:

**appsettings.json**

```json
{
  "Elsa": {
    "WorkflowChannels": {
      "Channels": [
        "High",
        "Normal",
        "Low"
      ],
      "Default": "Normal"
    },
}
```

**Startup.cs**

```csharp
services
    .AddElsa(elsa => elsa
        .ConfigureWorkflowChannels(options => elsaSection.GetSection("WorkflowChannels").Bind(options))
    );
```

To associate a workflow definition with a channel using the workflow builder API, use the `WithChannel` method. For example:

```csharp
public class MyWorkflow : IWorkflow
{
   {
      builder.WithChannel("Low");
   }
}
```

When creating workflows with the designer, you can select the desired channel from the Workflow Settings modal:

image


johnwc:
@sfmskywalker can you explain how you configure the channels in the backend, as in how the system knows that channel "High" is higher priority than channel "Low".

sfmskywalker:
Sure. So actually, the backend doesn't know what channel has a higher priority and do anything with it. This implementation was an initial attempt to solve an issue with limited throughput by allowing some critical workflows to use a less crowded queue.

A more intelligent implementation might have involved specifying a numerical value to the priority definition and provide strategies to determine when to schedule what workflow to what queue, for example. None of that has happened (yet).

johnwc:
Ok, so channels has no functionality as of today?

sfmskywalker:
It should still allow you to assign a workflow definition to channel, but that’s pretty much the extent of it, yes.

Issue #1130, opened 2021-06-15
Http Endpoint faulted flow returns 200 with no response content to caller


vlad-crisan:
I have defined a workflow that is started by exposing an endpoint through "Http Endpoint" activity, with method POST.
Following that, if the workflow faults at any point, the caller receives response 200, and no content body (header Content-Length: 0)
 
Navigating to Workflow Instances shows the proper details, exception message, stack trace, status "Faulted", etc.

I would have expected the caller to receive all of these details, instead of getting a 200.
Is there some additional configuration that can be done to acchieve this behavior? Or should the default be like I described it - get the exception details in the response?

I'm using version 2.0.0.159 of the library.
The way I forced an exception to happen in the flow:
- custom activity that explicitly throws an exception
- input property to an activity set to javascript and defining a faulty expression

both of these result in the same behavior - 200 and no response content

I'm working at the moment on an application that intends to proxy to another service that runs the elsa workflows (and exposes them by having an HTTP Endpoint step at the start of each flow),  so it's important to have the ability to get information about the faulty behavior.

Thanks in advance for any feedback!

sfmskywalker:
Makes sense to at the very least respond with an appropriate status code, i.e. 500 when the workflow faults.

For production, it would be unsafe to send back a full stack trace, but we should make this configurable in some way.

PR #1218, opened 2021-07-06
http endpoint respond with 500 for faulted workflow

vlad-crisan:
having a fault in a workflow started through http endpoint returns 500
only in case the response has not alerady been started


vlad-crisan:
fixes #1130

vlad-crisan:
@sfmskywalker ideas why docker login fails?

sfmskywalker:
I haven’t looked into that yet. But it works when run from master, so it’s not a big problem for now.

sfmskywalker (approved):

Issue #1223, opened 2021-07-06
Don#t use colon in appsettings


jwillmer:
Kubernetes does not support colons (`:`) in environment variables and I also have a problem in docker-compose.

Current appsettings.json
```
...
  "Elsa": {
    "Features": {
      "Persistence:EntityFrameworkCore:Sqlite": {
        "Enabled": true,
        "ConnectionStringName": "Sqlite"
      },
      "Console": true,
...
```

If I like to override `Console` in appsettings via env variable I can do it one of two ways:
```
-e Elsa__Features__Console="false"
-e Elsa:Features:Console="false"
```

But if I like to override `Persistence:EntityFrameworkCore:Sqlite` I have a problem because docker-compose treats `:` and `__` as delimiters. I haven't found any escape for it yet. For Kubernetes it is even worse since it does not allow `:` at all in the environment settings as described in this issue: [Environment variables containing colon not allowed](https://github.com/kubernetes/kubernetes/issues/53201)

Axeleron007:
I suggest to remove colons from (an example) "Persistence:EntityFrameworkCore:Sqlite" and just name it as "PersistenceEntityFrameworkCoreSqlite". If everyone is happy it will be a quick and cheap fix.

jwillmer:
doesn't help. there are more options with colons!

sfmskywalker:
@jwillmer Maybe I misunderstood, but @Axeleron007 said he will remove all colons from all options. That should help, right?

PR #1225, opened 2021-07-07
Don#t use colon in appsettings #1223

Axeleron007:
Colons are removed from any element name in the appsettings


Axeleron007:
Hi Sipke, can you please review and merge if everything is fine? Thanks

sfmskywalker (approved):

PR #1229, opened 2021-07-07
Add ITenantAccessor interface to provide custom ability for acquiring tenantid

GreenShadeZhang:
Mainly add the ability of tenantid customization for API projects, and add sample projects

### DIFF
--- a/src/core/Elsa.Core/Activities/Signaling/Services/Signaler.cs
+++ b/src/core/Elsa.Core/Activities/Signaling/Services/Signaler.cs
@@ -15,13 +15,11 @@ namespace Elsa.Activities.Signaling.Services
 
         private readonly IWorkflowLaunchpad _workflowLaunchpad;
         private readonly ITokenService _tokenService;
-        private readonly ITenantAccessor _tenantAccessor;
 
-        public Signaler(IWorkflowLaunchpad workflowLaunchpad, ITokenService tokenService, ITenantAccessor tenantAccessor)
+        public Signaler(IWorkflowLaunchpad workflowLaunchpad, ITokenService tokenService)
         {
             _workflowLaunchpad = workflowLaunchpad;
             _tokenService = tokenService;
-            _tenantAccessor = tenantAccessor;
         }
 
         public async Task<IEnumerable<CollectedWorkflow>> TriggerSignalTokenAsync(string token, object? input = default, CancellationToken cancellationToken = default)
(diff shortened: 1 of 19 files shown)


src/samples/aspnet/Elsa.Samples.CustomTenantIdSource/CustomTenantAccessor.cs
```diff
@@ -0,0 +1,29 @@
...
+            _accessor = accessor;
+        }
+        public Task<string> GetTenantIdAsync(CancellationToken cancellationToken = default)
+        {
+            var id = "elsa-core";
+
+            //You can customize the data
+            var httpContext = _accessor.HttpContext;
+
+            return Task.FromResult(id);
```
sfmskywalker, line 26: Since you injected `IHttpContextAccessor`, how about actually reading the tenant ID from the HTTP context? Could be query string or HTTP header. For example:

```csharp
var httpContext = _accessor.HttpContext;
var tenantId = httpContext.Request.Headers["x-tenant"]; 

return Task.FromResult(tenantId);
```

This would make for a more real-world example I think.
Otherwise, you may just as well not inject `IHttpContextAccessor` and return "elsa-core" directly as the tenant ID.

src/core/Elsa.Core/Activities/Signaling/Services/Signaler.cs
```diff
@@ -15,11 +15,13 @@ public class Signaler : ISignaler
 
         private readonly IWorkflowLaunchpad _workflowLaunchpad;
         private readonly ITokenService _tokenService;
+        private readonly ITenantAccessor _tenantAccessor;
 
-        public Signaler(IWorkflowLaunchpad workflowLaunchpad, ITokenService tokenService)
+        public Signaler(IWorkflowLaunchpad workflowLaunchpad, ITokenService tokenService, ITenantAccessor tenantAccessor)
```
sfmskywalker, line 20: I think we may need to be careful here. Since the `ITenantAccessor` service is potentially HTTP-Context bound, while this `Signaler` service could be used from e.g. background tasks, we would not have access to any HTTP context.

For this reason, the `IWorkflowLaunchPade.CollectAndExecuteWorkflowsAsync` method accepts a `tenantId` as an argument.

Therefore it's perhaps safest for now to also accept a `tenantId` argument on the `TriggerSignalAsync` method, and let the caller decide how they are going to provide a tenant Id.

Alternatively, we might redesign the persistence access layers to take on a dependency on `ITenantAccessor` so that none of the application services need to know anything about tenant IDs. This would have my absolute preference.

But then you do need to consider again scenarios where there may not be an HTTP context such as when code executes in background tasks such as Timers.

Another scenario are background workers such as Azure Service Bus, Rebus and MassTransit message consumers - these messages might carry the tenant ID as a header value.

In those cases, if the user installed an HTTP context-based tenant accessor, it won't work.

As you can see, there are a lot of considerations to make.
So perhaps the safest thing to do (for now) is to pass tenant IDs explicitly as method parameters, and let "entry points" such as HTTP middleware, HTTP controllers, message handlers, background tasks etc. figure out where to get the tenant ID from.

sfmskywalker (changes requested):

sfmskywalker (approved):

PR #1243, opened 2021-07-11
Workaround for double quotes

Axeleron007:
Workaround for double quotes. Given that based on Jen's investigation only variables wrapped with double quotes are allowed in the project environment we need to escape additional double quotes.

Sipke, can you do a code review and merge if it is fine with you please?


Axeleron007:
As discussed with Sipke the workaround does not resolve the main issue on the appsettings override.

PR #1297, opened 2021-07-21
Center start activity when loading workflow

### LINKED ISSUE #1155: Designer focus when loaded
When workflow is loaded in desinger it focuses on upper left corner of the workflow with default zoom.

In cases where workflow looks like a tree of activities, it often branches out from start so at the bottom it is much wider than at the top.

In this case designer will show empty part of canvas and you have to zoom out or scroll to find your blocks which is utterly confusing.

I think it would be good if at load time designer would always focus on start block.

tomy2105:
This resolves #1155. My humble apologies, I know it has Elsa 2.3 mark on it but I couldn't wait :)......

The start activity is now placed in the upper middle of visible area when workflow is loaded.


sfmskywalker:
Well, as the saying goes: “good things come to those whom can’t wait”.

Or maybe it was something else 😉

I’m merging this. Thanks!

Issue #1254, opened 2021-07-13
Default value for input property and output property storage per each input/output


tomy2105:
In Elsa 2.1 one can choose whether input/output value is persisted or not, default value is set for whole Elsa using  `services.AddElsa(elsa => elsa.UseDefaultWorkflowStorageProvider<TransientWorkflowStorageProvider>());`

Would it be possible that for custom activities and their properties you can specify initial value of storage flag (default, instance, transient, blob) as part of **each** activity property definition (ActivityInput/ActivityOutput attribute)? And same for properties created by IActivityProvider?

Moreover, would it be possible to add another flag which would indicate whether this control is even displayed in UI (so end workflow developer cannot change it :))?

sfmskywalker:
Yes, all possible. And probably not too hard to do either :))

PR #1337, opened 2021-07-31
Refactor storage

sfmskywalker:
This PR fixes issues #1254 and #1304


sfmskywalker:
@tomy2105 Please be aware of a small breaking change introduced with this PR with regards to the Activity Type Provider API.

The `Describe` property has been renamed to `DescribeAsync` and expects a return type of `ValueTask<ActivityDescriptor>`.
Since this is an advanced feature, I don't think it's worth making this change non-breaking, especially since it's an easy thing to fix (if not, let me know).

PR #1338, opened 2021-07-31
Disable config access

### LINKED ISSUE #1327: Expressions: Config access exposes vulnerability when working with user supplied expressions
Hello again 👋

**Project context**

We're building a user-facing workflow feature into our product and using Elsa as the execution engine. Thus, any workflow definitions and any expressions contained therein are user supplied and untrusted. The operations that the workflow engine can take on behalf of the user is strictly contained to just the actions they have permission to take in their group (we use a relatively straightforward group-based permission model).

**Vulnerability**

Because workflow definitions are user supplied and untrusted, the fact that [expressions can reference](https://elsa-workflows.github.io/elsa-core/docs/expressions/expressions-liquid#configuration) [.Net configuration](https://elsa-workflows.github.io/elsa-core/docs/expressions/expressions-javascript#getconfig) poses a security risk for us since .Net configuration tends to contain secrets (connection strings, auth headers, etc).

I've been able to disable config access in both javascript and liquid expressions but this was a surprising default to me. Exposing carte-blanche to .Net configuration by default seems dangerous. Although it could be that our use case for Elsa isn't aligning completely with what Elsa was designed for.

**Proposed Action Steps**

I have three ideas for how to improve this (ordered by preference)

1. My ideal solution is that this would be an opt-in feature. I'm sure there is a use case for this, although I wonder if it's ubiquitous enough to warrant the security risk in a context like this. Although I realize that would be a breaking change.
2. It would be nice if there were a convenient way to reliably disable this behavior (e.g. via [ElsaOptionsBuilder](https://github.com/elsa-workflows/elsa-core/blob/2ec915dabed9ffd03e9c9101e95a04f2438c1ed7/src/core/Elsa.Core/ElsaOptionsBuilder.cs#L18) - `elsa.DisableConfigurationAccess()` for instance).
3. At minimum, it would be nice to see better documentation on how to workaround it. My solution for JavaScript was relatively straightforward, following the [documentation here](https://elsa-workflows.github.io/elsa-core/docs/extensibility/extensibility-javascript), I was able to override the `getConfig` function to throw an `InvalidOperationException`. Disabling config access in Liquid took some digging. In order to know what type to override in the `MemberAccessStrategy`, I had to look at [the source code](https://github.com/elsa-workflows/elsa-core/blob/master/src/scripting/Elsa.Scripting.Liquid/Handlers/ConfigureLiquidEngine.cs#L61). Looking at source code isn't the end of the world, although it would be nice to have an officially recognized way to disable this behavior.

With my current familiarity with the project, I'd be willing to tackle any one of these approaches (although that first idea might be best saved for Elsa 3).

josephdangerstewart:
Fixes #1327

**After this change:**
- Liquid's [`{{Configuration}}`](https://elsa-workflows.github.io/elsa-core/docs/expressions/expressions-liquid#configuration) variable will now only work by opting into it using the service collection extension method `EnableLiquidConfigurationAccess`
- JavaScript's [`getConfig`](https://elsa-workflows.github.io/elsa-core/docs/expressions/expressions-javascript#getconfig) function will now only work by opting into by setting `options.EnableConfigurationAccess = true` within the `WithJavaScriptOptions` service collection extension method
- The JavaScript expression handler now defaults to denying CLR access, you can enable this by setting `options.AllowClr = true` within the `WithJavaScriptOptions` service collection extension method


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1338) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1338) before we can accept your contribution.<br/><br/>:x: josephdangerstewart [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1338)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=1338) it.</sub>

josephdangerstewart:
Ah, looks like I might need to contact my employer before signing the CLA, that'll have to wait until Monday

test/integration/Elsa.Core.IntegrationTests/Scripting/JavaScript/JavaScriptExpressionsIntegrationTests.cs
```diff
@@ -25,6 +31,97 @@ public async Task RunningAWorkflowThatIncludesJsonStringifyAParsedObjectShouldNo
...
+            {
+                builder.StartWith<AddExpressionToActivityState>(setup =>
+                {
+                    setup.Set(x => x.Expression, _ => "getConfig('SomeSecret')");
+                    setup.Set(x => x.ExpressionSyntax, _ => "JavaScript");
+                });
+            }
+        }
+
+        private class AddExpressionToActivityState : Activity
```
josephdangerstewart, line 98: I'm planning on adding this to the shared project later and using in the Liquid test. I'd also like to add an integration test for Clr access

josephdangerstewart:
I just realized, I also should update the docs

src/scripting/Elsa.Scripting.JavaScript/Options/ScriptOptions.cs
```diff
@@ -2,11 +2,20 @@ namespace Elsa.Scripting.JavaScript.Options
 {
     public class ScriptOptions
     {
-        public ScriptOptions()
-        {
-            AllowClr = true;
```
josephdangerstewart, line 7: Is defaulting this to false to major of a breaking change? I'm wondering if we should leave these options as opt out (but strongly encouraged to opt out) until the next major release

src/scripting/Elsa.Scripting.Liquid/Services/LiquidHandler.cs
```diff
@@ -33,7 +33,7 @@ public LiquidHandler(ILiquidTemplateManager liquidTemplateManager, IMediator med
 
         private async Task<TemplateContext> CreateTemplateContextAsync(ActivityExecutionContext activityExecutionContext, CancellationToken cancellationToken)
         {
-            var context = new TemplateContext();
+            var context = new TemplateContext(new TemplateOptions());
```
josephdangerstewart, line 36: This was causing test failures when the Liquid tests were run all together, my *guess* is that by default, Fluid mutates some static options object. So one test (the one asserting that config access can be enabled) adds the `Configuration` keyword and it persists into the next test causing it to fail. This ensures that getting a unique options object (and thus a unique member access strategy) for every test run. I don't think it would have any impact on consumers though

src/scripting/Elsa.Scripting.JavaScript/Options/ScriptOptions.cs
sfmskywalker: Perhaps not. From the [Jint documentation](https://github.com/sebastienros/jint#accessing-net-assemblies-and-classes) it sounds like we can just turn it off - there's no need to allow access to arbitrary objects from the System namespace. At least, I don't think we're doing that. Of course someone might be doing this, but then they will need to opt-in (and we need to add a note to the next release.

Let's just try it.

src/scripting/Elsa.Scripting.Liquid/Services/LiquidHandler.cs
sfmskywalker: Interesting. I agree, looks harmless enough.

sfmskywalker:
> Fixes #1327
> 
> I'm planning on adding tests for this although here's what I got so far, I thought it'd be good to get a WIP up to get buy in on the new options.

I'll try and take a look over the next coming days. In the meantime, do let me know whenever you think the PR is ready for a final review.

josephdangerstewart:
> In the meantime, do let me know whenever you think the PR is ready for a final review.

Ah my bad, I didn't update the description. I think this should be ready for a final review as is. It looks like the relevant documentation is in a separate repo so I'll open up a PR there when this is shipped

josephdangerstewart:
> I'll try and take a look over the next coming days.

Okay sweet, take your time! There isn't any rush on our end, as I mentioned in the issue, we already have a sufficient workaround for the Config access.

sfmskywalker (approved):
With tests and everything. Really nice!

PR #1362, opened 2021-08-06
Fixed problems regarding NCLOB columns

akghasemi:


src/persistence/Elsa.Persistence.EntityFrameworkCore/DbContexts/OracleContext.cs
```diff
@@ -12,5 +13,92 @@ public class OracleContext:ElsaContext
...
+                    .HasColumnType("NCLOB");
+
+                entity.Property(x => x.DisplayName)
+                    .HasColumnType("NCLOB");
+
+                entity.Property(x => x.Name)
+                    .HasColumnType("NCLOB");
+
+
+
```
sfmskywalker, line 32: Any chance I can get you to remove the excessive whitespace (throughout this file)? 😃

sfmskywalker (changes requested):

sfmskywalker:
@akghasemi Thank you for your PR! But I'm not very familiar with Oracle and best practices. I understand that `NCLOB` represents a data type that allows you to store strings of up to 4GB according to [Oracle FAQ's](https://www.orafaq.com/wiki/NCLOB). But what is the issue it fixes?

akghasemi:
> @akghasemi Thank you for your PR! But I'm not very familiar with Oracle and best practices. I understand that `NCLOB` represents a data type that allows you to store strings of up to 4GB according to [Oracle FAQ's](https://www.orafaq.com/wiki/NCLOB). But what is the issue it fixes?



> @akghasemi Thank you for your PR! But I'm not very familiar with Oracle and best practices. I understand that `NCLOB` represents a data type that allows you to store strings of up to 4GB according to [Oracle FAQ's](https://www.orafaq.com/wiki/NCLOB). But what is the issue it fixes?

Hi,

akghasemi:
Hi,the probelem is that the ODP.net for EF core doesn't map the string column to the NCLOB column by default,in SQL server if you have column type nvarchar(max) or nvarchar(2000) the query is being built by framework would be the same but in Oracle when you have NCLOB and nvarchar columns the query would be different ,so if you explicitly don't specify that this column is NCLOB then it is possible when data that you want save in the column got bigger you recieve an Error :  “ORA-01460: unimplemented or unreasonable conversion requested”
[https://docs.oracle.com/en/database/oracle/oracle-data-access-components/19.3/odpnt/EFCoreLargeCharacter.html#GUID-AE805872-88C9-4D67-95B5-1700B5539518](url)
this error happened to me several times and there are two solutions in Oracle documentations which I think the second one that I have implemented is better

src/persistence/Elsa.Persistence.EntityFrameworkCore/DbContexts/OracleContext.cs
akghasemi: sure ,no problem :)

sfmskywalker:
Got it. Thank you for the explanation!

sfmskywalker (approved):

PR #1512, opened 2021-09-10
Add page size control and some sanity checks, fix one off in pager

### LINKED ISSUE #1504: Enable choosing page size in dashboard
Currently dashboard displays 15 items per page by default.
You can change page size by manually changing parameters in the URL but would be nicer if there were a dropdown where this could be chosen.

In addition there should be some logic (probably both on client and server side) to prevent too large (or negative) page sizes.
Currently if you manually type negative page size you get all items and "negative" garbage in summary ("From 0 to -1 of 20results").
And if you enter too large page size (like 1000) something breaks (empty page displayed) provided there are more than 1000 items available......

tomy2105:
This fixes #1504.

Added page size dropdown.
Added bunch of sanity checks on page nad page size in case user enters wrong numbers in url.
Fixed one off error in pager summary.
Fixed small bug in package.json for dev build.

### DIFF
--- a/src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-list/elsa-workflow-instance-list-screen/elsa-workflow-instance-list-screen.tsx
+++ b/src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-list/elsa-workflow-instance-list-screen/elsa-workflow-instance-list-screen.tsx
@@ -26,6 +26,11 @@ import Tunnel from "../../../../data/dashboard";
   shadow: false,
 })
 export class ElsaWorkflowInstanceListScreen {
+  static readonly DEFAULT_PAGE_SIZE = 15;
+  static readonly MIN_PAGE_SIZE = 5;
+  static readonly MAX_PAGE_SIZE = 100;
+  static readonly START_PAGE = 0;
+
   @Prop() history?: RouterHistory;
   @Prop() serverUrl: string;
   @Prop() basePath: string;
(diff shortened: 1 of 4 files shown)


src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-list/elsa-workflow-instance-list-screen/elsa-workflow-instance-list-screen.tsx
```diff
@@ -26,6 +26,11 @@ import Tunnel from "../../../../data/dashboard";
   shadow: false,
 })
 export class ElsaWorkflowInstanceListScreen {
+  static readonly DEFAULT_PAGE_SIZE = 15;
+  static readonly MIN_PAGE_SIZE = 5;
+  static readonly MAX_PAGE_SIZE = 100;
+  static readonly START_PAGE = 0;
```
sfmskywalker, line 32: I admire that you put these readonly statics here, but do they really need to scream at me? 😅 
Maybe `DefaultPageSize` etc. is just as good?

sfmskywalker (approved):
Very nice contribution. Thank you!

src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-list/elsa-workflow-instance-list-screen/elsa-workflow-instance-list-screen.tsx
tomy2105: Apologies on this, switching too many project and languages combined with too many style guides.
So I was unintentionally screaming at you with style guide that mandates constants to be ALL UPPERCASE :) (so it IS immediately visible they don't change over program execution).
Feel free to convert it to any style you want :).

sfmskywalker: Completely understand, no worries 😃

PR #1513, opened 2021-09-10
Elsa 2.0/fileactivities

MilkyWare:
Hi,

The main addition is the creation of a FileSystemWatcher trigger activity using the native System.IO FileSystemWatcher. It includes the ability to select which change events to listen to (create, change, delete, renamed) as well as the attributes to watch for changes for (filename, lastwrite, size etc).

Similar to the AzureServiceBus activities, programmatically defined watcher workflows are queried at startup to create the appropriate watchers. 

Samples have also been included to demonstrate the watcher.

Also, a few fixes and updates. File activity required parameters have the correct attribute and default values are now specified in the ActivityInput attribute.


sfmskywalker (approved):
Very cool. Thank you, also for providing the sample 😃

PR #1542, opened 2021-09-16
Manual workflow trigger

Axeleron007:


src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx
```diff
@@ -213,6 +217,8 @@ export class ElsaWorkflowDesigner {
     eventBus.on(EventTypes.PasteActivity, this.onPasteActivity);
     eventBus.on(EventTypes.HideModalDialog, this.onCopyPasteActivityEnabled);
     eventBus.on(EventTypes.ShowWorkflowSettings, this.onCopyPasteActivityDisabled);
+    eventBus.on(EventTypes.TestActivityMessageReceived, this.onTestActivityMessageReceived);
+    eventBus.on(EventTypes.ActivityJsonSchemaUpdated, this.onActivityJsonSchemaUpdate);
```
sfmskywalker, line 221: All the testing-related code needs to be moved outside of the workflow designer and instead be made part of the workflow definition editor. It's OK to make the designer extensible as needed of course.

src/server/Elsa.Server.Api/Endpoints/Workflows/Test.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+using Elsa.Models;
+
+namespace Elsa.Server.Api.Endpoints.Workflows
+{
+    [ApiController]
+    [ApiVersion("1")]
+    [Route("v{apiVersion:apiVersion}/workflows/{workflowDefinitionId}/{version}/{signalRConnectionId}/test")]
+    [Produces("application/json")]
+    public class Test : Controller
```
sfmskywalker, line 16: This class looks very similar to `Execute`, the main difference being the presence of the SignalR connection ID. Can you merge these two classes into one, perhaps making the SignalR connection ID optional to indicate that it needs to be setup?

sfmskywalker (changes requested):

src/core/Elsa.Core/Services/Workflows/WorkflowRunner.cs
```diff
@@ -254,39 +254,48 @@ private async ValueTask RunCoreAsync(WorkflowExecutionContext workflowExecutionC
...
+                        await _mediator.Publish(new WorkflowExecutionBurstCompleted(workflowExecutionContext, activityExecutionContext), cancellationToken);
 
-                activityOperation = Execute;
+                    activityOperation = Execute;
+                }
+                catch (Exception e)
+                {
+                    await _mediator.Publish(new ActivityExecutionResultFailed(e, activityExecutionContext), cancellationToken);
+                    throw;
+                }
```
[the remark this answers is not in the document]
Axeleron007: Catching exceptions from the Test endpoint would catch exceptions only when a test started but would not catch exceptions when the http endpoint is triggered in browser.

src/designer/elsa-workflows-studio/src/models/events.ts
```diff
@@ -32,7 +32,9 @@ export const EventTypes = {
   WorkflowRegistryUpdated: 'workflow-registry.updated',
   ClipboardPermissionDenied: 'clipboard-permission-denied',
   ClipboardCopied: 'clipboard-copied',
-  PasteActivity: 'paste-activity'
+  PasteActivity: 'paste-activity',
+  TestActivityMessageReceived: 'test-activity-message-received',
+  ActivityJsonSchemaUpdated: 'activity-json-schema-updated'
```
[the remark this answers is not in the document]
Axeleron007: Since the schema update code moved to elsa-workflow-definition-editor-screen component we do not need this event type at all.

src/designer/elsa-workflows-studio/src/services/elsa-client.ts
```diff
@@ -114,6 +114,16 @@ export const createElsaClient = function (serverUrl: string): ElsaClient {
         return response.data;
       }
     },
+    workflowTestApi: {
+      execute: async (request) => {
+        httpClient.interceptors.request.use(function (config) {
+          config.headers['Accept'] = 'application/json';
+          config.data = Object.assign({}, config.data, {});
+          return config;
+        })
```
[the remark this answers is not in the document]
Axeleron007: Indeed, this is an obsolete code left there after some testing. Cleaned up.

src/server/Elsa.Server.Api/Endpoints/Workflows/Test.cs
Axeleron007: That one was original endpoint and later moved to it's separate WorkflowTest/Execute endpoint. Deleted.

src/scripting/Elsa.Scripting.JavaScript/Services/JsonTypeScriptDefinitionService.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using AutoMapper.Internal;
+using Elsa.Models;
+using Elsa.Scripting.JavaScript.Events;
+using MediatR;
+
+namespace Elsa.Scripting.JavaScript.Services
+{
+    public class JsonTypeScriptDefinitionService : ITypeScriptDefinitionService
```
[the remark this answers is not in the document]
Axeleron007: JsonTypeScriptDefinitionService  and DotNetTypeScriptDefinitionService removed as they have been replaced with providers

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
```diff
@@ -75,7 +75,11 @@ public class HttpEndpointMiddleware
                 return;
             }
 
-            var workflowBlueprint = await workflowRegistry.FindAsync(x => x.IsPublished && x.Id == pendingWorkflowInstance.DefinitionId && !x.IsDisabled, cancellationToken);
+            var isTestRun = pendingWorkflowInstance.MetaData.FirstOrDefault(x => x.Key == "isTestRun").Value;
+            var workflowBlueprint = (isTestRun != null && Convert.ToBoolean(isTestRun)) ?
+                await workflowRegistry.FindAsync(x => x.Id == pendingWorkflowInstance.DefinitionId, cancellationToken) :
+                await workflowRegistry.FindAsync(x => x.IsPublished && x.Id == pendingWorkflowInstance.DefinitionId && !x.IsDisabled, cancellationToken);
+
```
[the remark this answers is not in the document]
Axeleron007: I think I would need your more detailed suggestions on how I may improve it.

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
```diff
@@ -75,7 +75,11 @@ public class HttpEndpointMiddleware
                 return;
             }
 
-            var workflowBlueprint = await workflowRegistry.FindAsync(x => x.IsPublished && x.Id == pendingWorkflowInstance.DefinitionId && !x.IsDisabled, cancellationToken);
+            var isTest = pendingWorkflowInstance.MetaData.FirstOrDefault(x => x.Key == "isTest").Value;
+            var workflowBlueprint = (isTest != null && Convert.ToBoolean(isTest)) ?
+                await workflowRegistry.FindAsync(x => x.Id == pendingWorkflowInstance.DefinitionId, cancellationToken) :
+                await workflowRegistry.FindAsync(x => x.IsPublished && x.Id == pendingWorkflowInstance.DefinitionId && !x.IsDisabled, cancellationToken);
```
sfmskywalker, line 81: Similar to my other comment, this code will make testing work for HTTP endpoint activities, but not for all of the other trigger activities - they too try to get a workflow blueprint, but right now they are not aware of any "isTest" metadata. They will fail to find any unpublished workflow blueprints.

A neat solution to this problem would be to implement a decorator around the `IWorkflowRegistry` service that exists in the `Elsa.Server.Api` project as well (since this should be the only library containing test-related code).

The decorator should override a new method to be created on `IWorkflowRegistry` called `FindActiveAsync` (representing the "singular" sibling of `ListActiveAsync` that already exists and returns a list of results).

The overridden method would then simply replicate what is done in the decorated service, but include the "isTest" logic to include blueprints under test even if they are not published (like you're doing now).

src/core/Elsa.Abstractions/Models/WorkflowInstance.cs
```diff
@@ -33,13 +34,24 @@ public WorkflowInstance()
...
+        public IDictionary<string, object?> MetaData { get; set; } = new Dictionary<string, object?>();
 
         public HashSet<BlockingActivity> BlockingActivities
         {
             get => _blockingActivities;
             set => _blockingActivities = new HashSet<BlockingActivity>(value, BlockingActivityEqualityComparer.Instance);
         }
 
+        public object? GetMetaData(string key) { 
+            return MetaData.FirstOrDefault(x => x.Key == key).Value;
```
sfmskywalker, line 46: Although we won't ever notice this is less efficient, the correct way to access values from a dictionary by key is by actually leveraging the dictionary instead of a linear lookup.

Like this:

```csharp
public object? GetMetaData(string key) => MetaData.TryGetValue(key, out var value) ? value : default;
```

src/core/Elsa.Abstractions/Models/WorkflowInstance.cs
```diff
@@ -33,13 +34,24 @@ public WorkflowInstance()
...
+        }
+
+        public void SetMetaData(string key, object? value)
+        {
+            MetaData.Remove(key);
+            MetaData.Add(key, value);
```
sfmskywalker, line 52: This can be simplified (and optimized) as follows:

```csharp
MetaData[key] = value;
```

src/core/Elsa.Abstractions/Models/WorkflowInstance.cs
```diff
@@ -33,13 +34,24 @@ public WorkflowInstance()
         public WorkflowInputReference? Input { get; set; }
         public WorkflowOutputReference? Output { get; set; }
         public IDictionary<string, IDictionary<string, object?>> ActivityData { get; set; } = new Dictionary<string, IDictionary<string, object?>>();
+        public IDictionary<string, object?> MetaData { get; set; } = new Dictionary<string, object?>();
```
sfmskywalker, line 37: "Metadata" is one word, so the capitalization should be: `Metadata` (notice lowercase 'd'). This should be applied everywhere.

src/core/Elsa.Abstractions/Services/Models/StartableWorkflowDefinition.cs
```diff
@@ -1,4 +1,4 @@
 namespace Elsa.Services.Models
 {
-    public record StartableWorkflowDefinition(IWorkflowBlueprint WorkflowBlueprint, string? ActivityId, string? CorrelationId = default, string? ContextId = default);
+    public record StartableWorkflowDefinition(IWorkflowBlueprint WorkflowBlueprint, string? ActivityId, string? CorrelationId = default, string? ContextId = default, bool? isTest = default, string? signalRConnectionId = default);
```
sfmskywalker, line 3: Core models shouldn't know about testing. Please remove both `isTest` and `signalRConnectionId`.

You can set these properties on the `WorkflowInstance` from within the `WorkflowTest.Execute` class after you received the `StartableWorkflow` object (and before you actually execute).

Keep in mind: the core library should not know anything about testing.

src/core/Elsa.Abstractions/Services/Workflows/IWorkflowFactory.cs
```diff
@@ -12,6 +12,8 @@ public interface IWorkflowFactory
             string? correlationId = default,
             string? contextId = default,
             string? tenantId = default,
+            bool? isTest = default,
+            string? signalRConnectionId = default,
```
sfmskywalker, line 16: Remove

src/core/Elsa.Abstractions/Services/Workflows/IWorkflowLaunchpad.cs
```diff
@@ -29,11 +29,16 @@ public interface IWorkflowLaunchpad
         /// </summary>
         Task<StartableWorkflow?> FindStartableWorkflowAsync(string workflowDefinitionId, string? activityId = default, string? correlationId = default, string? contextId = default, string? tenantId = default, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Creates a new test workflow instance with execution pending for the specified workflow blueprint using the specified starting activity ID.
+        /// </summary>
+        Task<StartableWorkflow?> FindStartableWorkflowAsync(string workflowDefinitionId, int version, string? activityId = default, string? correlationId = default, string? contextId = default, string? tenantId = default, bool? isTest = default, string? signalRConnectionId = default, CancellationToken cancellationToken = default);
+
```
sfmskywalker, line 36: Remove.

src/core/Elsa.Abstractions/Services/Workflows/IWorkflowLaunchpad.cs
```diff
@@ -29,11 +29,16 @@ public interface IWorkflowLaunchpad
...
         /// <summary>
         /// Creates a new workflow instance with execution pending for the specified workflow blueprint using the specified starting activity ID.
         /// </summary>
-        Task<StartableWorkflow?> FindStartableWorkflowAsync(IWorkflowBlueprint workflowBlueprint, string? activityId = default, string? correlationId = default, string? contextId = default, string? tenantId = default, CancellationToken cancellationToken = default);
-        
+        Task<StartableWorkflow?> FindStartableWorkflowAsync(IWorkflowBlueprint workflowBlueprint, string? activityId = default, string? correlationId = default, string? contextId = default, string? tenantId = default, bool? isTest = default, string? signalRConnectionId = default, CancellationToken cancellationToken = default);
+
```
sfmskywalker, line 41: Revert.

src/core/Elsa.Core/Services/Workflows/WorkflowFactory.cs
```diff
@@ -23,6 +23,8 @@ public WorkflowFactory(IClock clock, IIdGenerator idGenerator)
             string? correlationId = default,
             string? contextId = default,
             string? tenantId = default,
+            bool? isTest = default,
+            string? signalRConnectionId = default,
```
sfmskywalker, line 27: Remove.

src/core/Elsa.Core/Services/Workflows/WorkflowFactory.cs
```diff
@@ -36,9 +38,12 @@ public WorkflowFactory(IClock clock, IIdGenerator idGenerator)
                 ContextId = contextId,
                 CreatedAt = _clock.GetCurrentInstant(),
                 Variables = new Variables(workflowBlueprint.Variables),
-                ContextType = workflowBlueprint.ContextOptions?.ContextType ?.GetContextTypeName()
+                ContextType = workflowBlueprint.ContextOptions?.ContextType?.GetContextTypeName()
             };
 
+            workflowInstance.MetaData.Add("isTest", isTest);
+            workflowInstance.MetaData.Add("signalRConnectionId", signalRConnectionId);
```
sfmskywalker, line 45: Remove. Instead, do this from the `WorkflowTest.Execute` class.

src/core/Elsa.Core/Services/Workflows/WorkflowLaunchpad.cs
```diff
@@ -96,7 +96,26 @@ public async Task<IEnumerable<StartableWorkflow>> FindStartableWorkflowsAsync(Wo
...
+
+        public async Task<StartableWorkflow?> FindStartableWorkflowAsync(
+            string workflowDefinitionId,
+            int version,
+            string? activityId,
+            string? correlationId = default,
+            string? contextId = default,
+            string? tenantId = default,
+            bool? isTest = default,
+            string? signalRConnectionId = default,
```
sfmskywalker, line 110: Remove

src/core/Elsa.Core/Services/Workflows/WorkflowLaunchpad.cs
```diff
@@ -105,14 +124,16 @@ public async Task<IEnumerable<StartableWorkflow>> FindStartableWorkflowsAsync(Wo
             string? correlationId = default,
             string? contextId = default,
             string? tenantId = default,
+            bool? isTest = default,
+            string? signalRConnectionId = default,
```
sfmskywalker, line 128: Remove.

src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-script-property/elsa-script-property.tsx
```diff
@@ -44,8 +47,49 @@ export class ElsaScriptProperty {
...
+          const errors = validate.errors;
+          if (errors != null)
+            this.isJsonSchemaValid = false;
+        }
+        catch (e){
+          const err = e;
+          this.isJsonSchemaValid = false;
+        }        
+      }
+    }
```
sfmskywalker, line 92: This code should be part of the HttpEndpointPlugin class, since it's the only activity that has the "Schema" property.

src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-script-property/elsa-script-property.tsx
```diff
@@ -88,7 +133,26 @@ export class ElsaScriptProperty {
...
+      {isSchema ?
+        <div class="elsa-mt-1">
+          <a href="#"
+              onClick={e => this.onConvertToJsonSchemaClick(e)}
+              class="elsa-relative elsa-inline-flex elsa-items-center elsa-px-4 elsa-py-2 elsa-border elsa-border-gray-300 elsa-text-sm elsa-leading-5 elsa-font-medium elsa-rounded-md elsa-text-gray-700 elsa-bg-white hover:elsa-text-gray-500 focus:elsa-outline-none focus:elsa-shadow-outline-blue focus:elsa-border-blue-300 active:elsa-bg-gray-100 active:elsa-text-gray-700 elsa-transition elsa-ease-in-out elsa-duration-150">
+              Convert to Json Schema
+          </a>
+        </div>
+        : undefined
+      }
```
sfmskywalker, line 155: Keep in mind that this script property component is reused by other activities as wel, so this validation that is specific to the "Schema" property doesn't belong here.

src/designer/elsa-workflows-studio/src/components/screens/workflow-definition-editor/elsa-workflow-definition-editor-screen/elsa-workflow-definition-editor-screen.tsx
```diff
@@ -381,19 +411,112 @@ export class ElsaWorkflowDefinitionEditorScreen {
...
+  async updateHttpEndpointJsonSchema(request: WorkflowTestUpdateHttpEndpoint) {
+    let workflowModel = {...this.workflowModel};
+    const activity = workflowModel.activities.find(x => x.activityId == request.activityId);
+    if (activity)
+    {
+      const activityModel = activity.properties.find(x => x.name == 'Schema');
+      activityModel.expressions[SyntaxNames.Literal] = request.jsonSchema;
+      eventBus.emit(EventTypes.UpdateActivity, this, activityModel);
+    }
+  };
```
sfmskywalker, line 461: This code is specific to the HTTP Endpoint activity, so you should find a way to decouple the workflow definition editor from this logic that is specific to the HTTP Endpoint by leveraging the HTTP Endpoint plugin.

src/designer/elsa-workflows-studio/src/plugins/default-drivers-plugin.ts
```diff
@@ -12,7 +12,7 @@ export class DefaultDriversPlugin implements ElsaPlugin {
     this.addDriver('checkbox', () => new CheckboxDriver());
     this.addDriver('dropdown', () => new DropdownDriver());
     this.addDriver('multi-text', () => new MultiTextDriver());
-    this.addDriver('code-editor', () => new CodeEditorDriver());
+    this.addDriver('code-editor', () => new CodeEditorDriver());
```
sfmskywalker, line 15: Revert.

src/samples/server/Elsa.Samples.Server.Host/Startup.cs
```diff
@@ -87,9 +88,13 @@ public void ConfigureServices(IServiceCollection services)
...
                 .AddElsaSwagger();
 
+            services.AddJavaScriptTypeDefinitionProvider<CustomerTypeDefinitionProvider>();
+            services.AddJavaScriptTypeDefinitionProvider<ClientTypeDefinitionProvider>();
+
             // Allow arbitrary client browser apps to access the API for demo purposes only.
             // In a production environment, make sure to allow only origins you trust.
-            services.AddCors(cors => cors.AddDefaultPolicy(policy => policy.AllowAnyHeader().AllowAnyMethod().AllowAnyOrigin().WithExposedHeaders("Content-Disposition")));
+            //services.AddCors(cors => cors.AddDefaultPolicy(policy => policy.AllowAnyHeader().AllowAnyMethod().AllowAnyOrigin().WithExposedHeaders("Content-Disposition")));
+            services.AddCors(cors => cors.AddPolicy("ClientPermission", policy => policy.AllowAnyHeader().AllowAnyMethod().WithOrigins("http://localhost:3333").AllowCredentials().WithExposedHeaders("Content-Disposition")));
```
sfmskywalker, line 97: Can we change this back to allow **any** origin?

src/samples/server/Elsa.Samples.Server.Host/Startup.cs
```diff
@@ -102,10 +107,16 @@ public void Configure(IApplicationBuilder app)
...
             app
-                .UseCors()
+                //.UseCors()
+                .UseCors("ClientPermission")
                 .UseElsaFeatures()
                 .UseRouting()
-                .UseEndpoints(endpoints => { endpoints.MapControllers(); });
+                .UseEndpoints(endpoints => { 
+                    endpoints.MapControllers();
+                    endpoints.MapHub<WorkflowTestHub>("/hubs/workflowTest");
```
sfmskywalker, line 116: It would be good to provide this as an extension method from the `Elsa.Server.Api` project called e.g. `MapWorkflowTestHub()`. Might be easier to discover for users and allows us to change the path in the future.

src/scripting/Elsa.Scripting.JavaScript/Elsa.Scripting.JavaScript.csproj
```diff
@@ -13,8 +13,14 @@
         <PackageTags>elsa, workflows, javascript, jint</PackageTags>
     </PropertyGroup>
 
+    <ItemGroup>
+      <Compile Remove="Services\DotNetTypeScriptDefinitionService.cs" />
+      <Compile Remove="Services\JsonTypeScriptDefinitionService.cs" />
+    </ItemGroup>
+
```
sfmskywalker, line 20: Remove.

src/scripting/Elsa.Scripting.JavaScript/Elsa.Scripting.JavaScript.csproj
```diff
@@ -13,8 +13,14 @@
...
     <ItemGroup>
         <PackageReference Include="Jint" Version="3.0.0-beta-2032" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
sfmskywalker, line 23: This reference can be removed.

src/server/Elsa.Server.Api/Models/WorkflowTestMessage.cs
```diff
@@ -0,0 +1,14 @@
+namespace Elsa.Server.Api.Models
+{
+    public class WorkflowTestMessage
+    {
+        public string? Path { get; set; }
```
sfmskywalker, line 5: What's this used for?

sfmskywalker (changes requested):

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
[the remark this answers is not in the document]
sfmskywalker: Sure, please find a little bit more details [here](https://github.com/elsa-workflows/elsa-core/pull/1542#discussion_r717875722).

src/server/Elsa.Server.Api/Models/WorkflowTestMessage.cs
Axeleron007: This one is used to generate the full path that can be copied/pasted to a browser window or Postman to execute suspended HttpEndpoint.

src/samples/server/Elsa.Samples.Server.Host/Startup.cs
Axeleron007: Fixed

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
Axeleron007: I think we may not need a new decorator. If we have this code in WorkflowRegistry:
`
        public async Task<IWorkflowBlueprint?> FindActiveAsync(Func<IWorkflowBlueprint, bool> predicate, CancellationToken cancellationToken) =>
            (await ListActiveAsync(cancellationToken).Where(predicate).OrderByDescending(x => x.Version)).FirstOrDefault();

        private async IAsyncEnumerable<IWorkflowBlueprint> ListActiveInternalAsync([EnumeratorCancellation] CancellationToken cancellationToken)
        {
            var workflows = await ListAsync(cancellationToken);

            foreach (var workflow in workflows)
            {
                // If a workflow is not published, only consider it for processing if it has at least one non-ended workflow instance.
                if (!workflow.IsPublished && !await WorkflowHasUnfinishedWorkflowsAsync(workflow, cancellationToken))
                    continue;

                yield return workflow;
            }
        }
`
it works fine for IsTest scenario too if we change the code in HttpEndpointMiddleware.cs to that one:
`
var workflowBlueprint = await workflowRegistry.FindActiveAsync(x => x.Id == pendingWorkflowInstance.DefinitionId && !x.IsDisabled, cancellationToken);
`

Axeleron007:
Hi Sipke, can you please review the changes you requested.

src/server/Elsa.Server.Api/Models/WorkflowTestMessage.cs
sfmskywalker: OK.

The issue with that is that `Path` is an attribute that specifically belongs to the HTTP Endpoint activity, while `WorkflowTestMessage` is activity type-agnostic (it shouldn't care about specific types of activities).

In other words, this is tight-coupling HTTP endpoint aspects with this workflow test model.

Imagine if Elsa didn't ship with the HTTP Endpoint activity but instead provided it as a separate package. How would you solve this problem?

Similarly, there might already exist 3rd party activities that offer their own version of HTTP Endpoints. Elsa should offer them a way to somehow display the test URL.

If you look at it from that perspective, you'll see it's necessary to handle this in a more extensible way.

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
sfmskywalker: But isn't the problem with that change, that now even un-published workflows will be invoked?

src/core/Elsa.Abstractions/Services/Workflows/IWorkflowRegistry.cs
```diff
@@ -36,5 +36,10 @@ public interface IWorkflowRegistry
         /// Returns a single workflow blueprint that fits the specified predicate. 
         /// </summary>
         Task<IWorkflowBlueprint?> FindAsync(Func<IWorkflowBlueprint, bool> predicate, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Returns a single workflow blueprint that fits the specified predicate. 
+        /// </summary>
+        Task<IWorkflowBlueprint?> FindActiveAsync(Func<IWorkflowBlueprint, bool> predicate, CancellationToken cancellationToken = default);
```
sfmskywalker, line 43: I'm not sure it makes much sense for a method that finds active workflows to still require a predicate, because when you want to find "active" workflows, that implies a predicate already: the user wants to find active workflows.

I suggest removing the predicate argument. If the user needs a more specific subset, they should use the other overload that accepts a predicate, giving the user full control over the query.

src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-script-property/elsa-script-property.tsx
```diff
@@ -58,6 +79,23 @@ export class ElsaScriptProperty {
     const fieldLabel = propertyDescriptor.label || propertyName;
     const fieldHint = propertyDescriptor.hint;
     const value = this.currentValue;
+    const isSchema = fieldName === "Schema";
```
sfmskywalker, line 82: "Schema" is a property specific to HTTP Endpoint, which means it doesn't belong here.

src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-script-property/elsa-script-property.tsx
```diff
@@ -89,6 +127,17 @@ export class ElsaScriptProperty {
...
       <input type="hidden" name={fieldName} value={value}/>
+      {renderValidationResult()}
+      {isSchema ?
+        <div class="elsa-mt-3">
+          <a href="#"
+              onClick={e => this.onConvertToJsonSchemaClick(e)}
+              class="elsa-relative elsa-inline-flex elsa-items-center elsa-px-4 elsa-py-2 elsa-border elsa-border-gray-300 elsa-text-sm elsa-leading-5 elsa-font-medium elsa-rounded-md elsa-text-gray-700 elsa-bg-white hover:elsa-text-gray-500 focus:elsa-outline-none focus:elsa-shadow-outline-blue focus:elsa-border-blue-300 active:elsa-bg-gray-100 active:elsa-text-gray-700 elsa-transition elsa-ease-in-out elsa-duration-150">
+              Convert to Json Schema
+          </a>
+        </div>
```
sfmskywalker, line 138: This portion should be provided by the HTTP Endpoint plugin as well, since it is the owner of the Schema property.

src/designer/elsa-workflows-studio/src/components/screens/workflow-definition-editor/elsa-workflow-definition-editor-screen/elsa-workflow-definition-editor-screen.tsx
```diff
@@ -451,10 +569,153 @@ export class ElsaWorkflowDefinitionEditorScreen {
...
+    const value = message.data["Body"];
+    let workflowModel = {...this.workflowModel};
+    const activityModel = workflowModel.activities.find(x => x.activityId == message.activityId);
+
+    const context: ActivityUpdatedContext = {
+      activityModel: activityModel,
+      data: JSON.stringify(convert(value), null, 1)
+    };
+    eventBus.emit(EventTypes.ActivityPluginUpdated, this, context);
+  }
```
sfmskywalker, line 587: The workflow editor shouldn't have any awareness of the Schema stuff. That's all specific to the HTTP endpoint activity.

src/designer/elsa-workflows-studio/src/plugins/http-endpoint-plugin.tsx
```diff
@@ -20,4 +22,53 @@ export class HttpEndpointPlugin implements ElsaPlugin {
...
+          isValid = false;
+      }
+      catch (e){
+        const err = e;
+          isValid = false;
+      }
+    }
+
+    context.isValidated = true;
+    context.isPositiveResult = isValid;
```
sfmskywalker, line 71: A better, more commonly used name would be `isValid`.

sfmskywalker (changes requested):

src/activities/Elsa.Activities.Http/Middleware/HttpEndpointMiddleware.cs
sfmskywalker: As just discussed over skype, the concern I raised previously is not actually a problem. As it turns out, the HTTP Middleware already requires a custom bit of logic to load the workflow blueprint. For other trigger sources such as Timer, they simply schedule work for the corresponding workflow instance, which in turn get a specific workflow blueprint by version.

Sorry for the confusion.

sfmskywalker (approved):

sfmskywalker (approved):

PR #1568, opened 2021-09-23
Multi-targeting Elsa.Activities.Http for .NET Core 3.1 and .NET 5

mohdali:
Since Elsa.Activities.Http added reference to Microsoft.AspNetCore.Authorization in 2.3.0, it is causing a runtime error in .NET Core 3.1.

Root cause is .NET Core 3.1 is using it's own Policy assembly instead. Which causes a null exception during the runtime due a change in behavior in the Authroization package for 5.0.

More details below:

https://docs.microsoft.com/en-us/answers/questions/222841/error-while-downgrading-project-from-net-5-to-net.html

Source code of Authorization packages for cross checking:

https://github.com/dotnet/aspnetcore/blob/v3.1.19/src/Security/Authorization/Policy/src/PolicyServiceCollectionExtensions.cs
https://github.com/dotnet/aspnetcore/blob/v5.0.10/src/Security/Authorization/Policy/src/PolicyServiceCollectionExtensions.cs

https://github.com/dotnet/aspnetcore/blob/v3.1.19/src/Security/Authorization/Core/src/AuthorizationServiceCollectionExtensions.cs
https://github.com/dotnet/aspnetcore/blob/v5.0.10/src/Security/Authorization/Core/src/AuthorizationServiceCollectionExtensions.cs

mainly the changed behavior is:

.NET Core 3.1
`
if (configure != null)
            {
                services.Configure(configure);
            }
`

.NET 5
`
 if (configure == null)
            {
                throw new ArgumentNullException(nameof(configure));
            }
`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 image: Previous Visual Studio 2019
 
 environment:
-    elsa_version: 2.3.0
+    elsa_version: 2.4.0
     elsa_build_version: 0.0.0
     nodejs_version: "14.6.0"
     # AppVeyor installs Java on its build agents by default, but it
@@ -25,7 +25,7 @@ init:
                 $env:elsa_build_version="$($env:elsa_version)-preview.$($env:APPVEYOR_BUILD_NUMBER)"
             }
 
-version: 2.3.0.{build}
+version: 2.4.0.{build}
 
 pull_requests:
     do_not_increment_build_number: true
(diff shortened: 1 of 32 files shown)


sfmskywalker (approved):

shykln:
Thanks for fixing this. Looking forward to releasing it in 2.3.x as this is breaking .NetCore 3.1 users.

vladimir-ilnytskyi:
@sfmskywalker 
Could you please tell when this will be released on nuget.org? Waiting for fix =) thanks a lot!

PR #1576, opened 2021-09-27
persistence: added Azure DocumentDB support

lucapisano:
Hi Sipke,
I've coded support for Azure DocumentDB persistence (alias CosmosDB)

Can you please review implementation and accept PR if it's ok?
Thanks,
Luca

### DIFF
--- a/src/activities/Elsa.Activities.Telnyx/Activities/CallBackRingGroup.cs
+++ b/src/activities/Elsa.Activities.Telnyx/Activities/CallBackRingGroup.cs
@@ -230,7 +230,7 @@ namespace Elsa.Activities.Telnyx.Activities
                 });
         }
 
-        protected override async ValueTask OnExitAsync(ActivityExecutionContext context, object? output)
+        protected override async ValueTask OnExitAsync(ActivityExecutionContext context, object? output, IList<string> outcomes)
         {
             // Hang up any pending calls.
             context.JournalData.Add("Exiting", true);
--- a/src/activities/Elsa.Activities.Telnyx/Activities/CallRingGroup.cs
+++ b/src/activities/Elsa.Activities.Telnyx/Activities/CallRingGroup.cs
@@ -377,7 +377,7 @@ namespace Elsa.Activities.Telnyx.Activities
                     ifFalse => ifFalse.Finish(TelnyxOutcomeNames.NoResponse));
         }
 
-        protected override async ValueTask OnExitAsync(ActivityExecutionContext context, object? output)
+        protected override async ValueTask OnExitAsync(ActivityExecutionContext context, object? output, IList<string> outcomes)
         {
             await CancelPendingCallsAsync(context);
             
--- a/src/core/Elsa.Abstractions/Services/Workflows/CompositeActivity.cs
+++ b/src/core/Elsa.Abstractions/Services/Workflows/CompositeActivity.cs
@@ -77,7 +77,7 @@ namespace Elsa.Services
                 output = finishOutput.Output;
             }
             
-            await OnExitAsync(context, output);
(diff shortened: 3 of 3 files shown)


sfmskywalker:
I like that you used EF Core to implement a DocumentDB provider.
Would it be possible to reuse the `Elsa.Persistence.EntityFramework.Core` project for the DB context and provide DocumentDB provider similar to the way it was done for SQL Server, MySQL etc.?

lucapisano:
I've tried to do so initially, but there're issues in using Elsa.Persistence.EntityFramework.Core due to it's relational orientation.This is why I've succeeded in implementing it by writing a provider that is "standalone" with respect to Elsa.Persistence.EntityFramework.CoreCertified Multi Cloud ArchitectIl 27 set 2021 13:52, Sipke Schoorstra ***@***.***> ha scritto:
I like that you used EF Core to implement a DocumentDB provider.
Would it be possible to reuse the Elsa.Persistence.EntityFramework.Core project for the DB context and provide DocumentDB provider similar to the way it was done for SQL Server, MySQL etc.?

src/persistence/Elsa.Persistence.DocumentDb/Services/DatabaseRegister.cs
```diff
@@ -0,0 +1,38 @@
...
+            }
+
+            return true;
+        }
+
+        private static void RegisterSerializers()
+        {
+        }
+    }
+}
```
sfmskywalker, line 38: Do we need to keep this class? It doesn't look like it's doing anything special.

sfmskywalker:
> there're issues in using Elsa.Persistence.EntityFramework.Core due to it's relational orientation

I see. I'm wondering if we could find a way to make it work by perhaps making some changes in the core project if necessary. It would save a lot of duplication.

I can merge this in and leave it as an exercise for later, but perhaps you can give me a quick pointer as to what causes the issue with the relational orientation? At first glance, the custom DB context doesn't appear to be that different compared to the core one.

src/persistence/Elsa.Persistence.DocumentDb/Services/DatabaseRegister.cs
lucapisano: no, sorry, I forgot it :D

lucapisano:
> > there're issues in using Elsa.Persistence.EntityFramework.Core due to it's relational orientation
> 
> I see. I'm wondering if we could find a way to make it work by perhaps making some changes in the core project if necessary. It would save a lot of duplication.
> 
> I can merge this in and leave it as an exercise for later, but perhaps you can give me a quick pointer as to what causes the issue with the relational orientation? At first glance, the custom DB context doesn't appear to be that different compared to the core one.

I've not done a complete assessment but please have a look at:

- RunMigrations.ExecuteAsync() EF Core throws exception stating that the method MigrateAsync can only be used with relational databases.
- builder.ToContainer(nameof(...)); in every IEntityTypeConfiguration. DocumentDB needs a property to set the container name

Maybe it's possible to overcome those issues creating a method similar to .IsMySql()

lucapisano:
Hi @sfmskywalker 
Any news on how to proceed here?

sfmskywalker:
Someone will need to take a look to see if there isn’t a way to make this work with the existing DB context and Related code. If it’s absolutely impossible or impossibly awkward to do, then I’ll gladly take this PR. But given the amount of code and code duplication I prefer to spend a little time to investigate. It would be absolutely neat if the Cosmos DB provider can work with EF Core in a similar way that the other DB providers do without having to have its own DB context.

Vendetta26:
Duplicate of #

PR #1588, opened 2021-09-28
WorkflowSettings for JavaScript vocabulary

constantine-v:
Extended javascript vocabulary with workflow settings.

### DIFF
--- a/src/modules/workflowsettings/Elsa.WorkflowSettings/Services/WorkflowSettingsManager.cs
+++ b/src/modules/workflowsettings/Elsa.WorkflowSettings/Services/WorkflowSettingsManager.cs
@@ -23,7 +23,7 @@ namespace Elsa.WorkflowSettings.Services
 
             var value = new WorkflowSetting();
 
-            foreach (var provider in providers)
+            foreach (var provider in providers.OrderByDescending(x => x.Priority))
             {
                 var providerValue = await provider.GetWorkflowSettingAsync(workflowBlueprintId, key, cancellationToken);
                 if (providerValue.Value != null)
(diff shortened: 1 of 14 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1588) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1588) before we can accept your contribution.<br/><br/>:x: constantine-v [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=1588)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=1588) it.</sub>

src/modules/workflowsettings/Elsa.WorkflowSettings/Services/WorkflowSettingsManager.cs
```diff
@@ -32,5 +34,23 @@ public async Task<WorkflowSetting> LoadSettingAsync(string workflowBlueprintId,
...
+            var providers = _workflowSettingsProviders;
+
+            var value = new List<WorkflowSetting>();
+
+            foreach (var provider in providers)
+            {
+                var providerValue = await provider.GetWorkflowSettingsAsync(workflowBlueprintId, cancellationToken, orderBy, paging);
+                if (providerValue != null)
+                {
+                    value = providerValue.ToList();
```
sfmskywalker, line 49: Instead of overriding the result of the last evaluated provider, you should replace existing values.
It should work similar to the way configuration works in .NET: you can have appsettings.json and environment variables.

Depending on which provider was registered last, takes precedence (in case both providers define the same setting key).

So what you might do instead is build a dictionary of key/setting pairs and simply set the setting into the dicitonary by key.

If the entry didn't exist before, it will be added. Otherwise it will be overridden.

Example:

```csharp
var providers = _workflowSettingsProviders;

            var dictionary= new Dictionary<string, WorkflowSetting>();

            foreach (var provider in providers.OrderByDescending(x => x.Priority) // Let's introduce a priority so we can control what providers take precedence.
            {
                // Since you return a list, it should never be null; always return a list even if empty.
                var settings = await provider.GetWorkflowSettingsAsync(workflowBlueprintId, cancellationToken: cancellationToken);

                foreach(var setting in settings)
                    dictionary[setting.Name] = setting;
            }

            return dictionary.Values;
```

sfmskywalker (changes requested):

src/modules/workflowsettings/Elsa.WorkflowSettings/Services/WorkflowSettingsManager.cs
constantine-v: Introduced priority property and refactored method according to example.

sfmskywalker (approved):

PR #2445, opened 2021-11-02
Fixes #2417

### LINKED ISSUE #2417: Intelisense proposes invalid properties
I get the `activities` path displayed on all levels. Same with some other paths that should only be proposed on specific levels. I guess it is all the same problem so fixing `activities` should solve the others as well.

image

constantine-v:
Removing allowJs option does not seem to be having any negative sides and it fixes the issue when properties form already entered expression always appear in intellisense (e.g. for "return test.prop." intellisense will suggest both "test" and "prop" as valid options even when they are not).


sfmskywalker:
Nice find!

PR #2450, opened 2021-11-04
Fix for oracle persistence provider

akghasemi:
when data is bigger than 2000 chars the columns of database should be NCLOB


sfmskywalker (approved):
Thanks!

PR #2494, opened 2021-11-18
Fix for semaphore issue in AzureServiceBus activities

constantine-v:
Moved return statement inside try block, otherwise finally block which releases semaphore could sometimes not be executed, causing semaphore to hold lock indefinitely.


sfmskywalker (approved):
Great catch!

Issue #2531, opened 2021-11-26
RuntimeSelectListProviderSettings dont work anymore


noxe:
Hi @sfmskywalker - in the latest preview - IRuntimeSelectListItemsProvider dont work anymore - in my activites no more drop down is rendered. i see the Interface is obsolete - changing it to IRuntimeSelectListProvider does not work either. 

i use this in my activity

        public object GetOptions(PropertyInfo property) => new RuntimeSelectListItemsProviderSettings(this.GetType());

from commit https://github.com/elsa-workflows/elsa-core/commit/ed327c3f9554679c38277c8c4aa0e695b7f6d133 i see a lot has changed here - i also see an RuntimeSelectListProviderSettings in the tsx files - but i dont see such class on the "server" ?

noxe:
@sfmskywalker  i see this in the console in case this helps?

![image](https://user-images.githubusercontent.com/158478/143677401-08c55f22-5d46-4b53-aea5-a827b27adb1e.png)

noxe:
Just a guees - but here  https://github.com/elsa-workflows/elsa-core/blob/ed327c3f9554679c38277c8c4aa0e695b7f6d133/src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-dropdown-property/elsa-dropdown-property.tsx#L50

shouldnt that be this.selectList.items instead of this.selectList ?

noxe:
created an PR :)

PR #2532, opened 2021-11-27
Fixes DropDown runtime select list

noxe:
fixes https://github.com/elsa-workflows/elsa-core/issues/2531 

using items instead selectlist


Vendetta26:
gh pr checkout 2532

Vendetta26:
requests

src/designer/elsa-workflows-studio/src/components/editors/properties/elsa-dropdown-property/elsa-dropdown-property.tsx
```diff
@@ -48,7 +48,7 @@ export class ElsaDropdownProperty {
     const fieldId = propertyName;
     const fieldName = propertyName;
     let currentValue = this.currentValue;
-    const {items} = this.selectList;
+    const { items } = this.selectList.items;
```
sfmskywalker, line 51: This should not work - if it does, it means the contents of the `this.selectList` contains an unexpected value (an object containing `items` which itself then also has an `items` field.

I just pushed an update that seems to fix the runtime select list items story (I tried it with a sample custom activity with runtime items implemented).

Please give it a go and let me know if it works or not.

Thanks!

noxe: ok - just did a compare with other implementations - and this was the only where this.selectlist was used - the other editor are using selectlist.items.

noxe: and btw - the webservice call worked - i got a javascript exception - but i try again once the build is finished

image

noxe: works perfect now - sorry for my wrong assumption :)

PR #2547, opened 2021-11-30
Closes #1660: Added RabbitMQ activities

### LINKED ISSUE #1660: RabbitMQ Activity
New activity based on rebus:

As a user I like to set rabbit credentials in the activity, select an exchange I like to listen to and define headers (filters). I like to use the rabbit activity as a trigger to start my workflow.

image

constantine-v:
- Added RabbitMQ activities for sending/receiving messages (for now with configuration on activity level only)
- Added handlers to WorkflowTesting module to make RabbitMQ activities 'testable' (not sure how good this approach is, but could be used to test other bus activities as well)
- Moved logic related to testing workflows into WorkflowTesting module


jwillmer:
@sfmskywalker you had some time to validate the pull request?

src/activities/Elsa.Activities.RabbitMq/Decorators/MessageHandlingTransportDecorator .cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public string Address => _transport.Address;
+
+        public void CreateQueue(string address) => _transport.CreateQueue(address);
+
+        public Task Send(string destinationAddress, TransportMessage message, ITransactionContext context)
+            => _transport.Send(destinationAddress, message, context);
+
+        public Task<TransportMessage> Receive(ITransactionContext context, CancellationToken cancellationToken)
+            => _toggle.IsReceivingMessages
```
sfmskywalker, line 29: Can you explain the purpose of the toggle and why it is needed?

src/activities/Elsa.Activities.RabbitMq/Services/Client.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Elsa.Activities.RabbitMq.Services
+{
+    public class Client : IClient
+    {
+        private BuiltinHandlerActivator _activator { get; set; }
+        private IBus _bus { get; set; }
```
sfmskywalker, line 20: Properties should be named `PascalCased`. Alternatively, these private members can be turned into fields, in which case the casing is correct as it is.

src/activities/Elsa.Activities.RabbitMq/Services/RabbitMqQueueStarter.cs
```diff
@@ -0,0 +1,168 @@
...
+        public async Task<SenderWorker> CreateSenderWorkerAsync(RabbitMqBusConfiguration config, CancellationToken cancellationToken = default)
+        {
+            var sender = await _messageSenderClientFactory.GetSenderAsync(config, cancellationToken);
+            return ActivatorUtilities.CreateInstance<SenderWorker>(_serviceProvider, (Func<IClient, Task>)DisposeSenderAsync, sender);
+        }
+
+        private async Task DisposeReceiverAsync(IClient messageReceiver) => await _messageReceiverClientFactory.DisposeReceiverAsync(messageReceiver);
+        private async Task DisposeSenderAsync(IClient messageSender) => await _messageSenderClientFactory.DisposeSenderAsync(messageSender);
+
+        private async IAsyncEnumerable<RabbitMqBusConfiguration> GetReceiverConfigurationsAsync([EnumeratorCancellation] CancellationToken cancellationToken)
```
sfmskywalker, line 99: This method appears to be unused.

src/activities/Elsa.Activities.RabbitMq/Services/ReceiverWorker.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class ReceiverWorker : WorkerBase
+    {
+        private readonly Scoped<IWorkflowLaunchpad> _workflowLaunchpad;
+        private string ActivityType => nameof(RabbitMqMessageReceived);
+        
+        public ReceiverWorker(
+            Scoped<IWorkflowLaunchpad> workflowLaunchpad,
+            IClient receiverClient,
+            Func<IClient, Task> disposeReceiverAction,
+            ILogger<WorkerBase> logger): base (receiverClient, disposeReceiverAction, logger)
```
sfmskywalker, line 22: Typing seems off: `ILogger<WorkerBase>` should be: `ILogger<ReceiverWorker>`.

src/activities/Elsa.Activities.RabbitMq/Services/SenderWorker.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace Elsa.Activities.RabbitMq.Services
+{
+    public class SenderWorker : WorkerBase
+    {
+        public SenderWorker(
+            IClient receiverClient,
+            Func<IClient, Task> disposeReceiverAction,
+            ILogger<WorkerBase> logger): base (receiverClient, disposeReceiverAction, logger)
```
sfmskywalker, line 12: Typing seems off: `ILogger<WorkerBase>` should be: `ILogger<SenderWorker >`.

src/activities/Elsa.Activities.RabbitMq/Services/WorkerBase.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace Elsa.Activities.RabbitMq.Services
+{
+    public class WorkerBase : IAsyncDisposable
+    {
+        protected readonly IClient _client;
+        protected readonly ILogger _logger;
+        protected readonly Func<IClient, Task> _disposeReceiverAction;
```
sfmskywalker, line 11: Protected fields should be named using Pascal Casing.

src/modules/Elsa.WorkflowTesting/Handlers/ConfigureRabbitMqActivitiesForTestHandler.cs
```diff
@@ -0,0 +1,62 @@
+using Elsa.Events;
+using Elsa.Testing.Events;
+using Elsa.WorkflowTesting.Services;
+using MediatR;
+
+namespace Elsa.WorkflowTesting.Handlers
+{
+    public class ConfigureRabbitMqActivitiesForTestHandler : INotificationHandler<WorkflowExecuting>, INotificationHandler<WorkflowFaulted>, INotificationHandler<WorkflowCompleted>, INotificationHandler<WorkflowTestExecutionStopped>
```
sfmskywalker, line 8: This class should be moved to the `Elsa.Activities.RabbitMQ` project. The `Elsa.WorkflowTesting` module itself should be agnostic to specialized activities such as RabbitMQ (think open-closed principle where anyone should be able to add more activities to the system without the need for us to update core modules such as testing).

src/modules/Elsa.WorkflowTesting/Services/IRabbitMqTestQueueManager.cs
```diff
@@ -0,0 +1,8 @@
+namespace Elsa.WorkflowTesting.Services
+{
+    public interface IRabbitMqTestQueueManager
```
sfmskywalker, line 3: Should be moved to `Elsa.Activities.RabbitMQ`.

src/modules/Elsa.WorkflowTesting/Services/RabbitMqTestQueueManager.cs
```diff
@@ -0,0 +1,136 @@
...
+using Elsa.Activities.RabbitMq.Services;
+using Elsa.Services;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System.Runtime.CompilerServices;
+
+namespace Elsa.WorkflowTesting.Services
+{
+    public class RabbitMqTestQueueManager : IRabbitMqTestQueueManager
```
sfmskywalker, line 12: Should be moved to `Elsa.Activities.RabbitMQ`.

src/modules/Elsa.WorkflowTesting/Services/RabbitMqTestQueueManager.cs
```diff
@@ -0,0 +1,136 @@
...
+        private readonly IDictionary<string, ICollection<WorkerBase>> _workers;
+        private readonly IRabbitMqQueueStarter _rabbitMqQueueStarter;
+        private readonly ILogger _logger;
+
+        public RabbitMqTestQueueManager(
+            IConfiguration configuration, 
+            IServiceScopeFactory scopeFactory, 
+            IServiceProvider serviceProvider,
+            IRabbitMqQueueStarter rabbitMqQueueStarter,
+            ILogger<RabbitMqQueueStarter> logger)
```
sfmskywalker, line 27: Should be: `ILogger<RabbitMqTestQueueManager>`.

sfmskywalker (changes requested):
Could you perhaps add a small sample application to test + show case the new activities?

src/activities/Elsa.Activities.RabbitMq/Decorators/MessageHandlingTransportDecorator .cs
constantine-v: The purpose of the toggle is to have more strict control on when activity is able to receive messages. The idea behind it, is that receiver activity is meant to handle a single message, but message handler can be triggered multiple times in a very short period, as soon as it is registered. Even though it doesn't seem to be affecting workflow in any way, messages from the broker are getting consumed by activity without any real need for it. Toggle is just a fast and easy way to flip a switch and prevent client from handling any more incoming messages.

sfmskywalker: I see. Do we understand why the message handler is triggered multiple times?

sfmskywalker: OK. But could it then not also happen that there are different messages enqueued that each need to be processed, but as long as this activity is executing, all subsequent messages are now being ignored?

If so, then perhaps a better approach might be to not discard messages, but actually handle them and let the engine ignore them. Unless you can be sure that a secondary message is in fact the same message that was already handled.

jwillmer: I think I agree with @sfmskywalker . Let the system handle the messages. It will need to anyway as soon as the workflow gets published. At least if I understand the issue correctly.

constantine-v: @sfmskywalker I've removed toggle in the last commit.

sfmskywalker (approved):

PR #2577, opened 2021-12-10
Update FileSystemEventBookmark.cs

### LINKED ISSUE #2527: WatchDirectory does not work with pattern *.*
Hi,

i am creating a new issue - so we dont get lost. as described here - https://github.com/elsa-workflows/elsa-core/issues/1654 - when using * . * as pattern - the workflow does not start. the reason is that the filewatcher normalizes the pattern from * . * to * - so in memory the tigger based on the activity is * . * - but the filesystem event does an bookmark query based on the pattern * - so it does not find any trigger. any suggestion how to fix this?

sfmskywalker:
Fixes #2527


noxe:
i was to late :) - thanks for this - looks good! 👌

sfmskywalker:
Sorry, I was too eager to close out issues ;) Thanks for checking!

noxe:
😎

PR #2603, opened 2021-12-19
Cleanup of WorkflowTesting and RabbitMQ

constantine-v:
1. Moved workflowtesting module projects into physical folder 'workflowsettings', so physical folder structure is consistent with solution structure.
2. Fixed number of warnings in RabbitMQ module.
3. Added description, copyright, license information to RabbitMQ project


src/activities/Elsa.Activities.RabbitMq/Services/Client.cs
```diff
@@ -53,7 +53,7 @@ public async Task PublishMessage(string message)
         {
             if (_bus == null) ConfigureAsOneWayClient();
 
-            await _bus.Advanced.Topics.Publish(Configuration.RoutingKey, message, Configuration.Headers);
+            await _bus!.Advanced.Topics.Publish(Configuration.RoutingKey, message, Configuration.Headers);
```
sfmskywalker, line 56: We shouldn't have to use the exclamation mark here since there is a null check on line 54.

sfmskywalker (approved):
Thank you!

PR #2619, opened 2021-12-21
Added feature panel.

### LINKED ISSUE #2507: Allow users to toggle between different UI/UX experiences
Changing Designer UI/UX may impact a lots of users. Reduce adoption frustration by allowing users to toggle between new and existing UI/UX features.

Example use case:
When enabling horizontal ordering of activities, we don't want to remove the vertical editor (many users may still be used to that).

Allow configuring global designer settings under designer expandable side panel where general and test tabs are located.
- [X] Allow API's to control the options in the widget.
- [X] Save information in browser store.
- [X] Integrated this with designer boottrap in Elsa Core so the settings can be controlled by configuration. Please consider the settings must be set at application level (for All users), and as individual user level.

Check proposal (example from Azure DevOps preview feature toggles).

Screenshot 2021-11-22 at 11 11 09

Screenshot 2021-11-22 at 11 01 44

simona-taran:
Closes #2507


sfmskywalker (approved):
Just in time for the 2.4 release. Very nice!

PR #2648, opened 2022-01-04
Update scripting client to use POST endpoint method

dominicasley:
Small fix to keep up to date with recent changes.

9f36d668 updated the JavaScript TypeDefinitions endpoint `Elsa.Server.Api/Endpoints/Scripting/JavaScript/TypeDefinitions/Get.cs` to use POST instead of GET.


sfmskywalker (approved):
Thanks!

Issue #2651, opened 2022-01-05
Composite Activities are not visible in ELSA Dashboard in version 2.4.1-preview.11


OlegLukash:
Hi all! I have updated to 2.4.1-preview.11 from 2.3.0 version. Found a problem that CompositeActivities are not visible anymore in Elsa Dashboard when want to create a new workflow. Has somebody faced this issue? Custom Composite Activities inherit from CompositeActivity class and is registered in DI as was when used 2.3.0 version. PS. Also updated the project to .NET 6.
Simple Activities are visible

jdevillard:
Try to specify [Browsable(True)] in head of your custom composite activity.

OlegLukash:
it works. thanks a million

PR #2652, opened 2022-01-06
Do not force Browsable of SubActivity to false if not defined.

jdevillard:
Allow Composite Activity or other subclass to not explicitly defined [Browsable(true)] if base class define [Browsable(false)]

refer to #2651


sfmskywalker (approved):
Nice!

PR #2716, opened 2022-01-30
Update Elsa.Client workflow definition models to match Elsa.Server

dominicasley:
Updates client workflow definition models to match the API.

`SaveWorkflowDefinitionRequest.cs`: 
- Adds Tag and Channel properties
- Changes Variables to be a string

`WorkflowDefinition.cs`:
- Adds CustomAttributes
- Changes Variables to be a string as this is the actual return type after the `VariablesConverter` is ran


sfmskywalker (approved):
Thanks for catching & fixing this one!

Issue #242, opened 2020-01-20
[FEAT] Multitenancy


IntranetFactory:
We currently use a TenantId in all our tables, so that data from multiple tenants can be stored in one database instance.

Are there any plans to support Multitenancy in the persistance layer?

sfmskywalker:
I do want support for multitenant applications, including the workflow server. Let's update the models with a `string? TenantId` property and update the persistence providers.

sfmskywalker:
Things to do:

*WorkflowDefinition*

Up until now, there was no need for a `WorkflowDefinition` model - the `WorkflowDefinitionVersion` model was sufficient, since all of the properties were versioned anyway.

However, we need to add a `TenantId` property, which itself is not versioned. Although we could technically add it to `WorkflowDefinitionVersion`, let's not; instead, let's introduce `WorkflowDefinition` which would look as follows:

```csharp
public class WorkflowDefinition
{
   public string Id { get; set; }
   public string? TenantId { get; set; }
   public Instant CreatedAt { get; set; }
}
```

The `WorkflowDefinitionVersion` model already contains a `DefinitionId` property that represents the logical workflow definition ID, so that's fine.

*IWorkflowDefinitionStore*

The `IWorkflowDefinitionStore` needs to be refactor-renamed to `IWorkflowDefinitionVersionStore`, including all of its current implementations.

Then, a new interface needs to be created called `IWorkflowDefinitionStore` that can read and write `WorkflowDefinition` objects to the underlying persistence layers.

When listing all workflow definitions, the method should accept an optional `TenantId` value. IF specified, the underlying persistence provider should use this value as a WHERE clause to make sure that only workflow definitions belonging to the specified tenant are returned.

*WorkflowInstance*

The `WorkflowInstance` model probably does not need to be updated, since a WorkflowInstance has a `DefinitionId`, which points to a `WorkflowDefinition`, which has a `TenantId`.

smadurange:
What's the status of this? Is this in the roadmap or in-view for now?

sfmskywalker:
Some initial exploratory work has been done, but it's far from complete. It's likely to get more attention later this year, after the Elsa 2 release.

tunechr:
Hi @sfmskywalker 

You've probably already thought of this but I'm just thinking about the multi-tenancy and the implications for activity configuration. 

An example might be that a tenant has their own email service, blob storage, telnyx provider etc. 
[...]

PR #2727, opened 2022-02-03
multitenancy

constantine-v:
The goal of this PR is to introduce multitenancy (Tenant per database).
It does not take into account existing code that assumes sharing existing database between all Tenants.
Multitenancy can be configured through settings, here's example for appsettings.json
```
"Multitenancy": true,
"Tenants": {
    "Default": {
        "Name": "Default Tenant",
        "Prefix": "default",
        "DatabaseConnectionString": "Data Source = elsa.sqlite.db;Cache = Shared"
        ...any other relevant settings
    }
}
```
Multitenancy (true/false) setting is needed for backwards compatibility purposes. When it is not set to true, default tenant will be created with DatabaseConnectionSttring taken from regular Elsa settings.
Prefix setting is needed to distinguish between different tenants on frontend or when dealing with cached data (which is accessed by hardcoded key).
Since user authentication is out of Elsa's scope, detecting current tenant and redirecting to appropriate URL also should be handled outside of Elsa. When dashboard loads for the first time, it gets tenant prefix from the URL and compares it with all tenant prefixes loaded from API. If there is a match, all links are adjusted accordingly by adding tenant prefix to them. For example, if initially user is redirected to https://elsa.mydomain.com/mytenant/, all subsequent requests will follow the same route e.g. https:\\elsa.mydomain.com\mytenant\v1\workflow-definitions.

On backend, current tenant is read from request URL prefix and stored in scoped service TenantProvider. For the scenarios when there is no request URL (for instance when running startup tasks during application start), tenant is set explicitly using CreateScopeForTenant(ITenant tenant) extension method for ServiceScopeFactory.


src/activities/Elsa.Activities.AzureServiceBus/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -70,5 +70,27 @@ private static ManagementClient CreateServiceBusManagementClient(IServiceProvide
...
+            else
+                return ActivatorUtilities.GetServiceOrCreateInstance<ServiceBusQueuesStarter>(serviceProvider);
+        }
+
+        private static IServiceBusTopicsStarter CreateServiceBusTopicsStarter(IServiceProvider serviceProvider)
+        {
+            var options = serviceProvider.GetRequiredService<IOptions<AzureServiceBusOptions>>().Value;
+            var multitenancyEnabled = options.MultitenancyEnabled;
+
+            if (multitenancyEnabled)
```
sfmskywalker, line 90: Is there any reason *not* to always assume multi-tenancy (with DEFAULT being the default tenant)?

constantine-v: Good point, I've had a reason not to do it earlier, but then I changed implementation and now it makes sense to get rid of this check and additional 'multitenant' classes.

sfmskywalker:
I noticed the addition of a new model called `Tenant` to Elsa.Core, which contains `ConnectionString` and `Prefix` properties.
It seems too opinionated to me that a tenant is always associated with a connection string, or at least too narrowly-scoped.

Having tenant-specific configuration makes sense, so perhaps we can replace Prefix and ConnectionString with a more general-purpose `IConfigurationSection` kind of thing?

Only some modules care about a tenant-specific DB connection string; others might want to have a tenant-specific Azure SB connection string, etc.

constantine-v:
I've added TenantConfiguration class that contains dictionary of settings (initially pulled from tenant configuration section). I've replaced Prefix and ConnectionString with TenantConfiguration and added couple of extension methods to retrieve prefix and database connection string. Modules can add their own extension methods to retrieve settings specific for them.

sfmskywalker:
@vargaendre there is [an issue about this topic](https://github.com/elsa-workflows/elsa-core/issues/242), but it doesn't provide a complete description (far from it). Perhaps @constantine-v can update the description of this PR with basic instructions on how to use it?

tomy2105:
I would definitely say it would be nice to have a place in documentation where multitenancy (what is supported and what is not is described).
A comment on this PR could be a good starting point, yes, but somewhere inside documentation would be far better :).

steviej08:
This is looking really good. I'm also interested in this PR. I've been implementing something similar to this for our product, as we need this fairly soon. I'm wondering, is this targeted for a particular release? Are there any rough timings on said release?

Also, I'm wondering if there is scope to provide asynchronous access to tenants. That is the ITenantStore and ITenantProvider having async functions. This would make it easy to switch out the current TenantStore where we may be accessing an external service to grab tenant information. I also wonder whether we could have an interface for Tenant. Then in the ITenantProvider we can do what we need in when setting the current tenant, like load a bunch of settings for that account (I haven't gone through the PR in a lot of detail btw and appreciate there may be some sort of specific model you guys are aiming for).

sfmskywalker:
Just as a quick reminder putting this here: need to update this question about multi-tenancy support once this feature goes in: https://github.com/elsa-workflows/elsa-core/discussions/140

sfmskywalker:
@steviej08 You're right, the `ITenantStore` and `ITenantProvider` services' methods should be made `async` (I assume they are not currently so based on your comment).

Same goed for introducing `ITenant` for `Tenant`.

@constantine-v This should be straightforward to do, right?

constantine-v:
@sfmskywalker, @steviej08 yes, that should be quite straightforward, I'll make sure to push it with next commit. I'll also put some details into description so it could be used as a base for documentation.

constantine-v:
Hi @tomy2105, 
I'm afraid at this point the plan is to support hardcoded url prefix scenario (http://example.com/tenant1/, http://example.com/tenant2/, etc..).

However, tenant detection logic could be moved to a middleware, so it could be swapped with another middleware with a different tenant detection logic if needed. I need to process it a bit, but it could be just enough to support more flexible urls.

For backwards-compatibility purposes, all endpoints support urls both with and without tenant prefix, so if multitenancy is not enabled, it would work just as it was before.

tomy2105:
@constantine-v thanks for considering my use case. If it is "relatively" easy to modify your changes to include my usecase that would be great, if not I'll find some way to work around it :).

One question more.
Is it going to be one tenant -> one database or is multiple tenants -> one database or multiple tenants -> muiltiple databases (some of tenants can be in one, some of them in other)?

constantine-v:
@tomy2105 it is going to be database per tenant. I'm planning to update this PR description with all the details soon.

tomy2105:
@constantine-v is the door for having one database -> multiple tenants and/or multiple database -> multiple tenants still open or not?
I ask because I know there was a tenantid inside db schema so far so wondering if this can be used to achieve these scenarios by pointing configuration of multiple tenants to same db?

tomy2105:
@constantine-v  in light of my last question(s) how are your changes be related to the existing stuff with ITenantAccessor?
Please include this relation (or lack of if none exist) in the documentation too :).

src/samples/server/Elsa.Samples.Server.Host/appsettings.json
```diff
@@ -121,6 +121,14 @@
...
                 }
             ]
+        },
+        "Multitenancy": false,
+        "Tenants": {
+            "Default": {
+                "Name": "Default Tenant",
+                "Prefix": "default",
+                "DatabaseConnectionString": "Data Source = elsa.sqlite.db;Cache = Shared"
+            }
```
sfmskywalker, line 131: How about we combine these two sections into one? E.g.

```json
"Multitenancy" : {
   "Enabled": true|false,
   "Tenants" {
      ...
   }
}
```

ricksearcy:
What about scenarios where Tenants can't be statically configured in appsettings? In our use case we use a reverse proxy to communicate with Elsa and send in tenant and user information using jwt. Tenants could be added at any time. Using the existing ITenantAccessor almost worked for our scenario, but there seems to be a bug that is keeping us from fully going down that path. https://github.com/elsa-workflows/elsa-core/issues/2861

tomy2105:
One more question :).
If one database per tenant is needed (for workflow, instance, etc.... storage), do we need per tenant configuration for  Service Bus Broker, Distributed Lock Provider and Distributed Cache Signal Provider or those stay as is?

sfmskywalker:
> What about scenarios where Tenants can't be statically configured in appsettings? In our use case we use a reverse proxy to communicate with Elsa and send in tenant and user information using jwt. Tenants could be added at any time. Using the existing ITenantAccessor almost worked for our scenario, but there seems to be a bug that is keeping us from fully going down that path. #2861

Sounds like what is needed for this is DB-level multitenancy (where tenant ID is used as a column on various DB tables). This is partially implemented (and you need #2861 to be fixed)

constantine-v:
Closing this PR, as it was agreed that multitenancy will be implemented using a better approach with Autofac library.

PR #2737, opened 2022-02-08
Optimize ForEach activity

noxe:
This PR optimizes the ForEach Items property - to use IEnumerable type - this way any collection can be assigned - and there is no need that the EnumerableResultConverter needs to Serialize and Deserialize the collection - since desiredType.IsAssignableFrom(enumerable.GetType() gives True - and returns the enumarable as is.

### DIFF
--- a/src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
+++ b/src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
@@ -24,6 +24,8 @@ namespace Elsa.Activities.ControlFlow
             DefaultSyntax = SyntaxNames.Json,
             SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid }
         )]
+        // Using IEnumerable here so the EnumerableResultConverter does not need to serialize/deserialize the collection instance - so to be able to return the instance as it is
+        // See here for further details: https://github.com/elsa-workflows/elsa-core/pull/2737
         public IEnumerable Items { get; set; } = new Collection<object>();
 
         [ActivityOutput] public object? Output { get; set; }
--- a/src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
+++ b/src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+using System.Collections;
 using System.Collections.ObjectModel;
 using System.Linq;
 using Elsa.ActivityResults;
@@ -24,8 +24,8 @@ namespace Elsa.Activities.ControlFlow
             DefaultSyntax = SyntaxNames.Json,
             SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid }
         )]
-        public ICollection<object> Items { get; set; } = new Collection<object>();
-        
+        public IEnumerable Items { get; set; } = new Collection<object>();
+
         [ActivityOutput] public object? Output { get; set; }
 
         private int? CurrentIndex
@@ -50,7 +50,7 @@ namespace Elsa.Activities.ControlFlow
                 return Done();
             }
 
-            var collection = Items.ToList();
+            var collection = Items.OfType<object>().ToList();
             var currentIndex = CurrentIndex ?? 0;
 
             if (currentIndex < collection.Count)
@@ -65,7 +65,7 @@ namespace Elsa.Activities.ControlFlow
                 Output = currentValue;
                 context.JournalData.Add("Current Index", currentIndex);
                 context.LogOutputProperty(this, nameof(Output), Output);
-                
+
                 return Outcome(OutcomeNames.Iterate, currentValue);
             }
 


sfmskywalker:
Although this works around getting the property from being (de)serialized, what if the user actually needs the objects to be serialized & deserialized?

For example, I have a list of `Customer` objects that I want to iterate on. In each iteration, I suspend the workflow, waiting for input, before resuming the iteration.

When the workflow is resumed, the `Items` property need to be re-constructed as well.
Or perhaps I'm misunderstanding something here?

noxe:
my current problem is that we are providing our Items based on JavaScript syntax - and because of this during evaluation of the Expression the EnumerableResultConverter is run. Here the enumarable is passed - as well as the desired type - which is ICollection<object> in this case - so none if the existing If's are true - and the collection is serialized / deserialized here.

i think you use case should still work fine and has nothing todo with resuming the workflow?

noxe:
so obasicly this will not change anything about resuming / persisting the workflow i guess - it simply changes the property definition - so the EnumerableResultConverter does not need to serialize / deserialize - and so it is able to 1:1 return the enumerable instance

src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
```diff
@@ -24,8 +24,8 @@ public class ForEach : Activity
             DefaultSyntax = SyntaxNames.Json,
             SupportedSyntaxes = new[] { SyntaxNames.Json, SyntaxNames.JavaScript, SyntaxNames.Liquid }
         )]
-        public ICollection<object> Items { get; set; } = new Collection<object>();
-        
+        public IEnumerable Items { get; set; } = new Collection<object>();
```
sfmskywalker, line 27: Would be good to add a comment here explaining why we're using `IEnumerable` (and reference this PR).

sfmskywalker (approved):

src/core/Elsa.Core/Activities/ControlFlow/ForEach/ForEach.cs
noxe: sure - i update the PR

noxe: done :)

sfmskywalker:
I had to revert this unfortunately, as it's causing an issue for certain users:

```json
"innerException": {
          "$id": "142",
          "type": "Newtonsoft.Json.JsonSerializationException, Newtonsoft.Json",
          "message": "Cannot create and populate list type System.Collections.IEnumerable. Path '', line 1, position 1.",
          "stackTrace": "   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewList(JsonReader reader, JsonArrayContract contract, Boolean& createdFromNonDefaultCreator)\r\n   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateList(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, Object existingValue, String id)\r\n   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)\r\n   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)\r\n   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)\r\n   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)\r\n   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type)\r\n   at Elsa.Expressions.JsonHandler.EvaluateAsync(String expression, Type returnType, ActivityExecutionContext context, CancellationToken cancellationToken)\r\n   at Elsa.Expressions.ExpressionEvaluator.EvaluateAsync(String expression, String syntax, Type returnType, ActivityExecutionContext context, CancellationToken cancellationToken)",
          "data": {
            "$id": "143",
            "$type": "System.Collections.ListDictionaryInternal, System.Private.CoreLib"
          }
        },
        "data": {
          "$id": "144",
          "$type": "System.Collections.ListDictionaryInternal, System.Private.CoreLib",
          "Expression": "[5,1,180,1]",
          "Syntax": "Json"
        }
      },
```

I'll need to give this some thought.

@noxe if you could send me a project that reproduces your problem (without the change introduced with your PR) then I can look into that and come up with a fix that doesn't break the other users.

noxe:
:( - ok i create an repo project!

noxe:
@sfmskywalker - do you probably also have a reference to the current isseu - github or discord?

sfmskywalker:
Not yet. This happened in production with a customer after updating their solution. I’ll try to reproduce their scenario in a separate solution.

noxe:
oh ok - i try to create a repro as soon as i can

sfmskywalker:
Thank you

noxe:
hey @sfmskywalker  - attached now you find a sample based on the ElsaDashboard.Samples.AspNetCore.Monolith

when you launch the project - you see it contains already an workflow - which can be launched via https://localhost:15265/test

you should see it failing because of the EnumerableResultConverter -> this is the key point - as the system tries to convert here collection to json and then back to the target type - which i need to avoid. this only happens when providing values via Javascript. So if changing the ForEach activity is not an option - maybe provide an better logic in the EnumerableResultConverter to avoid serializing/deserializing?

[ElsaDashboard.Samples.AspNetCore.Monolith.zip](https://github.com/elsa-workflows/elsa-core/files/8060710/ElsaDashboard.Samples.AspNetCore.Monolith.zip)


the project contains XPO as ORM mapper - it creates some records on startup in an inmemory provider. we also have an VDBaseObjectJsonConverter to serialize only the object id if such an object is serialized - because all our custom activites provide their values Transient - there is no need to restore such objects from an workflow.

PR #2771, opened 2022-02-22
1648 sql activity

andrewkorol:


src/activities/Elsa.Activities.SQL/Client/SqlClient.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private string _connectionString { get; set; }
+        public SqlClient(string connectionString)
+        {
+            _connectionString = connectionString;
+        }
+
+        public int Execute(string sqlCommand)
+        {
+            using (SqlConnection connection = new SqlConnection(_connectionString))
```
sfmskywalker, line 21: Use `var`.

src/activities/Elsa.Activities.SQL/Client/SqlClient.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Elsa.Activities.SQL.Client
+{
+    public class SqlClient
+    {
+        private string _connectionString { get; set; }
```
sfmskywalker, line 13: Can be made readonly.

src/activities/Elsa.Activities.SQL/Activities/SqlActivity.cs
```diff
@@ -0,0 +1,63 @@
...
+using Elsa.Design;
+using Elsa.Expressions;
+using Microsoft.Extensions.Configuration;
+using Elsa.Activities.SQL.Configuration;
+using Elsa.Activities.SQL.Client;
+
+namespace Elsa.Activities.SQL.Activities
+{
+    /// <summary>
+    /// Stores a set of possible user actions and halts the workflow until one of the actions has been performed.
```
sfmskywalker, line 21: Description does not fit this activity's purpose.

src/activities/Elsa.Activities.SQL/Activities/SqlActivity.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    [Trigger(
+        Category = "Sql",
```
sfmskywalker, line 24: "SQL" (it's written language).

src/activities/Elsa.Activities.SQL/Configuration/QueryConfiguration.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Elsa.Activities.SQL.Configuration
+{
+    public class QueryConfiguration
```
sfmskywalker, line 9: Is this class used?

src/activities/Elsa.Activities.SQL/Elsa.Activities.SQL.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <Import Project="..\..\..\common.props" />
+    <Import Project="..\..\..\configureawait.props" />
+    
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
```
sfmskywalker, line 8: Missing `<Description>` and `<PackageTags>`  elements.

src/activities/Elsa.Activities.SQL/Elsa.Activities.SQL.csproj
```diff
@@ -0,0 +1,16 @@
...
+
+    <ItemGroup>
+        <ProjectReference Include="..\..\core\Elsa.Core\Elsa.Core.csproj" />
+        <ProjectReference Include="..\..\persistence\Elsa.Persistence.EntityFramework\Elsa.Persistence.EntityFramework.Sqlite\Elsa.Persistence.EntityFramework.Sqlite.csproj" />
+        <ProjectReference Include="..\webhooks\Elsa.Webhooks.Persistence.EntityFramework.Sqlite\Elsa.Webhooks.Persistence.EntityFramework.Sqlite.csproj" />
```
sfmskywalker, line 13: Is this project reference needed?

src/activities/Elsa.Activities.SQL/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Elsa.Activities.SQL.Extensions
+{
+    using SqlActivity = Elsa.Activities.SQL.Activities.SqlActivity;
```
sfmskywalker, line 12: Is this alias necessary?

src/activities/Elsa.Activities.SQL/Activities/SqlActivity.cs
```diff
@@ -0,0 +1,63 @@
...
+using Elsa.Serialization;
+using Elsa.Services.Models;
+using Elsa.ActivityResults;
+using Elsa.Design;
+using Elsa.Expressions;
+using Microsoft.Extensions.Configuration;
+using Elsa.Activities.SQL.Configuration;
+using Elsa.Activities.SQL.Client;
+
+namespace Elsa.Activities.SQL.Activities
```
sfmskywalker, line 18: Namespace should be `Sql` instead of `SQL`.

src/activities/Elsa.Activities.SQL/Client/SqlClient.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
```
mohdali, line 7: `Microsoft.Data.SqlClient` is specific to Microsoft SQL Server I think. We need to abstract the client so any database can be supported (Oracle, SQLite, MySQL etc.)

src/activities/Elsa.Activities.SQL/Elsa.Activities.SQL.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <Import Project="..\..\..\common.props" />
+    <Import Project="..\..\..\configureawait.props" />
+    
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
mohdali, line 7: net5.0 will be out of support by May 2022. Can we add targets for 3.1 and 6?

src/activities/Elsa.Activities.SQL/Client/SqlClient.cs
sfmskywalker: There are a number of ways to look at this.

1. This activity is a low-level database activity that targets a specific engine (SQL Server). This means we could add multiple activities that target a specific database engine.
2. This activity is a high-level database activity that can target any supported engine. The user would have to choose the desired target engine from a dropdown list.
3. A variation on option 2, but instead of letting the user select from a dropdown list, there would be an activity type provider that dynamically adds database activities based on configured database engine client providers.

One nice advantage of the first option is that the activity can potentially expose database engine-specific settings. But if we think we never need specialized settings for a given database engine (and they all just need a connection string), then option 2 might be the sweet spot.

@andrewkorol @jwillmer Any thoughts on this?

jwillmer: I think for now this will cover option one (since we require it) but I agree that it would be a nice enhancement to support more. I guess the best option is to wait for user demand and granular improve it.

mohdali: > There are a number of ways to look at this.
> 
> 1. This activity is a low-level database activity that targets a specific engine (SQL Server). This means we could add multiple activities that target a specific database engine.
> 2. This activity is a high-level database activity that can target any supported engine. The user would have to choose the desired target engine from a dropdown list.
> 3. A variation on option 2, but instead of letting the user select from a dropdown list, there would be an activity type provider that dynamically adds database activities based on configured database engine client providers.
> 
> One nice advantage of the first option is that the activity can potentially expose database engine-specific settings. But if we think we never need specialized settings for a given database engine (and they all just need a connection string), then option 2 might be the sweet spot.
> 
> @andrewkorol @jwillmer Any thoughts on this?

I agree that as a start we can make it low level and specific to SQL server. Maybe the activity can be renamed to SqlServer to avoid confusion.

I don't think there will be any further customization needed for the different databases though. All will need to select a driver, provide a connection string and a query. Differences might occur if we introduce bluk activities in the future.

This activity will have good potential for future expansion to support SQL syntax with paramerized queries, support bulk operations, and so on. I really appreciate that it already started.

sfmskywalker: > Maybe the activity can be renamed to SqlServer to avoid confusion.

I was thinking the same thing. Then we can always later add multiple flavors, and/or add a "generic" version from which the user can select a provider from a dropdown list. At the very least, renaming to e.g. "SqlServer" gives us these options.

Also, the activity should probably be named a little bit more specifically, because what it's doing is: running a SQL query.
So the name should probably something like: `ExecuteSqlServerQuery` (a verb). The term `SqlServer` is just a noun and doesn't convey what it's doing.

src/activities/Elsa.Activities.SQL/Activities/ExecuteSqlServerQuery.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        [ActivityInput(
+            Hint = "SQl script to execute",
+            UIHint = ActivityInputUIHints.MultiLine,
+            SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid }
+        )]
+        public string Query { get; set; } = default!;
+
+        /// <summary>
+        /// The ConnectionStrings to run SQL
```
sfmskywalker, line 30: Should it be singular: "connection string" instead? Also notice the use of regular English: "connection string" rather than "ConnectionStrings". Same comment for the `Hint` property.

src/activities/Elsa.Activities.SQL/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+using Elsa.Activities.ExecuteSqlServerQuery.Activities;
+using Elsa.Activities.ExecuteSqlServerQuery.Factory;
+using Elsa.Options;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Elsa.Activities.ExecuteSqlServerQuery.Extensions
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static ElsaOptionsBuilder AddSqlActivities(this ElsaOptionsBuilder elsa)
```
sfmskywalker, line 11: Maybe rename to `AddSqlServerActivities`?

src/designer/elsa-workflows-studio/src/components/screens/workflow-blueprint-viewer/elsa-workflow-blueprint-viewer-screen/elsa-workflow-blueprint-viewer-screen.tsx
```diff
@@ -90,6 +90,7 @@ export class ElsaWorkflowBlueprintViewerScreen {
   async loadActivityDescriptors() {
     const client = await createElsaClient(this.serverUrl);
     state.activityDescriptors = await client.activitiesApi.list();
+    console.log(state.activityDescriptors);
```
sfmskywalker, line 93: Not sure if you added it or me or someone else, but perhaps we should remove this?

sfmskywalker (changes requested):

src/activities/Elsa.Activities.SQL/Elsa.Activities.ExecuteSqlServerQuery.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <TargetFrameworks>net6.0;netcoreapp3.1</TargetFrameworks>
+      <Description>
+          Elsa is a set of workflow libraries and tools that enable lean and mean workflowing capabilities in any .NET Core application.
+          This package provides activities to run SQL queries on given database using an connaction string.
+      </Description>
+      <PackageTags>elsa, workflows</PackageTags>
+  </PropertyGroup>
+    
+  <ItemGroup>
+    <PackageReference Include="System.Data.SqlClient" Version="4.8.3" />
```
mohdali, line 16: Please use `Microsoft.Data.SqlClient` instead.

Refer this post: https://devblogs.microsoft.com/dotnet/introducing-the-new-microsoftdatasqlclient/

sfmskywalker (approved):

PR #2801, opened 2022-03-02
Optimize string checking for Attachment

noxe:
Hi @sfmskywalker - it currently happens that in the SendEmail activity - when you enter something in the attachment field (default or javascript) - and then delete the whole string in der designer - that the Attachment property contains an empty string ( "" ) - which then leads to an error because it tries to attach an file - where the string is empty.

i added 2 checks where for the string case!


sfmskywalker (approved):

sfmskywalker:
Thanks!

PR #2819, opened 2022-03-09
Workflow Testing UI Improvement

constantine-v:
Currently, when workflow is executed from test panel and there is already an instance of this workflow running (typical scenario - execute workflow with blocking activity and refresh the page) - there is no feedback telling user what exactly is going on. Developer tools will display 404 error, but that tells very little as well.

I've added confirmation modal for this scenario, that is asking if user wants to stop current instance and start a new one.


sfmskywalker (approved):

sfmskywalker:
Great!

PR #2841, opened 2022-03-15
Sql query activity

andrewkorol:


mohdali:
@andrewkorol @sfmskywalker not directly related to this PR but there is a breaking change in `Microsoft.Data.SqlClient` v4 that forces encryption by default. This will break any connection if the SSL certificate is invalid and you have to add Encrypt=False to your connection string.

Can we downgrade to version 3? without that it will impact EF as well and break every connection especially for docker.

https://weblog.west-wind.com/posts/2021/Dec/07/Connection-Failures-with-MicrosoftDataSqlClient-4-and-later
https://techcommunity.microsoft.com/t5/sql-server-blog/released-general-availability-of-microsoft-data-sqlclient-4-0/ba-p/2983346

andrewkorol:
> @andrewkorol @sfmskywalker not directly related to this PR but there is a breaking change in `Microsoft.Data.SqlClient` v4 that forces encryption by default. This will break any connection if the SSL certificate is invalid and you have to add Encrypt=False to your connection string.
> 
> Can we downgrade to version 3? without that it will impact EF as well and break every connection especially for docker.
> 
> https://weblog.west-wind.com/posts/2021/Dec/07/Connection-Failures-with-MicrosoftDataSqlClient-4-and-later https://techcommunity.microsoft.com/t5/sql-server-blog/released-general-availability-of-microsoft-data-sqlclient-4-0/ba-p/2983346

yes, that's good point. i guess version downgrade is the best option for now. @sfmskywalker what do you think?

mohdali:
@andrewkorol @sfmskywalker  I found that EFCore.BulkExtensions is forcing 4.0.0 anyways. So probably needs to manage there.

Is there any reported issue with Elsa 2.5 regarding this?

image

mohdali:
@andrewkorol @sfmskywalker I created separate PR to address EFCore.BulkExtensions version as well.

https://github.com/elsa-workflows/elsa-core/pull/2853

sfmskywalker:
I propose we add a hint to the SQL activities to make people aware of the fact it's using the latest [Microsoft.Data.SqlClient](https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md) package which requires encryption by default, and an example of a connection string to disable it.

For Elsa in general, we can include in the release notes a breaking change with regards to the connection string.

As soon as users upgrade to Elsa 2.6 while using SQL Server, they will encounter the same issue, for which we will have instructions ready.

The alternative is staying on Sql.Client < 4, but assuming MS is unlikely to reverse the new default change, I think it's better to have a little pain now than dealing with an outdated package for too long.

src/activities/Elsa.Activities.SQL/Activities/SqlCommand/ExecuteSqlServerCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+using Elsa.Activities.Sql.Factory;
+
+namespace Elsa.Activities.Sql.Activities
+{
+    /// <summary>
+    /// Execute an SQL query on given database using connection string
+    /// </summary>
+    [Trigger(
+        Category = "SQL Server",
+        DisplayName = "Execute Sql Command",
```
sfmskywalker, line 16: For Display Name, a proper English name is appropriate: "Execute **SQL** Command" (In English, we should use call-caps for the "SQL" acronym, as opposed in C# where we use Pascal-casing for acronyms that take 3 characters or more).

src/activities/Elsa.Activities.SQL/Activities/SqlCommand/ExecuteSqlServerCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        [ActivityInput(
+            Hint = "SQL command to execute",
+            UIHint = ActivityInputUIHints.MultiLine,
+            SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid }
+        )]
+        public string Command { get; set; } = default!;
+
+        /// <summary>
+        /// Connection string to run SQL
```
sfmskywalker, line 33: Here we could include a note that one needs to be aware of the breaking SqlClient v4 change.

src/activities/Elsa.Activities.SQL/Activities/SqlCommand/ExecuteSqlServerCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        [ActivityInput(
+            Hint = "Connection string to run SQL",
```
sfmskywalker, line 36: Here too, we could include a note that one needs to be aware of the breaking SqlClient v4 change.

sfmskywalker:
@andrewkorol After some more considerations (see @mohdali 's and my comments [here](https://github.com/elsa-workflows/elsa-core/pull/2853#issuecomment-1072246699)), let's hold off using latest SqlClient for now until after .NET 7 ships (if we can - should we start running into versioning issues with dependency packages then we will revisit earlier).

sfmskywalker (approved):

Issue #2812, opened 2022-03-07
Azure Service Bus Trigger : Race Condition when restarting worker


jdevillard:
When working on a workflow and publishing a new version , sometime the Azure Service Trigger is not working anymore.
Listener seems to no working.

After adding some log in the WorkerManager.cs, it seems that the ICollection<Worker> _workers is not locked when receving TriggerIndexingFinished/TriggersDeleted events and so sometime the worker collection doesn't create the worker (or remove it just after adding...).

This avoid the Azure Service Bus trigger to work.

![image](https://user-images.githubusercontent.com/3254429/157129283-8012eb53-12e0-4224-a80e-e0a668b1760f.png)

jdevillard:
Hello @sfmskywalker , 

quick question before proposing a PR on this issue, 
- can you confirm that Mediator is configure with default strategy? ie: loop through the notification handlers and awaits each one
- is there any reason why you lock on Create and not on Remove ? 

Lock on Create using Semaphore

https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs#L36-L52

No Lock on Remove
https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs#L75-L90


I ask about Mediator to validate that lock on remove is enough to solve the issue. If so, I could propose the PR.

sfmskywalker:
Hi @jdevillard - Great catch. We definitely want to lock on remove as well. Mediatr isn't configured with a specific strategy, so I suppose it is using the default one. I'll happily accept your PR!

jdevillard:
Thanks, 

i'll do this today. I've also see a issue on the CreateAndAddWorkerAsync that throw exception.

https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs#L98-L110


[...]

PR #2846, opened 2022-03-16
Fix race condition for TriggerIndexingFinished/TriggersDeleted

jdevillard:
This PR fixes the race condition #2812  by adding a semaphore on RemoveWorkersAsync(Triggers)

### DIFF
--- a/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs
+++ b/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs
@@ -74,18 +74,26 @@ namespace Elsa.Activities.AzureServiceBus.Services
 
         public async Task RemoveWorkersAsync(IReadOnlyCollection<Trigger> triggers, CancellationToken cancellationToken = default)
         {
+            await _semaphore.WaitAsync(cancellationToken);
             var workflowDefinitionIds = Filter(triggers).Select(x => x.WorkflowDefinitionId).Distinct().ToList();
 
-            var workers =
+            try
+            {
+                var workers =
                 from worker in _workers
                 from workflowId in workflowDefinitionIds
                 where worker.Tag == workflowId
                 select worker;
 
-            foreach (var worker in workers.ToList())
+                foreach (var worker in workers.ToList())
+                {
+                    await worker.DisposeAsync();
+                    _workers.Remove(worker);
+                }
+            }
+            finally
             {
-                await worker.DisposeAsync();
-                _workers.Remove(worker);
+                _semaphore.Release();
             }
         }
 


jdevillard:
This PR can be used as is. 

About the second point of #2812,, the use of the latest Azure Service Bus and the use of the processor doesn't throw the exception ( Handler already register), but for me there is a potential starvation because the processor is never disposed. I'll look deeper into this.

jdevillard:
ok for what I've seen, bookmark are deleted in the store in the BookmarkIndexer : 

https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/core/Elsa.Core/Services/Bookmarks/BookmarkIndexer.cs#L95-L99

BUT the event BookmarkDeleted is never fired with the deleted bookmark : 

https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/activities/Elsa.Activities.AzureServiceBus/Consumers/UpdateWorkers.cs#L15

Q/ Should the method call not be 
```
await _workerManager.RemoveWorkersAsync(message.Bookmarks); 
```
and not
```
 await _workerManager.CreateWorkersAsync(message.Bookmarks); 
```

sfmskywalker (approved):

sfmskywalker:
> About the second point of #2812,, the use of the latest Azure Service Bus and the use of the processor doesn't throw the exception ( Handler already register), but for me there is a potential starvation because the processor is never disposed. I'll look deeper into this.

You're right - we need to remove the workflow from the list in the `WorkerManager` when the processor raises an error. One easy way may be to either expose an event from the worker that the worker manager listens for. Alternatively, the worker constructor could accept a callback from the manager, allowing the manager to provide cleanup logic. Conceptually it's the same.

sfmskywalker:
> ok for what I've seen, bookmark are deleted in the store in the BookmarkIndexer :
> 
> https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/core/Elsa.Core/Services/Bookmarks/BookmarkIndexer.cs#L95-L99
> 
> BUT the event BookmarkDeleted is never fired with the deleted bookmark :

That is a big problem. Will fix.

> 
> https://github.com/elsa-workflows/elsa-core/blob/2733a642f258187fb0717182d3ebd2e0d608eea8/src/activities/Elsa.Activities.AzureServiceBus/Consumers/UpdateWorkers.cs#L15
> 
> Q/ Should the method call not be
> 
> ```
> await _workerManager.RemoveWorkersAsync(message.Bookmarks); 
> ```
> 
> and not
> 
> ```
>  await _workerManager.CreateWorkersAsync(message.Bookmarks); 
> ```

Yes, that's another issue. Great find, this might explain some of the strange things I've seen in a production environment.

> final point : I think BookmarkDeleted event is missing after deleting oldBookmark in store.

Yep. I'll fix this too. 

Thank you for these findings!

sfmskywalker:
I pushed a few updates that fixes the issues you mentioned. I'm about to test these shortly. Please let me know if you see any other issues.

PR #2853, opened 2022-03-18
update SqlClient and EFCore.BulkExtensions version

mohdali:
As discussed in https://github.com/elsa-workflows/elsa-core/pull/2841. `Microsoft.Data.SqlClient` v4 introduces breaking changes that affects Sql Server connections.

Both `EntitiyFramework.SqlServer` and newer versions of `EFCore.BulkExtensions` are targeting 2.1.4 as minimum version.

Users are free to choose a higher version by explicitly adding the reference.

### DIFF
--- a/src/activities/Elsa.Activities.SQL/Elsa.Activities.Sql.csproj
+++ b/src/activities/Elsa.Activities.SQL/Elsa.Activities.Sql.csproj
@@ -13,7 +13,7 @@
   </PropertyGroup>
     
   <ItemGroup>
-    <PackageReference Include="Microsoft.Data.SqlClient" Version="4.1.0" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.4" />
   </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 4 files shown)


sfmskywalker:
What about other [changes & improvements](https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#improvements-and-changes) introduced with SqlClient 4.0 - don't want want to take advantage of that?

I realize there will be pain involved when upgrading, but at least these errors will happen immediately, and we can add a big breaking note to the release notes about this.

mohdali:
> What about other [changes & improvements](https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#improvements-and-changes) introduced with SqlClient 4.0 - don't want want to take advantage of that?
> 
> I realize there will be pain involved when upgrading, but at least these errors will happen immediately, and we can add a big breaking note to the release notes about this.

If user wishes to use v4 they can explicitly add reference to v4. It will then be used by both EFCore and SQL Activity.

I don't think there is any code dependency that currently requires v4.

.NET 7 / EF Core 7 preview refer v4 though, so problem will happen soon anyways if Microsoft doesn't change the behavior.

sfmskywalker:
> If user wishes to use v4 they can explicitly add reference to v4. It will then be used by both EFCore and SQL Activity.

Good point. 

> .NET 7 / EF Core 7 preview refer v4 though, so problem will happen soon anyways if Microsoft doesn't change the behavior.

OK. So perhaps we could wait for .NET 7. More people will be aware of this breaking change at that time.

sfmskywalker (approved):

Issue #2604, opened 2021-12-19
Activity Input property is null but designer shows a selected value when options are provided


mohdali:
When defining a property as an `ActivityInput` and providing it with options, the designer will show the options as a select box with the first option already selected.

However, the property remains null unless the selection is changed to a different value.

As a workaround, the property can be assigned a default value or the options may include a blank choice.

However, from user point of view it is misleading.

Example:

```cs
[ActivityInput(UIHint = ActivityInputUIHints.Dropdown, 
   Hint = "Select the Order By",
   Options = new[] { "ASC", "DESC" },
   SupportedSyntaxes = new[] { SyntaxNames.Literal, SyntaxNames.JavaScript, SyntaxNames.Liquid })]
public string OrderBy { get; set; }
```

It appears in the designer like this, which gives the impression that "ASC" is already selected. It will also not assign the value "ASC" unless the user changes to "DESC" then back to "ASC".
![image](https://user-images.githubusercontent.com/4145465/146679935-0898a770-bb72-4acb-a39c-5f92b4b0e7f8.png)

mohdali:
hi @sfmskywalker, any plan for this bug? Earlier it was targeted for 2.5.

I have noticed something similar in Cron activity where the default syntax provided in the activity is not captured as an input unless it is changed to something else.

sfmskywalker:
Hi, unfortunately I had to remove it to not delay 2.5 too much. I'm afraid I won't be able to get it in 2.6 either (which was originally slated as 2.5.1 for the end of last month). But let me assign it to the 2.7 milestone. And of course, if someone is able to work on this before 2.6 is released, it will be included as well :)

sfmskywalker:
Yeah probably something like that. I remember this issue was solved for some other control in the past [like this](https://github.com/elsa-workflows/elsa-core/commit/29b382eced12571300278935a46e81c7dcf9f3a0#diff-3e16fd621b926f0d4ecd26d91097b6a0f01939cecdd11b08b8de01e23f263701).

mohdali:
> Yeah probably something like that. I remember this issue was solved for some other control in the past [like this](https://github.com/elsa-workflows/elsa-core/commit/29b382eced12571300278935a46e81c7dcf9f3a0#diff-3e16fd621b926f0d4ecd26d91097b6a0f01939cecdd11b08b8de01e23f263701).

I will give it a try. Hopefully can get it done quickly.

mohdali:
[...]

PR #2857, opened 2022-03-18
Fix dropdown default value from select list

mohdali:
This is a quick fix for issue https://github.com/elsa-workflows/elsa-core/issues/2604

I tested with Join activity.

By default the designer UI has Mode set to undefined until change event is triggered. The activity still works because enum type takes first value as default.


sfmskywalker (approved):

zoomz00m:
@sfmskywalker good afternoon. Looks like after some more changes made to the dropdown, this fix doesn't work anymore. If I don't reselect elements in dropdown inputs, they are staying null in ActivityData. Doesn't matter if I use OptionsProvider or Options. The screenshot is when I fire the SendHttpRequest, just for the proof
image

Issue #2871, opened 2022-03-23
Composite Activity monitoring show wrong timing and occurence


jdevillard:
I use CompositeActivity and see that monitoring is not showing the good information about the timing of the activity and the nuber of occurence.

To demonstrate the issue, here are some code and reflexion.

Let's go with a quick sample of Composite activity. The aims of this activity is to simulate some delay.
```csharp
public class CompositeDelay : CompositeActivity
    {
        public override void Build(ICompositeActivityBuilder builder)
        {
            builder
                .WriteLine("Starting Composite Activity")
                .Then<Inline>(setup:setup => setup.Set(x=>x.Function,async (c)=> {
                    await System.Threading.Tasks.Task.Delay(4 * 1000);
                    return Done();
                    }
                ))
                .WriteLine("End of the composite activity");
        }
    }
```

if we integrate this activity in a sample workflow : 
![image](https://user-images.githubusercontent.com/3254429/159809083-f031f8a3-b730-401f-81ff-4311a60a49aa.png)

after testing the workflow , we get the following information on the Instance Viewer : 
![image](https://user-images.githubusercontent.com/3254429/159809183-0815cf97-6f83-4498-8f3c-119f3dd4e19d.png)
_The subactivities are not shown on the Designer but we can see them in the Journal Panel_

if we take a look at the WorkflowExecutionLogRecords Table, we can see some information : 
![image](https://user-images.githubusercontent.com/3254429/159809351-f71e11e7-62b5-4402-9f53-40a5e8079e0b.png)

Indeed the activity `CompositeDelay` seems to execute twice and for each the executing time is few ms.
We can see that the subactivity `Inline` has an expected duration of 4 sec but these information are not used in the tracking.

After digging a little bit :  
- The Activity Stats are retrieve using the ActivityStats API which do the following work : 
  - order each activity by id and  time desc
  - get the difference between Executed and Executing (the row before) to get the execution duration
  - count the number of Executed/Executing for an activity id to get the occurrence measurement.

- An CompositeActivityBuilder is used to create the subactivity when creating the Blueprint of the CompositeActivity : 
  - Here the Builder seems to create a **Pre-Activity** with an _Enter_ outcome and a **Post-Activity** with the expected outcome of the Composite Activity. These two activity have the same AcitvityId/Name so that is why the occurrence are multiply by 2 for each use of this activity. Also, these two activity doesn't execute the implementation of the Composite Activity so the duration is just few ms.

[...]

PR #2882, opened 2022-03-25
Fix Composite Activity Timing/Occurrence and measurement in UI Journal

jdevillard:
This PR aims to fix the composite Activity Timing/Occurrence in the stats. (refer to #2871 )

Because the use of Composite activity create two activities before/after the child activities, the statistics are wrong : 
- occurrence are x2
- Timing doesn't reflect the duration of all the child activities.

The proposal is to filter the fired events for Composite activity to be sure to fire only :
- the beginning of the first composite activity
- the end of the last composite activity

With this remark, we can see that begin/end are not necessary in a straightforward sequence, so UI journal for composite activity also need to be updated (front). --> about this remark, maybe this information should be taken from the stats api?


sfmskywalker (approved):
This is great!

PR #2897, opened 2022-03-29
Credential manager

andrewkorol:


src/activities/Elsa.Activities.SQL/Activities/SqlQuery/ExecuteSqlQuery.cs
```diff
@@ -19,15 +26,15 @@ namespace Elsa.Activities.Sql.Activities
...
+    public class ExecuteSqlQuery : Activity, IActivityPropertyOptionsProvider, IRuntimeSelectListProvider
     {
         /// <summary>
         /// Allowed databases to run SQL
         /// </summary>
         [ActivityInput(
             UIHint = ActivityInputUIHints.Dropdown,
             Hint = "Allowed databases to run SQL.",
-            Options = new[] { "MSSQL Server", "PostgreSql" },
+            Options = new[] { "", "MSSQLServer", "PostgreSql" },
```
mohdali, line 37: Is this still required? I submitted a fix in Elsa 2.6 to make the first option selected by default.

mohdali:
Can this be completed anytime soon? I have similar requirement to centrally manage the connection strings especially when moving from staging to production environment and need to update the connection string in each activity.

constantine-v:
@mohdali sorry it took so much time to get back to you, developer who was working on the feature is no longer available, so it will take some more time before it can be finalized and merged.
@sfmskywalker could you please do another round of review? I've merged number of fixes that were missing in initial PR (a2ec1397015eb04e4182e49a55cfa8b7faaadd10).

src/modules/secrets/Elsa.Secrets/Models/Secret.cs
```diff
@@ -0,0 +1,12 @@
+using Elsa.Models;
+
+namespace Elsa.Secrets.Models
+{
+    public class Secret : Entity
+    {
+        public string Type { get; set; } = default!;
+        public string? Name { get; set; }
+        public string? DisplayName { get; set; }
+        public string PropertiesJson { get; set; }
```
sfmskywalker, line 10: Suggestion: rename to `Data` and change its type to `object` or `JObject` or a dictionary. This way, serialization only needs to happen at the DB write/read step.

src/modules/secrets/Elsa.Secrets/Options/SectersOptions.cs
```diff
@@ -0,0 +1,17 @@
+using Elsa.Secrets.Persistence;
+using Elsa.Secrets.Persistence.InMemory;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+
+namespace Elsa.Secrets.Options
+{
+    public class SectersOptions
```
sfmskywalker, line 8: Type, needs to be `SecretsOptions`.

src/modules/secrets/Elsa.Secrets/Persistence/Decorators/EventPublishingSecretsStore.cs
```diff
@@ -0,0 +1,96 @@
...
+        public async Task DeleteAsync(Secret entity, CancellationToken cancellationToken = default)
+        {
+            await _mediator.Publish(new SecretDeleting(entity), cancellationToken);
+            await _store.DeleteAsync(entity, cancellationToken);
+            await _mediator.Publish(new SecretDeleted(entity), cancellationToken);
+        }
+
+        public async Task<int> DeleteManyAsync(ISpecification<Secret> specification, CancellationToken cancellationToken = default)
+        {
+            var webhookDefinitions = await FindManyAsync(specification, cancellationToken: cancellationToken).ToList();
```
sfmskywalker, line 37: Should probably be `secrets` :)

src/modules/secrets/Elsa.Secrets/Persistence/Decorators/InitializingSecretsStore.cs
```diff
@@ -0,0 +1,73 @@
...
+using Elsa.Secrets.Models;
+using Elsa.Services;
+using System.Threading.Tasks;
+using System.Threading;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace Elsa.Secrets.Persistence.Decorators
+{
+    public class InitializingSecretsStore : ISecretsStore
```
sfmskywalker, line 11: This looks like a blind copy & paste. There's no obvious reason to have this decorator class, unless I'm missing something.

src/modules/secrets/Elsa.Secrets/SecretsOptionsBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public SecretsOptionsBuilder(IServiceCollection services)
+        {
+            SectersOptions = new SectersOptions();
+            Services = services;
+            services.TryAddSingleton<ICacheSignal, CacheSignal>();
+        }
+
+        public IServiceCollection Services { get; }
+        public SectersOptions SectersOptions { get; }
```
sfmskywalker, line 20: Type, `SectersOptions ` should be `SectretsOptions`.

sfmskywalker (changes requested):

src/activities/Elsa.Activities.Sql/Activities/SqlCommand/ExecuteSqlCommand.cs
```diff
@@ -41,29 +48,46 @@ public class ExecuteSqlCommand : Activity
...
-            Hint = "Connection string to run SQL",
-            SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid }
-        )]
-        public string ConnectionString { get; set; } = default!;
+            UIHint = ActivityInputUIHints.Dropdown,
+            Label = "Connection string",
+            OptionsProvider = typeof(ExecuteSqlCommand),
+            SupportedSyntaxes = new[] { SyntaxNames.Literal, SyntaxNames.JavaScript, SyntaxNames.Liquid }
+         )]
+        public string? CredentialString { get; set; }
```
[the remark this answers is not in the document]
constantine-v: Renamed to ConnectionString.

src/modules/secrets/Elsa.Secrets/Extentions/SecretExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+                    {
+                        return ConvertToCredentialString(secret.PropertiesJson, "=", ";", includePropertyNamesInValues: true);
+                    }
+                case SecretType.AuthorizationHeader:
+                    {
+                        return ConvertToCredentialString(secret.PropertiesJson, ":", string.Empty, includePropertyNamesInValues: false);
+                    }
+                case SecretType.Ampq:
+                default: return string.Empty;
+            }
```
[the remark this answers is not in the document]
constantine-v: I introduced ISecretValueFormatter interface (and an implementation for each known secret type), which in their turn are injected in SecretsManager (as IEnumerable<ISecretValueFormatter> valueFormatters). Whenever new secret type is added, all that is needed is to register another ISecretValueFormatter implementation (what can be done even outside of Secrets module).

src/modules/secrets/Elsa.Secrets/Extentions/SecretType.cs
```diff
@@ -0,0 +1,10 @@
+namespace Elsa.Secrets.Extentions
+{
+    public static class SecretType
+    {
+        public const string Ampq = "AMPQ";
+        public const string MsSql = "MSSQLServer";
+        public const string PostgreSql = "PostgreSql";
+        public const string AuthorizationHeader = "Authorization";
```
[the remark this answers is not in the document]
constantine-v: I still want to keep those constants (just not to specify secret type in code explicitly). I renamed the file to Constants.cs and renamed each property, so it won't look like enumerator.

src/activities/Elsa.Activities.Http/Activities/SendHttpRequest/SendHttpRequest.cs
```diff
@@ -231,14 +240,20 @@ private HttpRequestMessage CreateRequest()
...
+                case nameof(ResponseContentParserName):
+                {
+
+                    var items = _parsers.Select(x => new SelectListItem(x.Name, x.Name)).ToList();
+
+                    items.Insert(0, new SelectListItem("Auto Select", ""));
+                    return items;
+                }
+                case nameof(ResponseContentTargetType): return null;
+                default: throw new ArgumentException($"Unsupported property: {property.Name}");
```
sfmskywalker, line 255: I don't think it's a good idea to throw an exception here, because if we add a new property to the activity, we have to add a case here for no other reason as not to throw.

Since this method only applies to the `ResponseContentParserName` property, I suggest to revert this change to its original form (unless you added something useful that I missed?)

src/activities/Elsa.Activities.Sql/Activities/SqlCommand/ExecuteSqlCommand.cs
```diff
@@ -26,7 +26,7 @@ public class ExecuteSqlCommand : Activity
         [ActivityInput(
             UIHint = ActivityInputUIHints.Dropdown,
             Hint = "Allowed databases to run SQL.",
-            Options = new[] { "MSSQL Server", "PostgreSql" },
+            Options = new[] { "", "MSSQLServer", "PostgreSql" },
```
sfmskywalker, line 29: This change will break existing users who are using this activity in production. I suggest we only add the empty string and leave the rest of the values unchanged (and perhaps open a separate PR where we can distinguish between technical name and display name if necessary).

src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/sqlserver.activity.ts
```diff
@@ -0,0 +1,99 @@
+import { Secret } from "../secret.model";
```
sfmskywalker, line 1: I think the filename is incorrect.

src/designer/elsa-workflows-studio/src/modules/credential-manager/services/credential-manager.client.ts
```diff
@@ -0,0 +1,67 @@
+import axios, {AxiosInstance, AxiosRequestConfig} from "axios";
```
sfmskywalker, line 1: I don't think we should copy the entire elsa client ts file. Surely we can reuse the existing one?

src/modules/secrets/Elsa.Secrets/Elsa.Secrets.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <LangVersion>default</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netcoreapp3.1|AnyCPU'">
+    <LangVersion>default</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net6.0|AnyCPU'">
+    <LangVersion>default</LangVersion>
+  </PropertyGroup>
```
sfmskywalker, line 21: Can't we use the `default` language version for all frameworks?

Also, I think we should import props here like we do in all the other class libraries.

sfmskywalker (changes requested):
Getting there! Just a few minor requests :)

src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/sqlserver.activity.ts
constantine-v: In this case, file is hosting SecretModel interface, so I think it is correct (I'm guessing that you are referring to one of the previous issues, where word mod**a**l was mistyped :)).

src/modules/secrets/Elsa.Secrets/Extentions/SecretExtensions.cs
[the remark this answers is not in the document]
sfmskywalker: Great! 🎉

src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/sqlserver.activity.ts
sfmskywalker: You're absolutely right, my mistake 👍🏻

sfmskywalker (approved):

PR #2921, opened 2022-04-07
Fixing the Azure Worker Processor

jruckert:
As the worker is created under WorkerManager.cs as per
```
worker = ActivatorUtilities.CreateInstance<Worker>(_serviceProvider, queueOrTopic, subscription ?? "", RemoveWorkerAsync);
```
The case of
```
subscription == null 
```
will never be fired and it will try to process a subscription with a empty name. This causes the Error event handler to be called which then disposes of the _processor early resulting in the following error.

System.ObjectDisposedException: The message processor is unable to continue and will stop processing; the host application has closed the connection to the Service Bus service. Object name: 'ServiceBusConnection'.

This pull request fixes this scenario where a queue is used, and not a topic/subscription model.


sfmskywalker (approved):
Great catch 👍🏻

PR #2927, opened 2022-04-08
Add Swagger integration for HTTP Endpoint activities

### LINKED ISSUE #1499: HTTP Endpoints Swagger
Currently Swashbuckle only looks at controller actions when building an Open API specification. We should look into the possibility of telling Swashbuckle to also look at HTTP Endpoint activities declared on workflow blueprints. This would provide a complete view of all of the exposed API endpoints when generating the specification.

sfmskywalker:
This PR adds a new package called `Elsa.Activities.Http.OpenApi` that enables integration with Swashbucke's Swagger Gen that adds HTTP Endpoint descriptions.
To enable this Swagger integration, do the following:

1. Add a package reference to `Elsa.Activities.Http.OpenApi` (introduced with this PR).
2. Add the new `HttpEndpointDocumentFilter` when configuring swagger. E.g. `services..AddElsaSwagger(options => options.DocumentFilter<HttpEndpointDocumentFilter>())`.

When you create workflows that start with the HTTP Endpoint activity, the path + method of these endpoints are now automatically listed.

For example, given the following HTTP Endpoint:

image

Will add the following Open API path & operations:

image

Note: additional work might be done to allow further control over additional details being provided by HTTP Endpoint activities, such as describing the supported content types, adding support for JSON schema, etc.

Closes #1499


hpposch:
Hi @sfmskywalker i really love this feature!
I was just wondering if you plan to split the 'Workflow Endpoints' from all the other Endpoints that are visible in swagger.
Or maybe a seperate swagger json that holds only the workflow endpoints.
In my point of view it is kind of a security issue when they show up together.
I was also wondering if it is possible to 'lock' or 'hide' the internal Endpoints?

Best regards and thanks for this great feature ;)

Juandavi1:
awesome !

Issue #1311, opened 2021-07-26
Version Management


sfmskywalker:
Workflow definitions are versioned, which means that every time a new change is published, a new version is created.

User should be able to view past versions, roll back to them (into a new version) and permanently delete versions.

tomy2105:
Another very useful (but tricky to implement) feature would be the ability to compare between versions.

sfmskywalker:
@tomy2105  I love that. Could be done as a (big) improvement down the line.

PR #3026, opened 2022-05-13
Manage & revert workflow definitions versions through UI

### LINKED ISSUE #2572: Allow listing of historical versions of process definitions with option to replace the draft with a historical version
It is very useful to view and revert a historical version. As all historical versions are already available in the database.

**Suggestion**
- Allow designers to see previous workflow definition versions
- Allow deleting of a historical version 
     => e.g. very useful if there ia historical version with hardcoded credentials.
- Allow opening a selected historical version 
    => its handy to allow getting some information from previous versions, it should be clear that user has opened a historical version
- Allow reverting a selected historical version 
   => Create new historical version from current draft, make the reverted version as current draft.

**UI/UX Suggestions**
- Please think a long and provide some suggestions as well.
- Add a tab to proces definition in side panel where these functionalities are added.
- Show the user that a historical version (not the active draft/published version) is open, e.g. by changing the bottom-down actions: Revert, Delete, Export, Import, (remove publish). Also show the version information next to the action button.

sfmskywalker:
This PR adds a tab to the workflow definition editor panel that lets the user view older versions of the workflow definition, revert back to them and permanently delete them.

Example video:

[Alt text](https://www.youtube.com/watch?v=O10MUh8SD14)

Closes #2572 and #1311

### DIFF
--- a/src/core/Elsa.Abstractions/Models/WorkflowDefinition.cs
+++ b/src/core/Elsa.Abstractions/Models/WorkflowDefinition.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using NodaTime;
 
 namespace Elsa.Models
 {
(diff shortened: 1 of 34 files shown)


tomy2105:
Great addition! Love your work.
One small detail, when you open completely new workflow what used to be "Publish" button now says "Revert version 1" which kind of doesn't make sense..... :)

sfmskywalker:
Indeed, thanks for catching! :)

Issue #3038, opened 2022-05-18
WorkflowExecutionLog is not cleaned by retention module


mlditga:
Hello,

We are currently trying the retention module with elsa 2.6 and we noticed that the WorkflowExecutionLog collection is not cleaned as well as the WorkflowInstances collection. After a clean it is left with entries with WorkflowInstanceId that does not exist anymore. Would it be possible to add deletion of these too?

Thanks!

sfmskywalker:
Hi @mlditga , which DB provider are you using? I checked out the code, and noticed that the EF Core provider of the `IWorkflowInstanceStore` takes care of deleting related records such as execution log records and bookmarks, but this wasn't done for MongoDB, YesSQL and others.

So perhaps it is better if we update thee providers to have the same behavior as that of EF Core.
That way, it doesn't matter what code is trying to delete a workflow instance - it will always delete related records.

mlditga:
Indeed, we use MongoDB. That would be great. Thank you!

PR #3039, opened 2022-05-18
3038: retention module: delete execution logs

mlditga:
Hello,
This may fix the issue #3038. Let me know if I need to change something.

### DIFF
--- a/src/activities/Elsa.Activities.BlobStorage/Elsa.Activities.BlobStorage.csproj
+++ b/src/activities/Elsa.Activities.BlobStorage/Elsa.Activities.BlobStorage.csproj
@@ -15,6 +15,7 @@
 
     <ItemGroup>
         <PackageReference Include="Storage.Net" Version="9.3.0" />
+        <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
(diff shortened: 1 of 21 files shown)


mlditga:
Hello, 
we may consider updating MongoDB provider but unfortunately we won't be able to do it for YesSQL as we don't know it at all. If we modify mongo provider should we put the deletion in a transaction?
Thank you!

sfmskywalker:
Hi @mlditga 

Just updating the MongoDB provider is enough, you don't have to do it also for YesSQL.
Doing it in a transaction would be nice, but not a hard requirement.

Thank you!

jdevillard:
Hello, 

@mlditga , have you update the mongoDb provider? 

I think this PR should be close without merge and a new PR with only the MongoDb provider should be done.
In the same way, I'm on my side proposing perf enhancement on the retention module.

mlditga:
Hello,
Sorry, no. We ended up implementing a custom code for db cleaning as we had a custom collection anyway :/ But this PR can probably be closed.

sfmskywalker:
Thanks for letting us know @mlditga. We'll close this PR.

PR #3063, opened 2022-05-24
Generic cascading drop down

jdevillard:
This PR is a proposal for Generic Cascading dropdown.

The idea is the following : 
- modify the UI component to be sure that everything could be defined on C# backend (sample is done using hard-coded value but could then be done using a Db)
- a property in an activity defined  :
     - dependent event : which property change trigger a refresh
     - dependent values : which properties are needed with the refresh
- Don't make any assumption on the dependent input type, I think I only need to get its value. 

the change on `1-level` is simple
It's more hard when you want to create a `2-level` or more.
In the sample with `DynamicVehiculeActivity`, `Color` depends on `Brand` and `Model`. 
So it need to refresh when the `Model` changed using `Brand` AND `Model` values. Also, if Brand change --> Model change --> Color need to change with a new init value.

The value are get using `interface IRuntimeSelectListProvider` implementation.

as you can see, It seems that I've some issues with HTML event because some items of the  `Color Collection` are duplicated (also Model when Tesla Brand is chosen).

image

once event issue will be cleaned, following discussion with @sfmskywalker , I think `ActivityInput` could be extend with property `DependsOnEvent`, `DependsOnValue`


jdevillard:
@sfmskywalker 

I've pushed the two properties in the ActivityInputAttribute.

So you can take a look at the event issue on UI 😉

sfmskywalker (approved):

PR #3128, opened 2022-06-13
Search Workflow Definitions

andreabalducci:
image

It's my first attempt on FE with Stencil. 
Manually tested with /samples/dashboard/aspnetcore/ElsaDashboard.Samples.AspNetCore.Monolith project.

Haven't found a FE dev guideline for Elsa, so I went the old good "Search / Copy / Paste / Replace / Test" road.

My workflow is:

1. Make edits on WorkflowStudio web components  (vs code)
2. npm run build (vscode or shell)
3. npm install on AspNet Bindings project (shell)
4. npm run build on AspNet Bindings project (shell)
5. run /samples/dashboard/aspnetcore/ElsaDashboard.Samples.AspNetCore.Monolith (Rider)

Please let me know if the workflow could be streamlined.

### DIFF
--- a/src/core/Elsa.Core/Persistence/Specifications/WorkflowDefinitions/WorkflowDefinitionSearchTermSpecification.cs
+++ b/src/core/Elsa.Core/Persistence/Specifications/WorkflowDefinitions/WorkflowDefinitionSearchTermSpecification.cs
@@ -1,0 +1,17 @@
+using System;
+using System.Linq.Expressions;
+using Elsa.Models;
+
+namespace Elsa.Persistence.Specifications.WorkflowDefinitions;
+
+public class WorkflowDefinitionSearchTermSpecification : Specification<WorkflowDefinition>
+{
+    public WorkflowDefinitionSearchTermSpecification(string searchTerm)
+    {
+        SearchTerm = searchTerm;
+    }
+
+    public string SearchTerm { get; set; }
+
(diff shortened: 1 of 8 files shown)


andreabalducci:
@sfmskywalker there's something I've to check on the FE for layouts? 

Not sure about markup structure (copied from instances search)

tomy2105:
Since this is supposed to be generic search my recommendation would be to search Workflow Display Name and Description too, not just Name.....

In addition, any reason you are merging this into 2.8.x branch and not master?

andreabalducci:
npm install triggers this warning messages (probably rewiring the lock files)

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

I'll leve out the .lockfiles on next PR

andreabalducci:
Moved to #3131

tomy2105:
> @tomy2105 just followed the instructions, npm install updated the refs (clean clone). I'll reissue the PR on master.
Probably the instructions should be more specific that you need to do **npm ci** instead of **npm install** if you don't want to update to the new versions of libs.....

> For search fields was unsure, the fields you mentioned are not visible on search results (and don't know if are indexed).
Actually I think Display Name and Description are what is displayed on the workflow definition screen but I'm not sure.
Never understood why we have two names :) but I guess there is some use case for that....
IMHO Best to include all three....

Issue #3103, opened 2022-06-01
MongoDb not serializing PropertyStorageProviders


stuartmcgillivray:
I noticed there was an issue with the property storage providers not returning the previously set value in the designer i.e. I want my ExampleActivity to store an ActivityOutput in blob storage and clicking save doesn't persist that property to the database.

I had a look into it and it seems the mongo db driver doesn't serialize readonly properties by default so PropertyStorageProviders in ActivityDefinition is by default left out and not saved to mongo.

https://github.com/elsa-workflows/elsa-core/blob/6daa8e456e8a3ab2b20da20d214626f259881b1f/src/core/Elsa.Abstractions/Models/ActivityDefinition.cs#L16

We can opt-in to storing this property by registering a class map, However this only appear to allow it to be serialized, not deserialised so it is still an empty list when it's read back.
```csharp
BsonClassMap.RegisterClassMap<ActivityDefinition>(cm =>
{
    cm.AutoMap();
    cm.MapProperty(c => c.PropertyStorageProviders);
});
```
A solution may be to just add the set in but I don't know if that would have implications in other places.

```csharp
public IDictionary<string, string> PropertyStorageProviders { get; set; } = new Dictionary<string, string>();
```

terry-delph:
Would this registration fix the issue without any changes to the ActivityDefinition class?
```
BsonClassMap.RegisterClassMap<ActivityDefinition>(cm => 
{ 
     cm.AutoMap();
     new ImmutableTypeClassMapConvention().Apply(cm);
});
```

[ImmutableTypeClassMapConvention](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Conventions/ImmutableTypeClassMapConvention.cs)

stuartmcgillivray:
It doesn't, the ImmutableTypeClassMapConvention tries to use constructors to populate the read only values, the activity definition doesn't currently have a constructor.

sfmskywalker:
Closed via #3150

PR #3150, opened 2022-06-22
Mongo db can't write to read only properties

stuartmcgillivray:
This change will fix an issue where PropertyStorageProviders on ActivityDefinition isn't be read or written in MongoDB

https://github.com/elsa-workflows/elsa-core/blob/6daa8e456e8a3ab2b20da20d214626f259881b1f/src/core/Elsa.Abstractions/Models/ActivityDefinition.cs#L16

As discussed in #3103, mongo by default ignores read only properties so the PropertyStorageProviders selected in the designer is never persisted.

Adding the setter allows mongo to read and write and doesn't appear cause any issues elsewhere.


sfmskywalker (approved):
Great catch, thanks!

PR #3151, opened 2022-06-22
[RabbitMQ] Add Javascript Syntax to Headers definition

borjaferv:
Add the posibility to define dyamic headers with JavaScript Syntax


sfmskywalker (approved):
Thanks!

Issue #3008, opened 2022-05-09
Breaking change from 2.6.0 to 2.7.0 using MongoDb


ricksearcy:
There seems to be a serialization issue between the two versions. Older WorkFlowDefintions stored in MongoDB had the CustomAttributes property stored as a string in the DB. 2.7.0 is now having a deserialization issue with those properties because of that at startup.

"An error occurred while deserializing the CustomAttributes property of class Elsa.Models.WorkflowDefinition: Expected a nested document representing the serialized form of a Elsa.Models.Variables value, but found a value of type String instead."

What is stored in the DB for all existing definitions prior to 2.7.0
"CustomAttributes" : "{\"Data\":{}}",

What the variable looks like in the class
public Variables CustomAttributes { get; set; }

An example of an entire definition stored in mongo. 
{
    "_id" : "e1661e0d999f40b4b8424d5e9d2c91eb",
    "_t" : [ 
        "Entity", 
        "WorkflowDefinition"
    ],
    "Variables" : "{\"Data\":{}}",
    "DefinitionId" : "313be0145ee046ea8b668adcdd621165",
    "TenantId" : "26c6ec78-6acd-4f89-b749-28720608c878",
    "Name" : "MikeTest",
    "DisplayName" : "MikeTest",
    "Description" : null,
    "Channel" : null,
    "Version" : 5,
    "CustomAttributes" : "{\"Data\":{}}",
    "ContextOptions" : null,
    "IsSingleton" : false,
    "PersistenceBehavior" : 1,
    "DeleteCompletedInstances" : false,
    "IsPublished" : false,
    "IsLatest" : false,
    "Tag" : null,
    "Activities" : [ 
        {
            "ActivityId" : "222d17ab-f956-415b-8c46-fde64f8b232f",
            "Type" : "WriteHttpResponse",
            "Name" : null,
            "DisplayName" : "HTTP Response",
            "Description" : null,
            "PersistWorkflow" : false,
            "LoadWorkflowContext" : false,
            "SaveWorkflowContext" : false,
            "Properties" : [ 
                {
                    "Name" : "Content",
                    "Syntax" : null,
                    "Expressions" : {
                        "Literal" : "<h1>Hello World</h1>\nThis is a user created workflow example in Coras!",
                        "JavaScript" : ""
                    }
                }, 
                {
                    "Name" : "ContentType",
                    "Syntax" : null,
                    "Expressions" : {
                        "Literal" : "text/html"
                    }
                }, 
[...]

Issue #2998, opened 2022-05-06
Update to 2.7 breaks on Mongo


andreabalducci:
https://github.com/elsa-workflows/elsa-core/blob/f2fa88f42f30e9db4f68d948ca02b9f8d28889b1/src/persistence/Elsa.Persistence.MongoDb/Services/DatabaseRegister.cs#L63

LookupSerializer is a lazy registration, this line of code adds (internally mongo use GetOrAdd), so upgrading to 2.7 break reads of previously saved workflows instance

As workaround had to manually register the serializers in startup.cs

![elsa-error](https://user-images.githubusercontent.com/212729/167092219-3ebf95f1-52e8-4299-aeec-3c55773d171b.png)
s

Jack-Hartman:
Hello, I am also getting a 2.7 Mongo break but with loading the workflow instances page and trying to execute a workflow. 
Seems to break on deserializing the NodaTime CreatedAt property of a WorkflowInstance.
I've managed to get around this issue so far by using the MongoDb.Bson.NodaTime package and registering the NodaTimeSerializers.
<details>
<summary> Error Message </summary>

```
[10:03:50.263 ERR][18] An unhandled exception has occurred while executing the request.
System.FormatException: An error occurred while deserializing the CreatedAt property of class Elsa.Models.WorkflowInstance: Value class NodaTime.Instant cannot be deserialized.
 ---> MongoDB.Bson.BsonSerializationException: Value class NodaTime.Instant cannot be deserialized.
   at MongoDB.Bson.Serialization.BsonClassMapSerializer`1.Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
   at MongoDB.Bson.Serialization.Serializers.SerializerBase`1.MongoDB.Bson.Serialization.IBsonSerializer.Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
   at MongoDB.Bson.Serialization.IBsonSerializerExtensions.Deserialize(IBsonSerializer serializer, BsonDeserializationContext context)
   at MongoDB.Bson.Serialization.BsonClassMapSerializer`1.DeserializeMemberValue(BsonDeserializationContext context, BsonMemberMap memberMap)
   --- End of inner exception stack trace ---
   at MongoDB.Bson.Serialization.BsonClassMapSerializer`1.DeserializeMemberValue(BsonDeserializationContext context, BsonMemberMap memberMap)
[...]

PR #3170, opened 2022-06-27
Mongo serializer registration

stuartmcgillivray:
Correcting issue #2998.

Adding try catch around the registration instead of if (BsonSerializer.LookupSerializer<JObjectSerializer>() == null).

As mentioned in the issue, BsonSerializer.LookupSerializer is not being used for the correct purpose here.

https://github.com/elsa-workflows/elsa-core/blob/f2fa88f42f30e9db4f68d948ca02b9f8d28889b1/src/persistence/Elsa.Persistence.MongoDb/Services/DatabaseRegister.cs#L63-L64

It will try and find a serializer for VariablesSerializer (MongoDB.Bson.Serialization.BsonClassMapSerializer<Elsa.Persistence.MongoDb.Serializers.VariablesSerializer>) if it doesn't find one then it will create one and return that.

This means those serializers never get registered.

I believe this issue is also the cause of https://github.com/elsa-workflows/elsa-core/issues/3008

### DIFF
--- a/src/persistence/Elsa.Persistence.MongoDb/Services/DatabaseRegister.cs
+++ b/src/persistence/Elsa.Persistence.MongoDb/Services/DatabaseRegister.cs
@@ -3,6 +3,7 @@ using Elsa.Models;
 using Elsa.Persistence.MongoDb.Serializers;
 using Elsa.Services.Models;
 using MongoDb.Bson.NodaTime;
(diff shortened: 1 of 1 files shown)


sfmskywalker (approved):

sfmskywalker:
Thanks for fixing this!

PR #3177, opened 2022-06-27
If eager join is set, clear scheduled activities on join

martinhjulstrom:
Another aspect of the recently added eager join option is that of branch order. You will get a little bit different behavior dependent on in which branch the block activity is. If the blocking activity is executed before the joining branch it will be cleared, but if the blocking activity is executed after the joining branch it will be blocking after the execution, because it was not blocking at the time of the join, it was scheduled.

With this PR a join of an EagerJoin will also clear all scheduled activities inside the fork, effectively stopping all execution.

If you don't think that this should be the behavior of EagerJoin, perhaps this could be added as another option to the Join?


sfmskywalker:
Yep, I think the eager join should clear all scheduled activities inside the fork and the join 👍🏻

PR #3178, opened 2022-06-29
Jint options configuration

stuartmcgillivray:
Adding notification to allow the jint options to be intercepted as part of the configuration Action

### DIFF
--- a/src/scripting/Elsa.Scripting.JavaScript/Messages/ConfigureJavaScriptOptions.cs
+++ b/src/scripting/Elsa.Scripting.JavaScript/Messages/ConfigureJavaScriptOptions.cs
@@ -1,0 +1,16 @@
+﻿using Elsa.Services.Models;
+using MediatR;
+
+namespace Elsa.Scripting.JavaScript.Messages
+{
+    public class ConfigureJavaScriptOptions : INotification
+    {
+        public ConfigureJavaScriptOptions(Jint.Options options, ActivityExecutionContext context)
+        {
+            JintOptions = options;
+            ActivityExecutionContext = context;
+        }
+        public Jint.Options JintOptions { get; set; }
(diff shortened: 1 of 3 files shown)


terry-delph:
This will allow an application to set up the engine for scenarios such as using the "map" and "filter" functions.

- https://github.com/sebastienros/jint/blob/a800bd32e7f534b7f026cb0c6f46d8e95308da5f/Jint.Tests/Runtime/InteropTests.cs#L934
- https://github.com/sebastienros/jint/issues/1030#issuecomment-999322450

Without the engine options being configurable, some functions are not able to be used.

sfmskywalker (approved):

maltegoetz:
Great to find this PR, just came across the same need. This will allow one to enable extension methods which is pretty amazing!
@sfmskywalker Are there any plans when a new release might be rolled out, so that we can make use of this?

jdevillard:
Hello, 

You can use the preview nuget to get the change if you don't want to wait for the release 


> @sfmskywalker Are there any plans when a new release might be rolled out, so that we can make use of this?

terry-delph:
> Yes, we wanted to release 2.9 as soon as (at least some of) the bugs assigned to the 2.9 milestone are fixed. But if the current 2.9 already contains enough useful features and improvements, then we could release now and reassign the bugs to the next milestone.

My team has taken a snapshot of a preview version in MyGet so we can move to Production with lots of the bug fixes we identified and fixed.
We would prefer to have a 2.9 release without the preview, so if a earlier 2.9 release could happen, that would be really appreciated :)
I will see if my team can pick up some of the other issues that would be moved to the next milestone soon.

sfmskywalker:
Sure, if 2.9 already contains valuable fixes and improvements then let's release sooner rather than later 👍🏻 
I'll try to do this soon. Any help you might offer is much appreciated 🙌🏻

Issue #3180, opened 2022-06-30
Workflow instance retry issue


reiven93:
I have a workflow with some activities and a custom one. That custom activity just does return Fault();
I'm using Hangfire Temporal activities. What I see is that if I press retry on the Elsa studio dashboard or using the retry API, the workflow status changes to Running but with a red label. Then, a warning appears in console of the Elsa server:
```
warn: Elsa.Services.Workflows.WorkflowInstanceExecutor[0]
      Did not run workflow f8575e9b296b4e75936eea2b1a6fba93 for activity 1b87793b-f7f6-4793-a377-de43550f2e47 because the workflow is not blocked on that activity nor is that activity scheduled for execution
```
If I check my workflow, the activity that is shown in the log is not the Faulted one, but the first of the workflow, so it fails. As I'm using Hangfire temporal activities, a job is created and it shows the wrong activityId as an argument, matching the one that the elsa warning log shows. So, it seems that the dispatcher is not sending the correct activityId to revive. Looking into the code, seems that CurrentActivity is null here and, when this happens, it gets the first activity of the workflow:

![imagen](https://user-images.githubusercontent.com/46783038/176661605-14975c3c-1467-49ff-bdf7-3abf0f15ec08.png)

When my workflowinstance is Faulted, I can see in database that there is a Faulted activity in the workflowinstances row of my workflow. Then, when retrying, the database shows what is expected to be scheduled right:

![imagen](https://user-images.githubusercontent.com/46783038/176661852-6627f35b-bceb-4637-92c5-f9192b9b8f42.png)

This is the ActivityId of the faulted Activity. But as I said, it dispatches the first of the workflow.

Because Elsa has a server down protection, if I just restart Elsa server, the workflow is auto dispatched with the faulted activity right and the execution continues.

sfmskywalker:
@smartwaystudio is working on this 💪🏻

Issue #1059, opened 2021-05-28
Introduce Multiple & Nested Execution Scopes


sfmskywalker:
Currently it is not possible for Elsa to handle multiple execution scopes, which makes activities such as **Parallel ForEach** very limited as reported in #1025 and #1056

### Execution Scope

An execution scope would maintain its own set of variables and blocking activities, so that when a blocking activity is triggered, it is resumed in the right scope with access to variables and other state recorded at the time of suspension.

Having execution scopes also simplify the "unwinding" logic that is currently in place for activities such as **If**, **While** and **For** and **ForEach** by having clear boundaries implemented as a scope.

Nokecy:
Need this feature ~~~ @sfmskywalker

ddydeveloper:
Just wondering would it be working solution to have a couple of variables like `numberOfIterations` (set once when ParallelForEach defined) and `finishedIterations` (update it when each iteration completed) inside `ActivityExecutionContext` which we could use to compare and unblock workflow.

```
...
.ParallelForEach(() => {
   // set numberOfIterations
})
.WaitAll(() => {
   // compare numberOfIterations with finishedIterations
})
.Then<Activity>
...
```

sfmskywalker:
Something like that could work. It would be similar to the way the `Join` activity blocks until all inbound connections have executed (when using `JoinMode.WaitAll`. But I don't think it will solve the issue of not being able to block within an iteration branch itself. Technically speaking, one could block (one bookmark per iteration would be created) - but the issue is when you want to resume one of the blocked iterations - there's no context being tracked. 

In Elsa 3, each parallel execution will have its own activity execution context, which can store contextual information such as the current value of the iteration (and other variables scoped to the iteration).

maltegoetz:
@sfmskywalker Is there a reason why this was removed from Elsa 3.0 milestone? This seems pretty important to me.

sfmskywalker:
@maltegoetz I removed all issues from the milestone in order to re-evaluate what should be part of the initial milestone and do some backlog grooming. And you're right, this issue is important and will be added back.

sfmskywalker:
Parallel ForEach is now available in Elsa 3 via https://github.com/elsa-workflows/elsa-core/pull/4306

PR #3182, opened 2022-07-04
Fault array, fix retry, new parameter in Join activity

smartwaystudio:
**This PR has breaking changes**

Hi, I will try to set up my needs to explain what is this PR about. We are migrating our app core from custom Hangfire jobs to Elsa workflows with custom activities to enrich our app and also to have the ability to modify the workflows without making changes on the code. So, the first thing I did was to copy my hangfire flow to custom activities on Elsa. Everything was OK and the workflow was being executing with the same results. Then, I started do dig more, so, I purpossely failed some activities (custom and fault activity) to try the retry functionality and it wasn't working. Already opened an [issue with the problem ](https://github.com/elsa-workflows/elsa-core/issues/3180). If I couldn't retry my failed activities, I would need to give up elsa, so I started to checking the core code. I saw that Workflowinstance.CurrentActivity was always null, so the reviver was always using the first workflow activity as retry. But also found another problem (at least in my scenario). Elsa was only saving to database the last fault activity, so, in a loop or fork, just the last fault activity was being tagged as faulted, the others were OK. I changed the fault attribute to faults as SimpleStack. Now, I can save all the faulted activities. And, then, changed the reviver to check if there are faulted activities to schedule on retry. Now it's working good, even the retry. Now you can retry all the faulted activities inside a loop. 
imagen
imagen

(Be careful with how you design your loop. If the activities input are dependant on other activities, for example, the output of the loop, it won't work as expected, because it will use the last output of the loop as input for every activity that uses "return foreach.Output()". The way to solve this is to transmit the input from the first activity in the loop to the input of the next activity, just returning it as output.)

Also, I found another problem. In a fork scenario, the final join continues the execution (with both configuration modes) even if the inbound activities has failed (tagged as faulted). I added a new parameter (bool) to the Join activity that prevents that. It checks if there are fault or scheduled activities (this change is chained with the previous) and stop the execution. Just changed IsDone method. Now, it will add to the journal the activities that have faulted, like the inbounds journal data.
imagen

There are 2 activity list checked on isDone. Faults is used to know if there are some faults in the current workflowinstance execution. The Scheduled activity list is used on retry. When you retry a faulted workflow, it moves the activities from Faults to scheduled list.

Finally, I updated the designer with the new Faults parameter (was Fault) and now, all the faulted activities in the designer are tagged red.
imagen

### DIFF
--- a/src/core/Elsa.Abstractions/Services/Models/WorkflowExecutionContext.cs
+++ b/src/core/Elsa.Abstractions/Services/Models/WorkflowExecutionContext.cs
@@ -188,7 +188,7 @@ namespace Elsa.Services.Models
             var clock = ServiceProvider.GetRequiredService<IClock>();
             WorkflowInstance.WorkflowStatus = WorkflowStatus.Faulted;
             WorkflowInstance.FaultedAt = clock.GetCurrentInstant();
-            WorkflowInstance.Fault = new WorkflowFault(SimpleException.FromException(exception), message, activityId, activityInput, resuming);
+            WorkflowInstance.Faults.Push(new WorkflowFault(SimpleException.FromException(exception), message, activityId, activityInput, resuming));
             Exception = exception;
         }
 
@@ -197,7 +197,7 @@ namespace Elsa.Services.Models
             var clock = ServiceProvider.GetRequiredService<IClock>();
             WorkflowInstance.WorkflowStatus = WorkflowStatus.Cancelled;
             WorkflowInstance.CancelledAt = clock.GetCurrentInstant();
-            WorkflowInstance.Fault = new WorkflowFault(SimpleException.FromException(exception), message, activityId, activityInput, resuming);
+            WorkflowInstance.Faults.Push(new WorkflowFault(SimpleException.FromException(exception), message, activityId, activityInput, resuming));
         }
 
         public async Task CompleteAsync()
@@ -218,7 +218,7 @@ namespace Elsa.Services.Models
                 WorkflowInstance.WorkflowStatus = WorkflowStatus.Cancelled;
                 WorkflowInstance.CancelledAt = now;
             }
-            if(WorkflowInstance.Fault != null)
+            if(WorkflowInstance.Faults != null && WorkflowInstance.Faults.Count > 0)
             {
                 WorkflowInstance.WorkflowStatus = WorkflowStatus.Faulted;
                 WorkflowInstance.FaultedAt = now;
--- a/src/core/Elsa.Core/Activities/ControlFlow/Join/Join.cs
+++ b/src/core/Elsa.Core/Activities/ControlFlow/Join/Join.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -47,6 +47,10 @@ namespace Elsa.Activities.ControlFlow
         )]
         public JoinMode Mode { get; set; }
 
+
+        [ActivityInput(Hint = "True if incoming activities executed should not be Faulted to continue.", UIHint = ActivityInputUIHints.SingleLine, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })]
+        public bool WaitFaulted { get; set; } = default!;
+
         public IReadOnlyCollection<string> InboundTransitions
         {
             get => GetState<IReadOnlyCollection<string>>(() => new List<string>());
@@ -55,11 +59,16 @@ namespace Elsa.Activities.ControlFlow
 
         protected override async ValueTask<IActivityExecutionResult> OnExecuteAsync(ActivityExecutionContext context)
         {
-            var isDone = IsDone(context);
+            var isDone = IsDone(context, WaitFaulted);
             
             context.JournalData.Add("Completed", isDone);
             context.JournalData.Add("Current Inbound Transitions", InboundTransitions);
-            
+
+            if (context.WorkflowInstance.Faults.Count > 0)
+            {
+                context.JournalData.Add("Current Faulted Transitions", context.WorkflowInstance.Faults.Select(x => x.FaultedActivityId));
+
+            }
             if (!isDone)
                 return Noop();
             
@@ -77,7 +86,7 @@ namespace Elsa.Activities.ControlFlow
             return Done();
         }
 
-        private bool IsDone(ActivityExecutionContext context)
+        private bool IsDone(ActivityExecutionContext context, bool waitFaulted)
         {
             var recordedInboundTransitions = InboundTransitions;
             var workflowExecutionContext = context.WorkflowExecutionContext;
@@ -85,8 +94,18 @@ namespace Elsa.Activities.ControlFlow
 
             return Mode switch
             {
-                JoinMode.WaitAll => inboundConnections.All(x => recordedInboundTransitions.Contains(GetTransitionKey(x))),
-                JoinMode.WaitAny => inboundConnections.Any(x => recordedInboundTransitions.Contains(GetTransitionKey(x))),
+                JoinMode.WaitAll => inboundConnections.All(x => recordedInboundTransitions.Contains(GetTransitionKey(x)) &&
+                        (!waitFaulted || 
+                        !context.WorkflowInstance.Faults.Contains(f => f.FaultedActivityId == GetActivityId(x)) &&
+                        !context.WorkflowInstance.ScheduledActivities.Contains(f => f.ActivityId == GetActivityId(x))
+                        )
+                    ),
+                JoinMode.WaitAny => inboundConnections.Any(x => recordedInboundTransitions.Contains(GetTransitionKey(x)) &&
+                        (!waitFaulted ||
+                        !context.WorkflowInstance.Faults.Contains(f => f.FaultedActivityId == GetActivityId(x)) &&
+                        !context.WorkflowInstance.ScheduledActivities.Contains(f => f.ActivityId == GetActivityId(x))
+                        )
+                    ),
                 _ => false
             };
         }
@@ -219,6 +238,10 @@ namespace Elsa.Activities.ControlFlow
             var sourceOutcomeName = connection.Source.Outcome;
 
             return $"@{sourceActivityId}_{sourceOutcomeName}";
+        }
+        private string GetActivityId(IConnection connection)
+        {
+            return connection.Source.Activity.Id;
         }
 
         public Task Handle(WorkflowExecutionPassCompleted notification, CancellationToken cancellationToken)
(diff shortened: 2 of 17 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3182) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3182) before we can accept your contribution.<br/><br/>:x: smartwaystudio [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3182)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=3182) it.</sub>

smartwaystudio:
Willing to add more changes to this PR asap about the "Run workflow" activity because I just found that If the secondary workflow is faulted, the Run workflow activity is faulted, but if you try to retry the parent workflow, it generates a new secondary workflow instance instead of retrying the faulted one.

mohdali:
This looks very interesting and we definitely need more options for fault handling.

@sfmskywalker can you please take a look and advise as it has breaking changes.

smartwaystudio:
Hi, as I told, new changes have been pushed. The RunWorkFlow activity was always creating a new ChildWorkflow instance when retrying. There could be some scenario that you don't want that to happen, so I added a new Advanced parameter in the activity to, instead of creating new instances, retry the faulted when retrying.
image

There is a general "problem" in all activities. The problem is that they are not unique in a branched workflow. Inside a loop, the same activity instance is used for all the iterations, so is hard to determine "who am I". That was happening also inside runworkflow. If not in a branched workflow, the retry was easy, but inside a loop, you can't determine what child workflow needs to be revived, because the activity just store the last iteration state inside, so ChildWorkflowInstanceID was pointing to the last. So, what I've done is to generate a hash for the values of the Input and store it. When retrying, using the input hash I can get the real ChildWorkflowInstance id and retry it.

image
image


Here is a test without and with that parameter active. I just have a 4 iterations loop that fault. If I retry without that option, I get another 4 instances of the ChildWorkflow:
image

If the option is active, it retry the faulted (I changed the Fault activity code to fast testing, returning done):
image

sfmskywalker:
Great findings and solutions!

> There is a general "problem" in all activities. The problem is that they are not unique in a branched workflow. Inside a loop, the same activity instance is used for all the iterations, so is hard to determine "who am I"

Exactly. Your fix for this with the RunWorkflow activity is clever. The "real" fix might be for each activity execution to have its own instance ID - but that might require some significant refactorings, perhaps via https://github.com/elsa-workflows/elsa-core/issues/1059 - obviously out of scope for this PR

As an FYI: Elsa 3 already handles this today where each activity execution acquires its ow unique instance ID.

src/core/Elsa.Core/Activities/Workflows/RunWorkflow/RunWorkflow.cs
```diff
@@ -98,22 +104,94 @@ public string ChildWorkflowInstanceId
...
+                    case WorkflowStatus.Faulted:
+                        await _workflowReviver.ReviveAndRunAsync(childWorkflowInstance, cancellationToken);
+                        break;
+                    default:
+                        break;
+                }
+                childWorkflowStatus = childWorkflowInstance.WorkflowStatus;
+                ChildWorkflowInstanceId = childWorkflowInstance.Id;
+
+
```
sfmskywalker, line 169: I noticed a few excessive blank lines here and there - let's tidy it up 😄

src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-viewer/elsa-workflow-instance-journal/elsa-workflow-instance-journal.tsx
```diff
@@ -242,7 +243,7 @@ export class ElsaWorkflowInstanceJournal {
               <div class="elsa-relative elsa-flex elsa-space-x-3">
                 <div>
                   <span
-                    class="elsa-h-8 elsa-w-8 elsa-rounded-full elsa-bg-green-500 elsa-flex elsa-items-center elsa-justify-center elsa-ring-8 elsa-ring-white elsa-mr-1"
+                    class={`elsa-h-8 elsa-w-8 elsa-rounded-full elsa-bg-${eventColor}-500 elsa-flex elsa-items-center elsa-justify-center elsa-ring-8 elsa-ring-white elsa-mr-1`}
```
sfmskywalker, line 246: I am guilty of this as well, but we shouldn't dynamically construct TailwindCSS classes like this, because they will be filtered out by the CSS cleanup phase in the build action.

Instead, it's better to assign the complete class value to the `eventColor` variable. Example (pseudo code):

`const eventColor = faulted ? 'elsa-bg-red-500' : 'elsa-bg-green-500'`

This way, the complete TailwindCSS class strings will be recognized.

src/core/Elsa.Core/Activities/ControlFlow/Join/Join.cs
```diff
@@ -47,6 +47,10 @@ public enum JoinMode
         )]
         public JoinMode Mode { get; set; }
 
+
+        [ActivityInput(Hint = "True if incoming activities executed should not be Faulted to continue.", UIHint = ActivityInputUIHints.SingleLine, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })]
+        public bool WaitFaulted { get; set; } = default!;
```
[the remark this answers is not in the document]
smartwaystudio: I agree that stopping when faulted should be the correct way to do the Join Activity. Changing it to be the default behaviour and removing the property in next commit.

src/core/Elsa.Abstractions/Services/Models/WorkflowExecutionContext.cs
```diff
@@ -188,7 +188,7 @@ public void Fault(Exception? exception, string message, string? activityId, obje
             var clock = ServiceProvider.GetRequiredService<IClock>();
             WorkflowInstance.WorkflowStatus = WorkflowStatus.Faulted;
             WorkflowInstance.FaultedAt = clock.GetCurrentInstant();
-            WorkflowInstance.Fault = new WorkflowFault(SimpleException.FromException(exception), message, activityId, activityInput, resuming);
```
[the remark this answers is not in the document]
sfmskywalker: > Maybe I can think something about moving the Fault value to the new Faults property so everything will be aligned

Exactly, that would be sweet.

Thanks man, appreciate your efforts!

src/core/Elsa.Core/Activities/ControlFlow/Join/Join.cs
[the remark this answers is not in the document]
sfmskywalker: Perfect.

src/core/Elsa.Core/Activities/Workflows/RunWorkflow/RunWorkflow.cs
smartwaystudio: My bad 😉 my brain was struggling hard with this Activity. Cleaning it next commit

src/designer/elsa-workflows-studio/src/components/screens/workflow-instance-viewer/elsa-workflow-instance-journal/elsa-workflow-instance-journal.tsx
smartwaystudio: Nice, changing it.

smartwaystudio:
> Great findings and solutions!
> 
> > There is a general "problem" in all activities. The problem is that they are not unique in a branched workflow. Inside a loop, the same activity instance is used for all the iterations, so is hard to determine "who am I"
> 
> Exactly. Your fix for this with the RunWorkflow activity is clever. The "real" fix might be for each activity execution to have its own instance ID - but that might require some significant refactorings, perhaps via #1059 - obviously out of scope for this PR
> 
> As an FYI: Elsa 3 already handles this today where each activity execution acquires its ow unique instance ID.

When I started to design some testing workflows to learn about Elsa I was a bit surprised when the execution log in a loop was showing the same ActivityId for one looped activity and also just storing the last executed activity value. Moreover, that means that the parallelism is limited. I thought that, using hangfire, it would create a hangfire job for each branch, at least in the parallel foreach. That issue you are talking about would be very nice but indeed, requires a lot of refactoring. Can you point me how have you done this in Elsa 3? Maybe I can think about it! 😄

src/core/Elsa.Core/Activities/Workflows/RunWorkflow/RunWorkflow.cs
sfmskywalker: I can relate 😄

sfmskywalker (approved):
Awesome job 👍🏻

mohdali:
This is really exciting and I want to test it soon. @smartwaystudio is it possible to share the json of the test cases you created?

mohdali:
> > This is really exciting and I want to test it soon. @smartwaystudio is it possible to share the json of the test cases you created?
> 
> What I've done to test was creating some workflow (Fork, foreach, pforeach, while, fork-join, runWorkflow, foreach-runworkflow) and temporary modifying the Fault activity. The first run, the workflow fails and then, I change the Fault activity to return Done() (also adding a outcome to connect an activity to it) to test the retry. I've done this way because there is no way to edit the current version of workflow with dashboard.
> 
> Do you want a export of those workflow with the modified fault?

Yes please share the export. I'm not sure how you made the modification but I think if you unpublish then republish the workflow it will modify the current version.

smartwaystudio:
> > > This is really exciting and I want to test it soon. @smartwaystudio is it possible to share the json of the test cases you created?
> > 
> > 
> > What I've done to test was creating some workflow (Fork, foreach, pforeach, while, fork-join, runWorkflow, foreach-runworkflow) and temporary modifying the Fault activity. The first run, the workflow fails and then, I change the Fault activity to return Done() (also adding a outcome to connect an activity to it) to test the retry. I've done this way because there is no way to edit the current version of workflow with dashboard.
> > Do you want a export of those workflow with the modified fault?
> 
> Yes please share the export. I'm not sure how you made the modification but I think if you unpublish then republish the workflow it will modify the current version.

[test_workflows.zip](https://github.com/elsa-workflows/elsa-core/files/9095199/test_workflows.zip)

Here you have, some json workflows and `Fault.cs`. What I do is to force force Fault activity to sometimes `return Done()`. First execution of the workflow I return fault or exception and then, I change the code to `return Done()` to verify that the workflow ends properly. Is like emulating a scenario when an activity suddenly fails and you just need to retry it (ex: because it was a network error)

mohdali:
> 

@smartwaystudio thanks for the examples!

I have tested with the scenario of having dependent activities in the foreach and they executed once the workflow is retired and the error is cleared. This is really awesome!

image

Issue #3187, opened 2022-07-07
HttpEndpoint Route Params are stored in lowercase


jdevillard:
Following discussion on discord, 

After some test, we can see that Route Params available in the RouteValues dictionary are stored in lowercase. RouteValues Keys AND Values are in lowercase
![image](https://user-images.githubusercontent.com/3254429/177851784-f79c9699-31d9-49d8-a2ee-c7449cadb909.png)

Though it seems to not be an issue for Keys because the dictionary used is CaseInsensitive on Keys. This is an issue for Values because we don't have to alterate the data. (Customer have issue because it used Base64 data for example).

The issue seems to be on the HttpEndpointMiddleware where the path is changed using ToLowerInvariant)

PR #3189, opened 2022-07-08
don't lowercase the Path Parameters in the URI (#3187)

jdevillard:
This PR resolve issue #3187

As what I read, only Scheme and Host part of the URI should be compared case-insensitive and can be normalized to lowercase.

This PR fix Elsa v2 , another have to be done for v3.


sfmskywalker:
What happens when the user uses e.g. `/hello-world` as the path, but then the site visitor navigates to e.g. `/Hello-World`? Will that still match after this change?

jdevillard:
I think it's ok 

image

it's not a unit test but for me the URI Matcher is not case sensitive.

sfmskywalker (approved):

PR #3216, opened 2022-07-19
#2857 attempt to reapply fix for dropdown

mohdali:
@sfmskywalker trying to reapply the fix for dropdown as in #2857.

Simply I'm trying to reassign the value at the end of `componentWillLoad()` to force the default value.
```js
if (this.currentValue != undefined) {
      this.propertyModel.expressions[defaultSyntax] = this.currentValue;
    }
```
This needs to be validated with the cascading dropdown feature from @jdevillard


jdevillard:
The fix seems to work correctly on the cascading dropdown 👍, however, it seems to have an other issue on the first load.

image

The value of the third dropdown is undefined : 
image

because the call to the api is done with only one dependand value instead of the two : 
image

color depends on Brand and Model and only Model is set at the init. 

If we change one dropdown, all the events/call are set correctly. It's only at init.
image

sfmskywalker:
@mohdali do you want to merge this or look at @jdevillard 's feedback first?

mohdali:
> @mohdali do you want to merge this or look at @jdevillard 's feedback first?

Will merge, he will create a separate PR

PR #3245, opened 2022-08-08
Cleaning up and adding simple support for XML response in HTTP Send Request Activity

### LINKED ISSUE #3244: Cleanup of SendHTTPRequest activity
Per discussion in #3229 .....

**Response Content Parser's** of the SendHTTPRequest activity should be cleaned up and simplified because some of them do 
same/similar processing.

1. Remove **JSON Raw** parser (it has identical implementation to **Plain text**)
2. Enhance **.NET Object** so it correctly handles arrays (as **Expando Object** does) when it falls back to `ExpandoObject` when no .NET target is given
3. Remove **Expando Object** (with change in 2., **.Net Object** when no .NET target is given implements same functionality)

In addition it would be nice if **.NET Object** and **JToken** had some basic support for XML.

tomy2105:
This resolves #3244.

**JSON Raw** and **Expando Object** parsers have been removed.
**.NET Object** parses has been improved to correctly handles arrays when it falls back to `ExpandoObject` when no .NET target is given.
Both **.NET Object** and **JToken** parsers now handle XML (by converting it first to JSON with help of [Newtonsoft](https://www.newtonsoft.com/json/help/html/ConvertXmlToJson.htm)).

The only problem I have is backward compatibility (my usual problem) with workflows that used to use **JSON Raw** and **Expando Object** parsers. I currently don't have an idea how to solve this best so @sfmskywalker or anyone else, feel free to suggest.

Here is a [simple designer workflow HttpSendTest.txt ](https://github.com/elsa-workflows/elsa-core/files/9281888/HttpSendTest.txt) to test **.NET Object** and **JToken** parsers with JSON, JSON Array and XML response.


src/activities/Elsa.Activities.Http/Parsers/Response/TypedHttpResponseContentReader.cs
```diff
@@ -12,24 +13,24 @@ public class TypedHttpResponseContentReader : IHttpResponseContentReader
...
-            }
-
-            if (GetIsXmlContentType(contentType))
-            {
-                // TODO: parse XML.
-                throw new NotImplementedException();
+                var responseText = (await response.Content.ReadAsStringAsync()).Trim();
+                if (GetIsXmlContentType(contentType))
+                {
+                    XmlDocument doc = new XmlDocument();
```
sfmskywalker, line 28: Please use `var`.

src/activities/Elsa.Activities.Http/Parsers/Response/JTokenHttpResponseContentReader.cs
```diff
@@ -1,22 +1,32 @@
...
         public async Task<object> ReadAsync(HttpResponseMessage response, object context, CancellationToken cancellationToken)
         {
-            var json = (await response.Content.ReadAsStringAsync()).Trim();
-            return GetJTokenValue(json);
+            var contentType = response.Content.Headers.ContentType.MediaType;
+
+            var responseText = (await response.Content.ReadAsStringAsync()).Trim();
+            if (GetIsXmlContentType(contentType))
+            {
+                XmlDocument doc = new XmlDocument();
```
sfmskywalker, line 25: Please use `var`.

sfmskywalker (approved):
Awesome!

tomy2105:
Usage of `var` resolved and rebased to resolve conflicts.

In addition added runtime mapping of old names of content parsers to new names since this is the only way I could think of at the moment to keep old workflows running without `The specified parser does not exist` exception.

This doesn't solve workflow designer issue displaying `auto select` if old parser name is used. It seems I would need to go deeper into workflow provider to fix that, and this doesn't seem to be that big deal to venture there :).

sfmskywalker (approved):
Awesome. Kudos for maintaining backwards compatibility 🤗

mohdali:
Adding XML support is really awesome. We had to create a custom activity for that to handle SOAP and similar cases.

The way Newtonsoft converts XML might no tbe intiuitive though. I remember running into issues with lists of elements and element attributes. I will try to check again for the specific issues and how we handled them.

Issue #3146, opened 2022-06-22
After implementing a CustomTenantAccessor Signals do not work


ricksearcy:
Related to this issue https://github.com/elsa-workflows/elsa-core/issues/2861 signals do not work after enabling the tenantId implementation.

jruckert:
@sfmskywalker - I had to do some investigation around this last night actually (had to implement my own workaround for the Azure Service Bus trigger) to make it work in a multi-tenant environment.

To start the conversation/example:

https://github.com/elsa-workflows/elsa-core/blob/bab82b545023f638110037feac86259d169bf63a/src/activities/Elsa.Activities.AzureServiceBus/Services/Worker.cs#L119

The query here by default is passing in a TenantId = null

**Question**

"Is the intention behind multi-tenancy when a WorkflowsQuery TenantId = null is set to pickup all of the triggers/bookmarks that either **have a TenantId set and Null entries**? or just specifically null entries?"

Depending on the answer to this question would determine where the fix would be, as modifying the underlying FindManyAsync call may impact multiple areas to basically ignore the TenantId = null from the underlying EF query etc as its only returning any entry with a NULL tenant id by default.

**Workaround for Azure Service Bus**

My current workaround is to create a new "Tenants" HashSet (same as how you use Tags) to the Worker.cs file (as each Trigger/Bookmark has a string? TenantId property already it)

```
public HashSet<string?> Tenants { get; } = new();
```

Tags are set in this function here, so I have introduced a new parameter "string? tenantId" to this function.

https://github.com/elsa-workflows/elsa-core/blob/bab82b545023f638110037feac86259d169bf63a/src/activities/Elsa.Activities.AzureServiceBus/Services/WorkerManager.cs#L98

```
worker.Tenants.Add(tenantId);
```
then modifying the code in the Worker.cs file to loop through the tenants as per

```
foreach (var tenant in Tenants)
            {
                var launchContext = new WorkflowsQuery(ActivityType, bookmark, correlationId, TenantId: tenant);
                using var scope = _serviceScopeFactory.CreateScope();
                var workflowLaunchpad = scope.ServiceProvider.GetRequiredService<IWorkflowLaunchpad>();
                await workflowLaunchpad.CollectAndDispatchWorkflowsAsync(launchContext, new WorkflowInput(model), cancellationToken);
            }
```
Not sure if this is a good approach or not, but does work for me as an interim fix.

[...]

PR #3250, opened 2022-08-11
Added Tenant Signaler Implementation and Options

jruckert:
Added Tenant Signaler as an option


sfmskywalker (approved):

tomy2105:
I believe this fixes #3146.

Issue #3001, opened 2022-05-07
ELSA foreach iterate for infinite loop


minal87jadhav:
I have just created a simple flow where json data is coming from HTTpPoint where foreach needs to iterate for orders. instead of running for 3 orders, its keep on running.

![image001](https://user-images.githubusercontent.com/43084016/167242283-f6020d3b-864d-4d1d-bf77-d7ec93cdafc7.png)

Order data is
Order:[
{
"RefId":"aa",
"Name":"bb"
},
{
"RefId":"aa1",
"Name":"bb1"
},
{
"RefId":"aa2",
"Name":"bb2"
}
]

please let me know anyone has solution for it

dgumbo:
I also have the same problem, but for me, the loop is causing my system to crash,

it was an error on my part thogh, because i set the loop variable wrong [**activities.ActivityName**], instead of [**activities.ActivityName.Output()**], what i would like is for the activity to error out rather than engage an infinite loop.

sfmskywalker:
@minal87jadhav 

Can you please share an export of your workflow definition so we can take a look?
As @dgumbo mentioned, it could be an issue with an expression. Of course, this should then fault the workflow rather than causing an infinite loop. But it would be good to see exactly what is happening first so we might try and provide a proper fix and instructions as appropriate.

PR #3252, opened 2022-08-11
Prevent rescheduling of looping/scoped activities if faulted

### LINKED ISSUE #3020: Workflow loops forever because of error in RunJavaScript
[error-loop.zip](https://github.com/elsa-workflows/elsa-core/files/8680534/error-loop.zip)

Following workflow loops forever because of the RunJavascript (that has error) inside While loop.
**It loops forever** instead of faulting. Workflow is attached.

It seems to be connected to RunJavascript because if I replace it with SetVariable with identical error, the workflow instance faults as expected.

tomy2105:
I believe this fixes #3020, maybe even #3001.
I've tested `For`, `For Each`, `While`.... loop inside loop, etc.....

What I don't know how it "interacts" with compensation activities......

### DIFF
--- a/src/core/Elsa.Core/Handlers/RescheduleBranchingActivitiesAndContainers.cs
+++ b/src/core/Elsa.Core/Handlers/RescheduleBranchingActivitiesAndContainers.cs
@@ -1,4 +1,4 @@
-﻿using System.Linq;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


sfmskywalker (approved):

sfmskywalker:
> What I don't know how it "interacts" with compensation activities......
Will test when I get a chance. For now, this is a great improvement.

Issue #3253, opened 2022-08-12
Retention Module Job could cause timeout when deleting large data (or small db)


jdevillard:
Issue detected using SQL Azure Standard S1: 20 DTUs.

I've dig into the retention module and see that no data was really deleted from the Db.
After going step by step , I've seen different things : 

https://github.com/elsa-workflows/elsa-core/blob/e4bab8810afc201cd7b474fccb1e0757e448d528/src/modules/retention/Elsa.Retention/Jobs/CleanupJob.cs#L52-L69

Record are paged for retrieval but they will be deleted in batch. This can cause timeout when deleting with silent error, and a retry only after the SweepInterval

![image](https://user-images.githubusercontent.com/3254429/184365265-11af4a24-db40-4790-82ea-195c0e130428.png)

Following the code : 
```
await _workflowInstanceStore.DeleteManyAsync(specification, cancellationToken);
```
goes through (EventPublishingWorkflowInstanceStore.cs)
```csharp
        public async Task<int> DeleteManyAsync(ISpecification<WorkflowInstance> specification, CancellationToken cancellationToken = default)
        {
            var instances = await FindManyAsync(specification, cancellationToken: cancellationToken).ToList();
            var count = await _store.DeleteManyAsync(specification, cancellationToken);

            if (instances.Any())
            {
                foreach (var instance in instances)
                    await _mediator.Publish(new WorkflowInstanceDeleted(instance), cancellationToken);

                await _mediator.Publish(new ManyWorkflowInstancesDeleted(instances), cancellationToken);
            }

            return count;
        }
```

Then in EntityFrameworkWorkflowInstanceStore.cs, a new time FindManyAsync
```csharp
        public override async Task<int> DeleteManyAsync(ISpecification<WorkflowInstance> specification, CancellationToken cancellationToken = default)
        {
            var workflowInstances = (await FindManyAsync(specification, cancellationToken: cancellationToken)).ToList();
            var workflowInstanceIds = workflowInstances.Select(x => x.Id).ToArray();
            await DeleteManyByIdsAsync(workflowInstanceIds, cancellationToken);
            return workflowInstances.Count;
        }
```

to finaly DeleteManyByIdsAsync with dependencies.

I think we loose the effort of using the pagination and I think we should delete records for each page instead of using page only for __first__  retrieval

jdevillard:
We can also change the way we select data, from older to newer.

[...]

PR #3278, opened 2022-08-31
Retention module improvement #3253

jdevillard:
This PR aims to enhance the Retention Module by just query the Store with the necessary data.

Indeed, in the current store implementation, when a delete is realize, some select are done which get All the data of the DB. This could be very huge (when considering Data column) even for one record.

The idea is to :
- query the Store and getting only Id and then delete the record.
- filter data as much as we can on server side (using specification) 
- lower the memory consumption of this module.

to avoid Breaking Change, some elements are re-defined in the module. Maybe some method could be up to the Core Module.


sfmskywalker:
@jdevillard what should we do with this PR?

jdevillard:
This one can be cancelled (replace by the other PR you've accepted just before !)

PR #3326, opened 2022-10-08
reuse createHttpClient

huhaixiao:
createHttpClient is duplicated


sfmskywalker (approved):
Thanks for the nice cleanup 😄

PR #3355, opened 2022-10-14
Feature/send http request

cristinamudura:
Implemented Send Http Request activity 
Implemented checkbox input


tomy2105:
@cristinamudura is this really intentional? In src/modules/Elsa.Http/Constants/MimeTypes.cs:
public const string ApplicationWwwFormUrlEncoded = "**VALUE**";

sfmskywalker (approved):
Beautiful, thank you!

PR #3309, opened 2022-09-26
fix error when TargetType not configured #3308

jxnkwlp:
#3308 
fix error when HttpEndpoint.TargetType not configured

Issue #3308, opened 2022-09-26
Recommended change to FirstOrDefault 


jxnkwlp:
https://github.com/elsa-workflows/elsa-core/blob/0611487036408ce2793d405b43a40da2be3cc996/src/activities/Elsa.Activities.Http/Scripting/JavaScript/HttpTypeDefinitionProvider.cs#L22

If HttpEndpoint.TargetType not set or the property is not set in database, these well be throw error . 

Strictly speaking, i think use `FirstOrDefault ` will be better

PR #3358, opened 2022-10-16
fix error when TargetType not configured #3308 #3309

jxnkwlp:
#3309 changes , ` HttpEndpointTypeDefinitionRenderer.cs ` also to modify.


sfmskywalker (approved):
Thanks 😄

Issue #2723, opened 2022-02-02
Display Faulted Activity in Journal


sfmskywalker:
Hello @sfmskywalker,

To make custom activity to **Failed ** state in case of any business logic exception, I have implemented the following options in the custom activity  exception block:

**Option1:** **throw(ex)** statement 
```
try
{
 //Business logic
}
catch (Exception ex)
 {
  _outcome = "Error Occured";
  context.JournalData.Add("Exception", ex.Message);
  throw (ex);
 }
```

If I use the throw statement, I have observed that in the workflow execution instance with fault exception along with fault message and activity state also as "**Faulted**" and also Workflow status as "**Faulted**"
![image](https://user-images.githubusercontent.com/67507118/150673517-5662dde5-7420-44e4-9fba-afbc7a8ffad3.png)

![image](https://user-images.githubusercontent.com/67507118/150673659-4993b8b5-5e79-4b64-91c5-746ae0dab2bc.png)

**Option II: Return fault(ex)**
```
try
{
 //Business logic
return Done();
}
catch (Exception ex)
 {
  _outcome = "Error Occured";
  context.JournalData.Add("Exception", ex.Message);
return Fault(ex)
 }
```

In this case I have observed in the Workflow execution instance as follow: Activity state is "**Executed**" and Workflow status as "**Faulted**"
![image](https://user-images.githubusercontent.com/67507118/150674104-7d4aef61-c2b2-43a9-a5d6-011416d2b881.png)"
![image](https://user-images.githubusercontent.com/67507118/150674118-766c4c56-da9e-46c4-b320-98fb76302c72.png)

**Option -III:**  Implemented **WorkflowFault** method. 
```
try
{
 //Business logic
}
catch (Exception ex)
 {
  _outcome = "Error Occured";
  context.JournalData.Add("Exception", ex.Message);
context.WorkflowInstance.Fault = new Elsa.Models.WorkflowFault(null, "Exception occured at activity level", context.ActivityId.ToString(), context.Input, context.Resuming);
 }
```

In this case Workflow status is "**Finished**" and Activity Status is "**Executed**" 
![image](https://user-images.githubusercontent.com/67507118/150674466-a133eb68-b747-4535-a3d3-d6f424550af7.png)

![image](https://user-images.githubusercontent.com/67507118/150674508-39e1d73e-8a4f-400b-8e23-f6039790ad8b.png)

[...]

PR #3377, opened 2022-10-21
Check the result and throw if it's a fault

stuartmcgillivray:
Updating the workflow runner to throw if a FaultResult is returned by the Activity.

As highlighted in #2723 if you return a FaultResult from an activity, the execution log doesn't get the fault added to it so it won't be displayed in the designer.

To correct this i've checked to see if the result of the activity is a faultResult and throw an exception so sends a ActivityFaulted event.

We could also expand the ActivityFaulted to accept a message as well as the exception so it matches FaultResult and in the handler resolve whether it's to use the Message or the Exception rather than having to create/thow an exception.

### DIFF
--- a/src/core/Elsa.Core/Services/Workflows/WorkflowRunner.cs
+++ b/src/core/Elsa.Core/Services/Workflows/WorkflowRunner.cs
@@ -325,14 +325,14 @@ namespace Elsa.Services.Workflows
 
                     workflowExecutionContext.CompletePass();
                     workflowInstance.LastExecutedActivityId = currentActivityId;
-                    
+
                     await CheckIfCompositeEventAsync(isComposite
                         , compositeScheduledValue
                         , new ActivityExecutionResultExecuted(result, activityExecutionContext)
                         , _mediator
                         , cancellationToken);
 
-                    if (workflowInstance.WorkflowStatus == WorkflowStatus.Faulted) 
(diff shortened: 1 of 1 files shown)


src/core/Elsa.Core/Services/Workflows/WorkflowRunner.cs
```diff
@@ -381,12 +381,24 @@ private async Task CheckIfCompositeEventAsync(bool isComposite, bool scheduleVal
...
+
+                if (result is FaultResult)
+                {
+                    var faultResult = (result as FaultResult);
+
+                    if (faultResult?.Exception != null)
+                        throw faultResult.Exception;
+
+                    throw new Exception(faultResult?.Message);
+                }
```
sfmskywalker, line 394: Maybe simplify this a little bit? E.g.

```csharp
if (result is FaultResult faultResult)
{
    if (faultResult.Exception != null)
        throw faultResult.Exception;

    throw new Exception(faultResult?.Message);
}
````

sfmskywalker (approved):

src/core/Elsa.Core/Services/Workflows/WorkflowRunner.cs
stuartmcgillivray: I've made this change, didn't know is could be used like this :D

PR #3409, opened 2022-11-08
Use log scopes with a better `ToString()` implementation.

Chryssie:
For loggers such as the console logger just write the result of `ToString()` method (see [here](https://github.com/dotnet/runtime/blob/0589ce6c3022f4b9b7571268be5f97f45a003ad7/src/libraries/Microsoft.Extensions.Logging.Console/src/SimpleConsoleFormatter.cs#L207)).

Which results in it writing something like:
```
=> SpanId:f126fcaa1f140d25, TraceId:0bd70f4174cd963684d8fd55e4dc2ecb, ParentId:0000000000000000 => ConnectionId:0HMM16A1C5J1U => RequestPath:/test-path RequestId:0HMM16A1C5J1U:00000009 => System.Collections.Generic.Dictionary`2[System.String,System.Object]
```

Of note is the `System.Collections.Generic.Dictionary`2[System.String,System.Object]` part.

This just replaces those dictionaries with some concrete types (based on [aspnetcore/ConnectionLogScope.cs](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/SignalR/common/Http.Connections/src/Internal/ConnectionLogScope.cs)) to override the `ToString()` method and allow for better log messages.

### DIFF
--- a/src/activities/Elsa.Activities.Telnyx/TelnyxWebhookLogScope.cs
+++ b/src/activities/Elsa.Activities.Telnyx/TelnyxWebhookLogScope.cs
@@ -1,0 +1,38 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Elsa.Activities.Telnyx
+{
+    internal class TelnyxWebhookLogScope : IReadOnlyList<KeyValuePair<string, object?>>
+    {
+        private string? _cachedToString;
+
+        public string? CorrelationId { get; }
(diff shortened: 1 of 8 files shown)


sfmskywalker:
I didn't know about this. Thanks!

sfmskywalker (approved):

PR #3423, opened 2022-11-10
Publish symbol packages to the v3 NuGet API.

Chryssie:
The symbol packages don't seem to be available from nuget.org.

I can't verify that this fixes the issue as it requires actually pushing the packages but based on [How to publish NuGet symbol packages using the new symbol package format '.snupkg' | Microsoft Learn](https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#publishing-a-symbol-package) it looks like this _might_ resolve the issue.


sfmskywalker (approved):

sfmskywalker:
Worth a try 👍🏻

PR #3478, opened 2022-11-26
fix EnumerableTypeDefinitionProvider #3477

### LINKED ISSUE #3477: Typescript definition not valid for list/array of complex types
We have a custom TypeDefinitionProvider that collects our C# types in order to provide intellisense in Javascript syntax.
It works well, but when the type has a list or an array (or other collections) of customs types, the intellisense return a not valid typescript definition.

To replicate the problem, here is a very basic TypeDefinitionProvider:

```csharp
    public class CustomTypeProvider : TypeDefinitionProvider
    {
        public override IEnumerable<Type> CollectTypes(TypeDefinitionContext context)
                => new[] { typeof(CustomType), typeof(ComplexType) };
    }
```
where the types are:

```csharp
    public class CustomType
    {
        public string Primitive { get; set; } = string.Empty;
        public List<ComplexType> Complex { get; set; } = new List<ComplexType>();
    }

    public class ComplexType
    {
        public string Name { get; set; } = string.Empty;
        public int Number { get; set; } = 0;
    }
```

In this scenario, the intellisens is not valid
image

williamverdolini:
This should fix #3477 
I'm quite new with xUnit, if the test could be better, please provide me some guidance.

Imho I found the design of the TypeDefinitionProvider a bit confusing: if I've understood it correctly, I would have preferred to split the responsibility of collecting types (`CollectTypes`) from that of returning the Typescript definition types (`SupportsType/GetTypeDefinition`); because they are not even cohesive responsibilities.
In fact, up to date, I don't see any providers that do both.

### DIFF
--- a/src/scripting/Elsa.Scripting.JavaScript/Typings/EnumerableTypeDefinitionProvider.cs
+++ b/src/scripting/Elsa.Scripting.JavaScript/Typings/EnumerableTypeDefinitionProvider.cs
@@ -1,8 +1,8 @@
-﻿using System;
-using System.Collections;
-using System.Linq;
 using Elsa.Scripting.JavaScript.Services;
 using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections;
+using System.Linq;
 
 namespace Elsa.Scripting.JavaScript.Typings
 {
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3478) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3478) before we can accept your contribution.<br/><br/>:x: williamverdolini [sign now](https://cla.dotnetfoundation.org/elsa-workflows/elsa-core?pullRequest=3478)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/elsa-workflows/elsa-core?pullRequest=3478) it.</sub>

sfmskywalker:
@williamverdolini 100% agree with your feedback on the TypeDefinitionProvider. I'll try a better approach for V3.

sfmskywalker (approved):

PR #3487, opened 2022-11-29
Add MySql secrets provider

johnwc:


sfmskywalker (changes requested):
All looks good! There is only a small typo in the project names  ("Persistance" should be "Persistence"). Once that's changed, I think it's good to go :shipit:

sfmskywalker (approved):

PR #3473, opened 2022-11-25
Allow sending input as http request body and output the response content

tanelkuhi:
Our client has this specific requirement to fetch a file from an url, and then make a request to another url with that file attached as content. These changes aim to solve this.

PR #3489, opened 2022-11-30
Allow handling SendHttpRequest content as bytes

tanelkuhi:
This is a continuation of the topic started with the previous PR, which I closed:
https://github.com/elsa-workflows/elsa-core/pull/3473
The picture below is the gist of it:
image

It turns out, that by using javascript in the content field, we can take the output of the previous SendHttpRequest, and put it into the body. However, since the Content field currently is a string, the file data will get corrupt in the middle of the conversion, even if we use something like `String.fromCharCode.apply(null, activities.PictureGet.ResponseContent())` to turn the bytes into string.

But this is issue is easily solved, by turning Content into object, and then check by type whether to send as string or as byte array.

### DIFF
--- a/src/activities/Elsa.Activities.Http/Activities/SendHttpRequest/SendHttpRequest.cs
+++ b/src/activities/Elsa.Activities.Http/Activities/SendHttpRequest/SendHttpRequest.cs
@@ -64,7 +64,7 @@ namespace Elsa.Activities.Http
         /// The body to send along with the request.
         /// </summary>
         [ActivityInput(Hint = "The HTTP content to send along with the request.", UIHint = ActivityInputUIHints.MultiLine, SupportedSyntaxes = new[] { SyntaxNames.JavaScript, SyntaxNames.Liquid })]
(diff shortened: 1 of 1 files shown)


johnwc:
Are you unable to send as base64 encoded?

tanelkuhi:
@johnwc I am afraid the endpoint we need to send the file to, expects the file as binary.

sfmskywalker (approved):

sfmskywalker:
It makes sense to accept an object as content instead of string, because it allows user to provide things like a byte array, a stream or even a URL that the activity internally knows how to handle. The SendEmail works in the same way, where users can provide an object (or array of objects) for attachments.

sfmskywalker:
> Are you unable to send as base64 encoded?

Should be possible indeed. In addition, allowing other types like byte arrays and streams makes it that much more easy to use from a user's perspective. For example, if you downloaded a file and have a byte array on hand, you can pass it in directly instead of having to encode it using base64.

PR #3505, opened 2022-12-05
Make improvements related to x6 graphs

tanelkuhi:
1. Improve graph connection generation by calculating `step` value to the edge routing depending on the layout of the target nodes
2. Update connection views, when moving one node, so nodes won't hide the connections underneath them
3. Allow pasting of nodes to cursor position (useful when pasting to another workflow)
4. Fix showing the node selection style in all environments (updated the stencil configs - purge css safelist)
5. Make auto-layout use batch update, so the whole operation can be undone with ctrl+z
6. Improve the per-activity test result window positioning, so it wouldn't appear outside the screen and also make it scrollable


tanelkuhi:
More improvements incoming soon. No need to merge yet.

sfmskywalker (approved):

PR #3510, opened 2022-12-07
Enable opening expression editor in a modal dialog

tanelkuhi:
For switch activity, the cases are single line expression editors and that makes it really hard to write any meaningful code there. But making the editor multi-line and increasing its height would bloat the contents of the modal too much, so it was our client's idea to open a dialog with a bigger editor inside to make it comfortably editable and not bloat anything.

### DIFF
--- a/src/designer/elsa-workflows-studio/src/components.d.ts
+++ b/src/designer/elsa-workflows-studio/src/components.d.ts
@@ -109,6 +109,7 @@ export namespace Components {
         "editorHeight": string;
         "expression": string;
         "language": string;
+        "opensModal": boolean;
         "padding": string;
         "serverUrl": string;
         "setExpression": (value: string) => Promise<void>;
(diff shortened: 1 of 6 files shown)


sfmskywalker:
Being able to open the editor in a separate modal is great! And I have some suggestions to make it a little bit better than the way it behaves right now.

Right now, the modal opens as soon as the single-line Monaco editor receives focus, which is too disruptive because you should still be able to write short, single line expressions and not get the modal open up all the time.

Instead, it might be better to have a small icon be displayed somewhere that the user can click in order to open the editor in a separate modal.

Maybe something like this:

image


As a bonus, this could work generically for all single line monaco editors, not just cases for Switch.

What do you think?

tanelkuhi:
@sfmskywalker Seems like a good idea to me. Also our client wants the changes you write inside the dialog to not take effect immediately, but that you should have to click save - so then you could undo the changes you just made.

sfmskywalker (approved):

Issue #3485, opened 2022-11-29
Credential Manger UI hardcoded path


johnwc:
Credential manager frontend UI is hardcoding the path for `initializeMonacoWorker(...)`, this is causing a javascript error in the browser.

https://github.com/elsa-workflows/elsa-core/blob/master/src/designer/elsa-workflows-studio/src/modules/credential-manager/elsa-secret-editor-modal/elsa-secret-editor-modal.tsx#L36

johnwc:
This still exists.

johnwc:
@sfmskywalker I think this is resolved via latest release.

sfmskywalker:
Great, thanks @johnwc 👍🏻

Issue #3480, opened 2022-11-26
Elsa.Secrets abstraction


johnwc:
Adding `Elsa.Secrets` to our WF server project, adds what seems to be unnecessary packages. Can whatever that is requiring `Azure.Identity`, `Microsoft.Identity.Client.Extensions.Msal`, `Elsa.Activities.Sql`, and `Elsa.Scripting.Sql` be abstracted out of the core `Elsa.Secrets` library. Then whoever needs to use it can pull it into their project individually?

```
Updates:

Microsoft.Data.SqlClient.2.1.4 -> Microsoft.Data.SqlClient.3.0.0
Microsoft.Data.SqlClient.SNI.runtime.2.1.1 -> Microsoft.Data.SqlClient.SNI.runtime.3.0.0
Microsoft.Identity.Client.4.21.1 -> Microsoft.Identity.Client.4.22.0
Npgsql.6.0.1 -> Npgsql.6.0.3

Installing:

Azure.Identity.1.3.0
Elsa.Activities.Sql.2.10.0-preview.564
Elsa.Scripting.Sql.2.10.0-preview.564
Elsa.Secrets.2.10.0-preview.564
Microsoft.Identity.Client.Extensions.Msal.2.16.5
```

sfmskywalker:
Indeed, the Secrets package shouldn't have to depend on `Elsa.Activities.Http` and `Elsa.Activities.Sql`.
There should be a new package like e.g. `Elsa.Secrets.Bundle` or whatever that bundles various providers for various other modules (like HTTP and SQL).

mohdali:
I think it is also better to retarget it to `netstandard2.1` once abstraction is done. One less project for multi-targeting :)

PR #3520, opened 2022-12-09
Elsa Secrets imporvments

johnwc:
Abstract Sql and Http activities out of main Elsa.Secreets #3480
Fix hardcoded peth #3485
Other enhancements for access from activities


johnwc:
@sfmskywalker while working on these updates, I stumble upon a potential issue. A selected secret's value is stored within the workflow definition. Rather than just the name of the secret. I was under the impression that the credential manager was going to be a store, where an activity uses a secret by ref, but the actual secret value is only stored in one place. The value being looked up and applied to the input upon execution. I think we would need something similar to IActivityPropertyValueProvider, similar to how script expressions are handled. What are your thoughts?

johnwc:
> > @sfmskywalker while working on these updates, I stumble upon a potential issue. A selected secret's value is stored within the workflow definition. Rather than just the name of the secret. I was under the impression that the credential manager was going to be a store, where an activity uses a secret by ref, but the actual secret value is only stored in one place. The value being looked up and applied to the input upon execution. I think we would need something similar to IActivityPropertyValueProvider, similar to how script expressions are handled. What are your thoughts?
> 
> Hi. I have addressed this issue inside another pending PR #3275. It will evaluate secrets for each activity when executed. Secrets will be referenced by IDs hence no need to republish all workflows on every secret change.

Referenced by IDs or Names? Names would be best, especially when you are sharing workflow designs between two different environments.

johnwc:
@sfmskywalker can we get this approved?

sfmskywalker:
@johnwc I'll try and set aside some time to review this as soon as possible. Thank you for your contributions!

sfmskywalker (approved):
Lovely!

sfmskywalker:
@ke-real Before merging this, perhaps you'd like a chance first to try out some of these changes to make sure it didn't accidentally break anything or take advantage of some of the improvements introduced with this PR?

sfmskywalker:
@johnwc With the merge of @ke-real 's PR, your PR received some conflicts - sorry about that!

johnwc:
@ke-real do you think you can create a sample working project for credential manager in the solution, similar to the other samples here? If you can do that, I can write up documentation on how to implement and use secrets/credential manager.

image

ke-real:
> @ke-real do you think you can create a sample working project for credential manager in the solution, similar to the other samples here? If you can do that, I can write up documentation on how to implement and use secrets/credential manager.
> 
> image

might be a good idea. There is in fact quite some room for improvement in this module. (e.g. moving the secret definitions to the backend like it is with Activities)

src/modules/secrets/Elsa.Secrets/Providers/SecretsProvider.cs
```diff
@@ -40,16 +46,50 @@ public SecretsProvider(ISecretsManager secretsManager, IEnumerable<ISecretValueF
...
         {
             var secrets = await _secretsManager.GetSecrets(type);
-            return secrets.Select(x => (x.Name ?? x.DisplayName, x.Name)).ToArray();
+
+            var formatter = _valueFormatters.FirstOrDefault(x => x.Type == type);
+
+            try
+            {
+                if (formatter != null)
+                    return secrets.ToDictionary(x => x.Name ?? x.DisplayName ?? x.Id, x => formatter.FormatSecretValue(x).Result);
```
ke-real, line 83: This actually breaks the referencing by name. It was intentionally changed compared to the original implementation. Formatting every secret for the options select list will just copy secret value during design time and will not be managable in future. In addition this complete breaks Oauth2 secret which only works when evaluated during runtime.

johnwc: This can be changed to the way you changed it to, value being the name of the secret. But, I would be sure to use the new format `<secret type>:<secret name>`. So that when the expression is evaluated, it can determine exactly what secret to use when there may be two secrets with the same name, but different type.

johnwc: @ke-real did you happen to update this, or should I do it real quick for 2.10 release?

ke-real: @johnwc it should be fixed in #3575 but we can also make a smaller PR if needed.

PR #3530, opened 2022-12-13
Added correlation activity

cristinamudura:


src/modules/Elsa.Workflows.Core/Activities/Correlate.cs
```diff
@@ -0,0 +1,27 @@
+using System.ComponentModel;
+using System.Text.Json.Serialization;
+using Elsa.Workflows.Core.Attributes;
+using Elsa.Workflows.Core.Models;
+
+namespace Elsa.Workflows.Core.Activities;
+
+[Activity("Elsa", "Control Flow", "Set the CorrelationId of the workflow to a given value.")]
```
sfmskywalker, line 8: Maybe change the category from "Control Flow" to "Primitives", what do you think?

src/modules/Elsa.Workflows.Core/Activities/Correlate.cs
```diff
@@ -0,0 +1,27 @@
...
+public class Correlate : Activity
+{
+    [JsonConstructor]
+    public Correlate()
+    {
+    }
+    
+    /// <summary>
+    /// The text to write.
```
sfmskywalker, line 17: Should probably read something like "The correlation ID to set." ;)

sfmskywalker (changes requested):

src/modules/Elsa.Workflows.Core/Activities/Correlate.cs
cristinamudura: Good point. Makes sense therefore I updated

sfmskywalker (approved):

PR #3554, opened 2022-12-20
Consolidate dependency references

johnwc:
Consolidate dependency references to same versions among all projects sharing same reference


sfmskywalker (approved):

johnwc:
@sfmskywalker is there a specific reason some of the projects target `netcoreapp3.1`, instead of only `netstandard2.1`, `net5.0`, etc...?

sfmskywalker:
@johnwc I don't think so, probably an omission that we should correct.

Issue #3555, opened 2022-12-20
Designer method setOutcomes causes exception


AdamJohnSwan:
Using the `setOutcomes` method in a javascript input will cause an `InvalidCastException`

I put together a test workflow definition to show what I mean.
[set-outcomes-fail.txt](https://github.com/elsa-workflows/elsa-core/files/10271835/set-outcomes-fail.txt) (change to `.json` since Github only allows uploading txt files)

To test it run the `src\samples\dashboard\aspnetcore\ElsaDashboard.Samples.AspNetCore.Monolith` project, import the workflow definition above, and then publish it.

Then hit the endpoint to trigger the workflow

```
curl --request GET \
  --url https://localhost:15265/workflows/start
```

You will get this exception:

```
System.InvalidCastException: Object must implement IConvertible.
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Jint.Runtime.Interop.DefaultTypeConverter.Convert(Object value, Type type, IFormatProvider formatProvider)
   at Jint.Runtime.Interop.DelegateWrapper.Call(JsValue thisObject, JsValue[] jsArguments)
   at Jint.Engine.Call(FunctionInstance functionInstance, JsValue thisObject, JsValue[] arguments, JintExpression expression)
   at Jint.Engine.Call(ICallable callable, JsValue thisObject, JsValue[] arguments, JintExpression expression)
   at Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateCall(EvaluationContext context, JsValue func, Object reference, NodeList`1& arguments, Boolean tailPosition)
   at Jint.Runtime.Interpreter.Expressions.JintCallExpression.Call(EvaluationContext context)
   at Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateInternal(EvaluationContext context)
   at Jint.Runtime.Interpreter.Statements.JintExpressionStatement.ExecuteInternal(EvaluationContext context)
   at Jint.Runtime.Interpreter.JintStatementList.Execute(EvaluationContext context)
   at Jint.Engine.<>c__DisplayClass65_0.<Execute>g__DoInvoke|0()
   at Jint.Engine.ExecuteWithConstraints[T](Func`1 callback)
   at Jint.Engine.Execute(Script script)
   at Jint.Engine.Execute(String source, ParserOptions parserOptions)
   at Jint.Engine.Evaluate(String source)
[...]

PR #3558, opened 2022-12-21
Different Datype for JS function setOutcomes

AdamJohnSwan:
https://github.com/elsa-workflows/elsa-core/issues/3555

casting to `<IEnumerable<string>>` causes an exception. Changing the datatype to `string[]` removes the need for the cast. I tested this successfully using the workflow in the attached issue.


sfmskywalker (approved):

sfmskywalker:
Thanks @AdamJohnSwan !

PR #3559, opened 2022-12-21
Workflow activation strategies

sfmskywalker:
This PR adds the ability for workflows to be activated only if a configured activation strategy permits so.

For example, the Singleton strategy will only allow a workflow instance to be created if there are currently no unfinished workflow instances of the same workflow definition.

To create custom strategies, simply create a class that implements `IWorkflowActivationStrategy` (and register it with DI) and have its method return true or false to control whether or not the workflow about to be activated should be activated or not.

Out of the box we have the following strategies:

- Allow always - this strategy always allows a workflow to be activated.
- Singleton - only allow the workflow to be activated if there isn't already a running instance.
- Correlation - only allow the workflow to be activated if the correlation ID to be associated with it isn't associated with any other currently running workflow.
- Correlated singleton - a combination of Singleton and Correlation, where the workflow may only be activated if it is not currently running.

Registered strategies are made avaialable through the UI:

image

When creating workflows from C#, simply configure the workflow's `Options` property with the desirered strategy type:

```csharp
public class MyWorkflow : WorkflowBase
{
    {
        builder.WithActivationStrategyType<SingletonStrategy>();
    }
}
```


gribunin:
It seems the activation strategy doesn't work anymore #5308

PR #3571, opened 2022-12-28
Implemented workflow sink module

gurkanguran:


src/modules/Elsa.Persistence.EntityFrameworkCore/Common/Store.cs
```diff
@@ -25,6 +25,13 @@ public async Task SaveAsync(TEntity entity, Func<TDbContext, TEntity, TEntity>?
         entity = onSaving?.Invoke(dbContext, entity) ?? entity;
         await dbContext.BulkInsertOrUpdateAsync(new[] { entity }, config => { config.EnableShadowProperties = true; }, cancellationToken: cancellationToken);
     }
+    
+    public async Task SaveAsync<TDto>(TEntity entity, TDto dto, Func<TDbContext, TEntity, TDto, TEntity>? onSaving = default, CancellationToken cancellationToken = default)
```
sfmskywalker, line 29: The DTO thing seems like an odd duck in the nest. Can't we do the mapping from DTO to Entity in a higher layer, same approach as is done elsewhere?

src/modules/Elsa.Persistence.EntityFrameworkCore/Modules/WorkflowSink/Configurations.cs
```diff
@@ -0,0 +1,21 @@
+using Elsa.Workflows.Sink.Models;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Elsa.Persistence.EntityFrameworkCore.Modules.WorkflowSink
+{
+    public class Configurations :
+        IEntityTypeConfiguration<WorkflowSinkEntity>
```
sfmskywalker, line 8: I think the term "Workflow Sink Entity" is confusing, because it's not intuitive what it means. What it should mean, is a view of the workflow state being stored in the database. So perhaps a better name would be e.g. "WorkflowInstance", which is more accurate.

So perhaps better would be to move the current EF core code surrounding the WorkflowInstance entity to the WorkflowSinks folder in this project and remove WorkflowSinkEntity.

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowDefinitions/Execute/Endpoint.cs
```diff
@@ -20,6 +20,7 @@ public override void Configure()
     {
         Post("/workflow-definitions/{definitionId}/execute");
         ConfigurePermissions("exec:workflow-definitions");
+        AllowAnonymous();
```
sfmskywalker, line 23: I think you added this for testing purposes. Let's remove it.
For testing these endpoints that require authenticated users, you can create an access token using the `/identity/login` endpoint using the following curl:

```shell
curl --location --request POST 'https://localhost:5001/elsa/api/identity/login' \
--header 'Content-Type: application/json' \
--data-raw '{
    "username": "admin",
    "password": "password"
}'
```

The result contains an access token you can include as a bearer token in the Authorization header. For example:

```shell
curl --location --request POST 'https://localhost:5001/elsa/api/workflow-definitions/94afd3fa039e4efa931affab952d831f/execute' \
--header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiQWRtaW4iLCJwZXJtaXNzaW9ucyI6IioiLCJuYmYiOjE2NzIwODg3MDYsImV4cCI6MTY3MjA5MjMwNiwiaWF0IjoxNjcyMDg4NzA2LCJpc3MiOiJodHRwOi8vZWxzYS5hcGkiLCJhdWQiOiJodHRwOi8vZWxzYS5hcGkifQ.6LBaF8HY_QF7sxlgj96IWmDzk_vPY9KcWgwMMxx8XMQ' \
--header 'Content-Type: application/json'
```

src/modules/Elsa.Workflows.Sink/Elsa.Workflows.Sink.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
sfmskywalker, line 6: To include common settings for Elsa library projects distributed as NuGet packages, please include these sections at the top of the file (but inside the <Project> element):

```xml
<Import Project="..\..\..\common.props" />
<Import Project="..\..\..\configureawait.props" />
```

the common props already enable Nullable, so you can remove that one.
We also need to include a description and tags.

Here's an example:

```xml
<Project Sdk="Microsoft.NET.Sdk">

    <Import Project="..\..\..\common.props" />
    <Import Project="..\..\..\configureawait.props" />

    <PropertyGroup>
        <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
        <Description>
            Provides workflow state sinks that store workflow state in various formats and locations.
        </Description>
        <PackageTags>elsa module</PackageTags>
    </PropertyGroup>
...
```

src/modules/Elsa.Workflows.Sink/Extensions/FeatureExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class FeatureExtensions
+{
+    public static WorkflowSinkFeature UseInMemoryDatabase(this WorkflowSinkFeature feature, Action<MemoryWorkflowSinkPersistenceFeature>? configure = default)
+    {
+        feature.Module.Configure(configure);
+        return feature;
+    }
```
sfmskywalker, line 12: I might be missing the point here, but I don't think we need an in-memory sink. The user either installs one or more sinks, or they don't.

src/modules/Elsa.Workflows.Sink/Extensions/FeatureExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    
+    public static WorkflowSinkFeature UseMassTransitServiceBus(this WorkflowSinkFeature feature, Action<MassTransitSinkTransportFeature>? configure = default)
+    {
+        feature.Module.Configure(configure);
+        return feature;
+    }
```
sfmskywalker, line 18: I think the dependency between Workflow Sink and MassTransit should be the inverse: the Workflow Sink module provides a transport contract, but not the actual implementation. That is the role of e.g. Elsa.MassTransit, which then provides a concrete implementation of ISinkTransport. Same goes for the EF Core one. 

This way, whenever we add another sink implementation, we just add a new project without having to update the Elsa.Workflows.Sinks project itself (i.e. open/closed).

src/modules/Elsa.Workflows.Sink/Extensions/ModuleExtensions.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using Elsa.Features.Services;
+using Elsa.Workflows.Sink.Features;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class ModuleExtensions
+{
+    public static IModule UseWorkflowSink(this IModule module, Action<WorkflowSinkFeature>? configure = default)
```
sfmskywalker, line 10: Let's use the plural form `UseWorkflowSinks` - since we're gonna use a feature, within which we may enable one or more sinks.

src/modules/Elsa.Workflows.Sink/Features/WorkflowSinkFeature.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// <summary>
+    /// A factory that instantiates a concrete <see cref="ISinkTransport"/>.
+    /// </summary>
+    public Func<IServiceProvider, ISinkTransport> SinkTransport { get; set; } = sp => ActivatorUtilities.CreateInstance<InProcessSinkTransport>(sp);
+    
+    /// <summary>
+    /// A factory that instantiates a concrete <see cref="IWorkflowSinkClient"/>.
+    /// </summary>
+    public Func<IServiceProvider, IWorkflowSinkClient>? WorkflowSinkClient { get; set; } = sp => ActivatorUtilities.CreateInstance<MemoryWorkflowSinkClient>(sp);
```
sfmskywalker, line 33: As mentioned in a previous comment, I think we don't need a memory sink and can remove this one.

Elsa.sln
```diff
@@ -121,8 +121,12 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Samples.HelloWorld", "
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Testing.Shared", "src\common\Elsa.Testing.Shared\Elsa.Testing.Shared.csproj", "{AD8C4BB0-C78F-496C-B326-1D82C02C7568}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Workflows.Sink", "src\modules\Elsa.Workflows.Sink\Elsa.Workflows.Sink.csproj", "{7F1FDA45-C731-4DCC-BFF6-403D4B857A93}"
```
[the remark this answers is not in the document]
gurkanguran: Yes, that makes sense!

src/modules/Elsa.Persistence.EntityFrameworkCore/Common/Store.cs
gurkanguran: Unfortunately not without exposing DbContext to higher levels. I could also include the properties of dto in the entity and ignore them in EntityTypeConfiguration. But there was a problem with serializing in that way + I find it cleaner to separate dto and entity and keep entity as clean as possible(only with the properties we want to persist). 

I can expose the dbContext to higher level if you dont find this approach clean?

src/modules/Elsa.MassTransit/Extensions/ModuleExtensions.cs
```diff
@@ -1,13 +1,43 @@
...
+    public static void AddMassTransitServiceBusConsumerType(this IModule module, Type type)
+    {
+        var types = DictionaryExtensions.GetOrAdd(module.Properties, ServiceBusConsumerTypesKey, () => new HashSet<Type>());
+        types.Add(type);
+    }
+
+    /// <summary>
+    /// Returns all collected types for discovery of service bus consumers.
+    /// </summary>
+    private static IEnumerable<Type> GetServiceBusConsumerTypesFromModule(this IModule module) => DictionaryExtensions.GetOrAdd(module.Properties, ServiceBusConsumerTypesKey, () => new HashSet<Type>());
```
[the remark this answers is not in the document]
gurkanguran: How can I call this from other features then? The purpose was to be able to use it from different modules like FastEndpoints implementation.

src/modules/Elsa.MassTransit/Extensions/ModuleExtensions.cs
```diff
@@ -1,13 +1,43 @@
...
 
 public static class ModuleExtensions
 {
-    public static IModule AddMassTransitWorkflowDispatchers(this IModule module)
+    private static readonly object ServiceBusConsumerTypesKey = new();
+
+    /// <summary>
+    /// Registers the specified type for MassTransit service bus consumer discovery.
+    /// </summary>
+    public static void AddMassTransitServiceBusConsumerType(this IModule module, Type type)
```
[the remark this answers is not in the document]
gurkanguran: Same as below

PR #3575, opened 2022-12-30
Implement authorization_code grant flow with OAuth2

tanelkuhi:
1. Inside secret dialog, if you choose grant type `Authorization code`, a button to "Authorize" appears.
2. When you click that button, a popup window appears which takes you to the OAuth2 provider
3. When authorizing is finished, the provider makes a request to url we gave it when we started authorization, which is `/v1/oauth2/callback`
4. The callback endpoint receives the autorization code and requests the provider for the tokens and stores them inside the secret's properties as a json with key `Token`
5. The endpoint redirects to `/oauth2-authorized` which is an url in frontend and which sends a postmessage to its `window.opener` (=our own page) to update the secrets.
6. The button disappears, because the tokens have been attached to the secret
7. When a SendHttpRequest activity tries to refresh access token, but fails due to refresh token expiration, the `Token` property is removed from the secret and the user will start to see "Authorize" button in the secret dialog again.


src/modules/secrets/Elsa.Secrets/Providers/SecretsProvider.cs
```diff
@@ -74,19 +74,9 @@ public async Task<ICollection<string>> GetSecretsAsync(string type)
...
-                if (formatter != null)
-                    return secrets.ToDictionary(x => x.Name ?? x.DisplayName ?? x.Id, x => formatter.FormatSecretValue(x).Result);
-            }
-            catch (Exception ex)
-            {
-                _logger.LogError(ex, "Error creating dictionary of secrets for {type}", type);
-            }
-            return new Dictionary<string, string>();
+            return secrets
+                .GroupBy(x => x.Name).Select(x => x.First())
```
johnwc, line 78: Why is this grouped by Name? This should be returning a dictionary with key as `<secret type>:<secret name>`. This should just be `return secrets.ToDictionary(x => $"{x.Type}:{x.Name}", x => x.Name ?? x.DisplayName ?? x.Id);`

tanelkuhi: `GroupBy(x => x.Name).Select(x => x.First())` gets rid of duplicates, since this is a dictionary anyways and the secrets to be used get matched by Name.

tanelkuhi: If that is the case, I am surprised it wasn't this way already before. I will change it then.

ke-real: @johnwc where do you see the constraint? It would be logical to have one as it is quite safe to assume noone needs duplicate names especially within the same type (as you wont be able to distinguish between the two identical select options). I think currently there is no constraint except for PK and the name+type requirement you introduced. But also we have to keep in mind that different types of storages might be in use that don't support Unique indexes - so this might need to be handled in business logic instead. Anyway it seems like a minor issue and selecting groupBy+first is quite safe (especially when for someone there are alrady some duplicate names in the database and launching updated Elsa would fail) - we also handled referencing/handling secrts as first or default.

but yes good catch about key should be type+name

tanelkuhi: I tested. I can easily create secrets of same type and with same name again without anything stopping me.

johnwc: > I tested. I can easily create secrets of same type and with same name again without anything stopping me.

You tested this with what RDMS? MySql has constraints, if the others don't, they should be updated to include them as well.

johnwc: > @johnwc where do you see the constraint? It would be logical to have one as it is quite safe to assume noone needs duplicate names especially within the same type (as you wont be able to distinguish between the two identical select options). I think currently there is no constraint except for PK and the name+type requirement you introduced. But also we have to keep in mind that different types of storages might be in use that don't support Unique indexes - so this might need to be handled in business logic instead. Anyway it seems like a minor issue and selecting groupBy+first is quite safe (especially when for someone there are alrady some duplicate names in the database and launching updated Elsa would fail) - we also handled referencing/handling secrts as first or default.
> 
> but yes good catch about key should be type+name

Quite safe is creating a bad UX, when the user doesn't see their second secret, or they don't understand why the wrong secret keeps getting applied in the activity at runtime. 

Seeing there is ZERO documentation on how to use the credential manager, along with no working sample; the likely hood of an existing user having duplicate names already is very minuet.

@sfmskywalker  This needs be addresses before it gets too far with deployments in the wild.

sfmskywalker: @johnwc I agree. Unfortunately I don't have the bandwidth. @tanelkuhi is there any chance you can provide a basic README that explains how to enable and configure the credential manager? A sample project would be helpful as well.

johnwc: @sfmskywalker  @tanelkuhi we also need to address the issue of other persisted storages for secrets allowing multiple secrets with same name and type. The MySql persisted storage already has a constraint set to make sure secret type & name are unique.

ke-real: > Quite safe is creating a bad UX, when the user doesn't see their second secret
If that is the case then user is creating a bad UX for himself by creating duplicate creds that you would not be able to distinguish. It is far worse UX when app crashes after upgrade without a reason. Having a unique constraint that would work for all storages is the only proper solution here and it is currently not there for any storage (I don't know what is in your MySql instance but there is no unique cosntraint definition in any of the storage migrations in master branch). Also Elsa might be lacking API error handling in general.
In any way this is out of scope of this specific PR, we can exclude the dictionary handling completely and let it be fixed in scope of  #3520 that breaks the intended logic - however until then it will fall back to the initial limited functioning when every secret is stored as a value under each workflow version.

We will add a Readme explaining the main concepts, sample project is rather redundant as Secrets are enabled in the sample host and the readme will expalin how to use it (no pre-configuration is required other than enabling in config, just click around).

sfmskywalker: Thanks @ke-real !

sfmskywalker: I agree, input validation, name uniqueness etc. needs to be tackled holistically - it's not just a problem with this module.

sfmskywalker (approved):

Issue #3592, opened 2023-01-03
Webhooks module


sfmskywalker:
The Webhooks module builds on the RunTask activity (#3591) and asynchronously invokes all registered webhooks.
This makes it easy to integrate external applications with a workflow.

As part of this module, a UI should be provided to allow the user to manage webhook registrations from the designer.
It should also be possible to register webhooks from appsettings (using standard IConfiguration and Options API).

PR #3598, opened 2023-01-05
Webhooks

sfmskywalker:
This PR adds a new module called **Webhooks**.

The module allows users to register HTTP endpoints to be invoked when certain events occur.

Currently, it only invokes webhooks for the "RunTask" event, but the goal is for this module to be extensible by other modules that can add more webhook event types.

Webhook endpoints can be configured via appsettings, but additional sources can be implemented (e.g. a DB persistent provider could be added to allow users to define webhooks from the dashboard)

Combining the webhooks module together with the RunTask activity from a workflow makes it easy to integrate external applications with a workflow via HTTP communications, without necessarily having to write custom activities.

Unrelated to the Webhooks module, but a future addition could be to have an activity type provider that dynamically produces "RunTask" activities based on task descriptions created by the user.


sfmskywalker:
Closes #3592

PR #3601, opened 2023-01-05
Fix for dynamic model changes are not triggering re-rendering x6 designer

sabitertan:
Found issue while using WorkflowEditor blazor component in dynamically changing workflow definition ids. Api calls were successful, model setup correctly by elsaStudio but x6 designer weren't re-rendering. This PR fixes that issue.


sfmskywalker (approved):
Thanks!

PR #3623, opened 2023-01-13
Elastic updates

sfmskywalker:
This PR contains a few refactorings, mostly to take advantage of DI. Also fixes listing workflow instances when no filter is supplied.


src/modules/Elsa.Elasticsearch/Features/ElasticsearchFeature.cs
```diff
@@ -1,59 +1,67 @@
...
+    /// <summary>
+    /// True to enable index name rollovers, false otherwise.
+    /// </summary>
+    public bool EnableRollover { get; set; }
+
+    /// <inheritdoc />
     public override void ConfigureHostedServices()
     {
-        Module.ConfigureHostedService<ConfigureMappingHostedService>(-1);
+        Module.ConfigureHostedService<ConfigureElasticsearchClientHostedService>(-2);
```
gurkanguran, line 39: I think this line can be also under if EnableRollover condition as it is creating an alias which is redundant if a rollover is not being used.

Also, ConfigureElasticsearchClientHostedService can be renamed to ConfigureAliasesHostedService as it is the only responsibility of the service.

src/modules/Elsa.Elasticsearch/HostedServices/ConfigureIndexRolloverHostedService.cs
```diff
@@ -1,30 +1,52 @@
...
-        var job = new ConfigureIndexRolloverJob();
-        var schedule = new CronSchedule
+        while (!stoppingToken.IsCancellationRequested)
         {
-            // At the beginning of every month
-            CronExpression = "0 0 1 * *"
-        };
-        
-        await _jobScheduler.ScheduleAsync(job, GetType().Name, schedule, cancellationToken: cancellationToken);
+            await Task.Delay(_options.RolloverInterval, stoppingToken);
```
gurkanguran, line 33: I think if we use time interval instead of cron schedule (which makes sense) then we don't need a hosted service. We can basically pass MaxAge condition parameter with time interval data to rollover api of elastic client. This can be done directly in ConfigureClient method of IElasticConfiguration.

sfmskywalker: Oh, nice! Then that simplifies things significantly 👍🏻

sfmskywalker: Done

src/modules/Elsa.Elasticsearch/Features/ElasticsearchFeature.cs
sfmskywalker: Makes sense, I'll rename it 👍🏻

PR #3628, opened 2023-01-14
Fix #3621 by waiting for a resize event with width and height > 0

### LINKED ISSUE #3621: Workflow visualizer is partially broken in latest version of Chrome (109.0.5414.75) and Edge (109.0.1518.49)
Hello Elsa Team!

I'd like to report a strange problem that I've spotted recently in Elsa workflow visualizer, after upgrading Chrome to the latest version (109.0.5414.75): the **first rendering** of the workflow diagram is broken.

When I open a workflow definition or a workflow instance, what is initially displayed is not the workflow as expected, but this:
image

After dragging the diagram (to see all of it), it turns out that what is actually rendered is some kind of _mini skeleton_ of the workflow, with only the connections between activities:
image

Note: I don't see any error in the JavaScript console:
image

Oddly enough, if I refresh the page at that point (**F5**), the bug _appears_ to be gone (the rendering is normal).
Here's what it looks like (after zooming out, so you can compare with my second screenshot):
image

The booby-trap is that the next **first** rendering of the same diagram will trigger the same bug again (e.g. by going back to the list of workflow definitions and reopening the same workflow definition).

Initially, I thought the problem was in Elsa as it so happened that I had upgraded from 2.8.2 to 2.9.3 _on the same day_ that I upgraded Chrome to its latest version. So I've tried in Edge and I saw that the problem wasn't present. I might have jumped to conclusions at that point, but then I also noticed that Edge had not yet been upgraded to its own latest version. So I upgraded it and  **the same visual problem also appeared**!

Notes:
- Force-reloading the page (**Ctrl+F5**) doesn't do anything more than **F5**, in either browser.
- Even after downgrading Elsa to 2.8.2, the problem is still present, so it's clearly linked to the latest version of the browsers, and the previous version of them (108.0.*) didn't have the problem.
- Chrome version (with the problem) is **Version 109.0.5414.75 (Official Build) (64-bit)**. I don't remember the precise previous Chrome version.
- Initial Edge version (without the problem) was **Version 108.0.1462.76 (Official build) (64-bit)**.
- Final Edge version (with the problem) is **Version 109.0.1518.49 (Official build) (64-bit)**.

I don't know it the problem lies in the way Elsa draws workflow diagrams (which might somehow use a non-standard HTML/CSS/JavaScript/SVG feature) or if both browsers both have introduced a _regression_ related to that area in their latest version, but you should be aware of this. I won't be the only Elsa user using these versions of Chrome or Edge.

Thanks in advance and keep up the good work !

mohdali:
Fix for #3621 by waiting for a resize event with width and height > 0 before rendering.

### DIFF
--- a/src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx
+++ b/src/designer/elsa-workflows-studio/src/components/designers/tree/elsa-designer-tree/elsa-designer-tree.tsx
@@ -269,8 +269,11 @@ export class ElsaWorkflowDesigner {
     if (rect.height === 0 || rect.width === 0) {
       const observer = new ResizeObserver(entries => {
         for (let entry of entries) {
-          this.renderTree();
-          observer.unobserve(this.el);
+          const rect = this.el.getBoundingClientRect();
+          if (rect.height > 0 && rect.width > 0) {
+            this.renderTree();
+            observer.unobserve(this.el);
+          }
         }
       });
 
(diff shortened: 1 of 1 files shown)


sfmskywalker (approved):
Nice find! 👍🏻

PR #3630, opened 2023-01-16
Removed EFCore.BulkExtensions and used EF7 bulk operations

gurkanguran:


src/modules/Elsa.EntityFrameworkCore/Extensions/ExpressionExtensions.cs
```diff
@@ -0,0 +1,30 @@
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Elsa.EntityFrameworkCore.Extensions;
+
+public static class ExpressionExtensions
+{
+    public static PropertyInfo GetPropertyInfo<TType, TReturn>(this Expression<Func<TType, TReturn>> propertyExpression)
```
sfmskywalker, line 8: I think you should be able to reuse the `PropertyAccessorExtensions` found in `Elsa.Workflows.Core`.

sfmskywalker (approved):

src/modules/Elsa.EntityFrameworkCore/Extensions/ExpressionExtensions.cs
sfmskywalker: Done.

Issue #3627, opened 2023-01-14
HTTP Endpoint Body


sfmskywalker:
The HTTP Endpoint needs to expose the received body content.

PR #3631, opened 2023-01-16
HTTP activity body parsing

sfmskywalker:
This PR implements request content parsing for both the HTTP Endpoint and Send HTTP Request activities.

The content is parsed automatically into whatever target variable type is specified.


sfmskywalker:
Resolves #3627

PR #3654, opened 2023-01-26
Use workflow definition as activity

gurkanguran:


src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
```diff
@@ -60,6 +60,12 @@ public async Task<IEnumerable<WorkflowDefinition>> FindManyByDefinitionIdAsync(s
         return await _store.FindAsync(predicate, Load, cancellationToken);
     }
 
+    public async Task<IEnumerable<WorkflowDefinition>> FindByPredicateAsync(Expression<Func<WorkflowDefinition, bool>> predicate, VersionOptions versionOptions, CancellationToken cancellationToken = default)
```
sfmskywalker, line 63: What if the implementation of `IWorkflowDefinitionStore` does not support translating arbitrary expressions to the underlying storage mechanism? For example, some users may implement an HTTP API client underneath, which makes translating these expressions considerably hard, if not impossible.

gurkanguran: Good point, I will hide that implementation at a lower level.

sfmskywalker (approved):

PR #3662, opened 2023-02-02
WorkflowDefinitions: OrderBy and SortBy

brunomartinspro:
- Added support for the endpoint /v1/workflow-definitions to order by Name, Description or CreatedAt
- Added support for the endpoint /v1/workflow-definitions to sort by Descending or Ascending
- Added UnitTests to test these cases
- Added Postman Collection for testing these cases
- Added field CreatedBy to WorkflowDefinitionSummaryModel 
- Added Swagger UI enum support as strings
- Added Swagger specification enum OpenApi Generator support as strings
- Renamed current OrderBy enum that was only being used in Workflow Instances to: WorkflowInstanceOrderBy and created WorkflowDefinitionOrderBy

#### Swagger UI Support
image

image

#### Unit Tests
image

#### Postman Tests
image

 ### Note
if the pipeline can create a review app, or after a deploy, we can use newman to run the postman collections we can guarantee this doesn't get broken in the future.


sfmskywalker (approved):
This looks really good. My only concern is with the existing persistence providers, like MongoDB, YesSQL and maybe even EF Core. Any chance you tested at least the EF core one?I don't expect any issues with EF core, really, since these expressions should translate well, but would be good to know for sure.

brunomartinspro:
> 

I will do more testes locally but i reused the existing classes, so my expectation is that all still works.
I will get back to you later today.

sfmskywalker:
@brunomartinspro Much appreciated, thank you for double checking. When you confirm it's good to go, we'll get this merged right away.

brunomartinspro:
Hey @sfmskywalker, 
i added a postman collection with tests to validate if the entire flow works as intended with the sample application and updated the description.
I tested it locally against the default values of the app settings of the project Elsa.Samples.Server.Host by running the script run-server.ps1
Do you feel these test are sufficient?
thank you

sfmskywalker:
@brunomartinspro Yes, splendid. And you didn't run into DB query exceptions or anything like that, so should be all good. Thanks again!

Issue #3667, opened 2023-02-02
Feature: Azure Service Bus Tenant Id Resolver


Nico-VanHaaster:
This is a feature request with a correlating Pull Request.

As a multi-tenant solution, the `Elsa.Activitites.AzureServiceBus` is not capable of executing workflows for a specific tenant when receiving a message. This means that mult-tenancy can not easily be achieved with these activities.

Request:
Introduce the capability for a the AzureServiceBus Worker to provide the workflow tenant Id when the worker receives a message.

Impact:
By default the Elsa product will not provide a tenant Id and the consuming application (custom application) can override the ability to resolve the tenant id (through DependencyInjection).

Nico-VanHaaster:
Please review PR and advise. Cheers.

PR #3668, opened 2023-02-02
feat: 🎸 AzureServiceBus - TenantIdResolver

Nico-VanHaaster:
This PR is in response to #3667 Feature: Azure Service Bus Tenant Id Resolver.

Introduces a new interface `IAzureServiceBusTenantIdResolver` with a single method `ResolveAsync`  this method will pass all the relevant details from the Worker to the Resolver.

By Default Elsa will ship with a "NoOp" resolver called `DefaultAzureServiceBusTenantIdResolver` which returns a null string.

This will be added to dependency injection as a singleton (resolved from the root scope when workers are created).

Applications may override this behavior by registering a custom instance of the `IAzureServiceBusTenantIdResolver` **after** the call to `AddAzureServiceBusActivities(...)`.

Such as the code snippet below.

```
services.AddElsa(c =>
              {
                 // ...omitted
                  c.AddAzureServiceBusActivities(az => az.ConnectionString = "");
              })
              .AddSingleton<IAzureServiceBusTenantIdResolver, CustomAzureServiceBusTenantIdResolver>();
```


sfmskywalker (approved):
Nice one 👍🏻

PR #3673, opened 2023-02-04
fix: add missing tenant id specifiers; limit tailwind css to elsa-studio-root

sjd2021:
This PR attempts to address two issues my team is facing:

1. Some missing tenant id filters
2. Elsa's css files conflict with many of our website's styles because they aren't limited to `elsa-studio-root` and below


sfmskywalker (approved):

sfmskywalker:
They should be excluded but got included at some point accidentally and I never bothered to fix that. Since they are generated, no need to update them. Thanks again for the contribution.

sfmskywalker:
@sjd2021 For some reason, adding the TailwindCSS configuration line causes some issues with some of the elements not being hidden. See also #3695

When I remove that line and rebuild the TailwindCSS file, the problem disappears.

I will revert this change for the 2.10.2.2 version so that other users can upgrade, but of course I am happy to add the scoping again once we figure out why it breaks (at least for me and others).

Issue #3661, opened 2023-02-02
Feature Request: Kafka Schema Registry & Auto.Offset.Reset


Snotax:
Hi, thanks for the v1 of the Kafka implementation for Elsa (@borjaferv / @sfmskywalker).  

I have some more ideas for a next version that would greatly improve the current implementation. 
If something is already implemented and I missed it, feel free to ignore the request.
 

- [ ] Kafka implementation should support the integration of a Kakfa.SchemaRegistry to create Triggers based on different Schemas
- [x] Auto.Offset.Reset property should be configurable for each Trigger (or at least per server instance)
- [x] KafkaMessageReceiver Trigger should be customizable to attach certain custom code on consume.

Snotax:
Hi @sfmskywalker 
Can you tell me if my assumption is correct, that if I want to override an existing Trigger (e.G. KafkaMessageReceived), I have to create a new BookmarkProvider just for this new Trigger (KafkaReceiveOverride)? 

Or is there another way to overwritte an existing trigger without providing a BookmarkProvider for each new Trigger?

PR #3702, opened 2023-02-14
Feature Activities Kafka: Overridable Trigger feat: 3661

Snotax:
Features:
- Added the possibility to add custom triggers that override the default KafkaMessageReceived Trigger. 

This helps when we want to implement custom business logic in the Kafka trigger activity without adding a second activity after the trigger. 

Open to suggestions.


Snotax:
#3661

sfmskywalker (approved):

PR #3717, opened 2023-02-19
Retention specification fixes

martinhjulstrom:
I couldn't get the current default retention to work. The default specification generated a query that was looking for a workflow with the status running AND finished, which could never be true. This PR adjusted that to be OR instead.

I might misunderstand something but to me it feels like only finished workflows should be removed by default so I removed the running from the list

### DIFF
--- a/src/modules/retention/Elsa.Retention/Specifications/CompletedWorkflowFilterSpecification.cs
+++ b/src/modules/retention/Elsa.Retention/Specifications/CompletedWorkflowFilterSpecification.cs
@@ -1,13 +1,11 @@
 using Elsa.Models;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Elsa.Retention.Specifications
 {
     public class CompletedWorkflowFilterSpecification : WorkflowStatusFilterSpecification
     {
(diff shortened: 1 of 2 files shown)


sfmskywalker:
Makes sense to me. @jdevillard do you see any issues with this?

jdevillard:
Hello, 

Thanks for pointing this @martinhjulstrom , I think you're true. Don't remember how I've missed out this.

johnwc:
I came to create an issue about this as well. I noticed our instances weren't getting cleaned up. Ran a debug to see the query and notice 2 things. The issue this PR fixes, and that the query is using a `&` instead of the word `AND`. I believe using `&` in sql will cause it to do a bitwise check instead of logical grouping.

```
SELECT `w`.`Id`
FROM `WorkflowInstances` AS `w` 
WHERE `w`.`CreatedAt` <= '2023-03-17'
	AND ((`w`.`WorkflowStatus` = 1) & (`w`.`WorkflowStatus` = 2)) 
ORDER BY `w`.`CreatedAt` DESC 
LIMIT 1000 OFFSET 0
```

PR #3718, opened 2023-02-19
DeleteMany fixes

martinhjulstrom:
BatchDeleteWithWorkAroundAsync was referencing an issue that looks to be resolved, I tested it and it works on a postgres database, I was however not able to test it on a mysql db, so I didn't change the mysql implementation.

DeleteMany generated sql that couldn't be executed with psql (WorkflowInstanceId would need to be within quotation marks) but the batch delete looked like it generated performant sql.


sfmskywalker (approved):
Thanks 😃

johnwc:
@martinhjulstrom if you can give me quick steps how to test, I can get it tested officially for MySQL.

PR #3739, opened 2023-02-27
Fix credential manager in designer.

Chryssie:
Fixes two issues:
1. Fix the `elsa-credential-manager-items-list` failing to open from the Elsa dashboard due to the `monacoLibPath` not being injected from the tunnel.
2. Fixes the `elsa-secret-editor-modal` throwing errors for secrets that don't have any of the following properties:
   1. `GrantType`
   2. `AuthorizationUrl`
   3. `AccessTokenUrl`
   4. `ClientId`
   5. `ClientSecret`


sfmskywalker (approved):
Awesome1

PR #3749, opened 2023-03-01
 Implemented workflow instance journal partially

gurkanguran:
This PR is extending the journal design to a new journal tab in activity properties which can be accessed by clicking the status of the execution log in the journal.

The journal tab contains exception info for faulted activities for now. It will be extended later with other info for successful activities.


src/modules/Elsa.Workflows.Core/Models/Activity.cs
```diff
@@ -143,10 +143,15 @@ protected async ValueTask CompleteAsync(ActivityExecutionContext context)
 
     async ValueTask IActivity.ExecuteAsync(ActivityExecutionContext context)
     {
+        // Invoke execution logging behavior prior to execution start.
+        var executionLoggingBehavior = Behaviors.First(x => x.GetType() == typeof(ExecutionLoggingBehavior));
```
sfmskywalker, line 147: Can we add a "Priority" property to behaviors so that we can sort by them, instead of having hardcoded knowledge about the ExecutionLoggingBehavior?

sfmskywalker (approved):

sfmskywalker (approved):

PR #3781, opened 2023-03-08
Implement Customize SqlClientFactory with discovery database types

mircotamburini:
i need to implement Sql Activities for AS400.
i add possibilities to pass CustomSqlClientFactory that reference AS400 drivers and automatic discover database types from Activities.


sfmskywalker (commented):
Looks good! Just one small request: can you sort the using statements?

PR #3857, opened 2023-03-29
Handle multiple output connections on import

gurkanguran:


sfmskywalker (approved):
Great work @gurkanguran !

PR #3875, opened 2023-04-05
Fix HttpEndpointDocumentFilter throwing when workflows mix HttpEndpoint triggers with other triggers.

Chryssie:


sfmskywalker (approved):
Thanks @Chryssie !

Issue #3907, opened 2023-04-14
Refactor Workflow Instance serialization


sfmskywalker:
Currently, workflow instances containing multiple object references to the same object, are serialized multiple times (once per reference). This can potentially lead to large workflow instance JSON strings.

We should refactor this, such that multiple references to the same object are serialized as exactly that: references.
Newtonsoft supports this out of the box, but we (presumably) need to write a custom converter or rely on a third party converter to implement reference serialization.

PR #3913, opened 2023-04-16
Improve JSON serializationfor workflow state

sfmskywalker:
This PR improves JSON serialization for Workflow State objects by taking advantage of reference handling to avoid multiple serialization of the same object reference.


sfmskywalker:
Closes #3907

PR #3920, opened 2023-04-18
#3845 Tooltip Fixed Positioning

Daxton47:
Tooltip component now calculates fixed positions to fix issues with hidden overflow. Ideally, I think the tooltip DOM element should be "teleported"(Vue) or "portaled"(React) to a higher node in the DOM tree, but I wasn't able to get around to that with Stencil. 

New proposed Tooltip Component API:
tooltipPosition?: `right`, `left`, `top`, `bottom`, `auto`. Defaults to `right`

Specifying `auto` will default to `bottom`, but will reposition to `top` if the tooltip height runs offscreen:

https://user-images.githubusercontent.com/38263263/232676324-1307a8f0-bfbc-4c39-a1a6-ec5646a3c917.mp4

NOTE: Temporary `auto` demo. Tooltip for activities still uses `right` :) 

#3845

### DIFF
--- a/src/designer/designer_packages/elsa-workflows-designer/src/components.d.ts
+++ b/src/designer/designer_packages/elsa-workflows-designer/src/components.d.ts
@@ -243,7 +243,10 @@ export namespace Components {
     }
     interface ElsaTooltip {
         "tooltipContent": any;
-        "tooltipPosition"?: string;
+        /**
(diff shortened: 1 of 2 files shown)


Daxton47:
@dotnet-policy-service agree

sfmskywalker (approved):
Amazing, thank you!

PR #3950, opened 2023-04-25
Auto update consuming workflows' composite activity references

gurkanguran:


src/designer/designer_packages/elsa-workflows-designer/src/modules/workflow-definitions/plugins/workflow-definitions-plugin.tsx
```diff
@@ -126,7 +126,7 @@ export class WorkflowDefinitionsPlugin implements Plugin {
       await this.workflowDefinitionEditorElement.loadWorkflowVersions();
     }
 
-    return definition;
+    return {...definition, consumingWorkflowsBeingUpdated: updatedWorkflow.consumingWorkflowsBeingUpdated};
```
sfmskywalker, line 129: Suggestion: `consumingWorkflowsBeingUpdated` -> `affectedWorkflows`. I understand that the process is still happening in the background somewhere, so technically your field name is spot-on, but `affectedWorkflows` feels simpler to interpret mentally and is close enough to the truth.

src/designer/designer_packages/elsa-workflows-designer/src/modules/workflow-definitions/plugins/workflow-definitions-plugin.tsx
```diff
@@ -223,8 +223,19 @@ export class WorkflowDefinitionsPlugin implements Plugin {
...
-      .then(async () => {
+      .then(async (response) => {
         NotificationService.updateNotification(notification, {title: 'Workflow published', text: 'Published!'})
+
+        setTimeout(function() {
+          if(response.consumingWorkflowsBeingUpdated.length > 0)
+          NotificationService.createNotification({
+            title: 'Consuming Workflows',
+            id: uuid(),
+            text: 'Following consuming workflows are being published to update the references of this composite activity:\n\n' + response.consumingWorkflowsBeingUpdated.join('\n'),
```
sfmskywalker, line 234: Proper English would be to start the sentence with: `The` ('The following consuming ya da ya da').

src/designer/designer_packages/elsa-workflows-designer/src/modules/workflow-definitions/components/workflow-properties/properties.tsx
```diff
@@ -179,6 +180,18 @@ export class WorkflowDefinitionPropertiesEditor {
             <option value="true" selected={this.workflowDefinition.usableAsActivity}>Yes</option>
           </select>
         </FormEntry>
+      },
+      {
+        name: 'autoUpdateConsumingWorkflows',
+        order: 0,
+        content: () => 
+          <CheckboxFormEntry fieldId="UpdateConsumingWorkflows" label="Auto-update consuming workflows" hint="When you publish new version, all the consuming workflows will be updated to point the new version of this workflow.">
```
sfmskywalker, line 188: `"When you publish **a** new version, all **of** the consuming workflows will be updated to point **to** the new version of this workflow."`

src/designer/designer_packages/elsa-workflows-designer/src/modules/workflow-definitions/plugins/workflow-definitions-plugin.tsx
sfmskywalker: I see what you mean. I didn't notice the context of where this code was living, I assumed it would be it's own API call. But in this context, your previous name is better. But even better than better would be to decouple the auto-updating of consuming workflows from the logic that simply saves the current workflow. In other words, the client app would make two calls:

1. Saving the workflow definition, as per usual.
2. Updating the consuming workflows.

This would also simplify the backend logic because that too now becomes separate.
These operations don't need to succeed or fail as a whole either: if the saving of the workflow succeeds, but the publishing of the consuming workflows fails, for whatever reason, that's OK.

What do you think?

sfmskywalker: Perfect. Let's do the signalr thing separately, but separate out the updating of the consuming workflows in this PR - at least, I don't see a reason to push this first and then changing it immediately, but you'll correct me if I'm wrong 😄

src/designer/designer_packages/elsa-workflows-designer/src/modules/workflow-definitions/plugins/workflow-definitions-plugin.tsx
```diff
@@ -270,4 +273,30 @@ export class WorkflowDefinitionsPlugin implements Plugin {
...
+      text: 'Consuming workflows are being updated.',
+      type: NotificationDisplayType.InProgress
+    });
+
+    await this.api.updateWorkflowReferences({definitionId: definition.definitionId})
+      .then(async (response) => {
+        debugger;
+        NotificationService.updateNotification(notification, {
+          title: 'Consuming Workflows', 
+          text: 'The following consuming workflows have been successfully updated:\n\n' + response.affectedWorkflows.join('\n'),
```
sfmskywalker, line 290: If there were 0 affected workflows, the text should be a little bit different, e.g. `There were no consuming workflows`. Or perhaps the notification shouldn't display in the first place. What do you think?

sfmskywalker (approved):
It looks & works great! Just one smal improvement related to the notification message in case 0 workflows were affected.

PR #3951, opened 2023-04-25
Fix authorization code flow to work with openiddict library

tanelkuhi:
When I tried to make authorization code flow with one of our apps that uses OpenIddict, there were some validation errors. So these changes should fix that.

### DIFF
--- a/src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
+++ b/src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
@@ -48,23 +48,25 @@ public class OAuth2TokenService : IOAuth2TokenService
 			{ "grant_type", secret.GetProperty("GrantType") },
 			{ "client_id", clientId },
 			{ "client_secret", clientSecret },
-			{ "scope", secret.GetProperty("Scope") }
+			{ "offline_access ", "true" }
 		};
 
 		if (authCode != null)
 		{
 			content.Add("code", authCode);
 		}
+		else
+		{
+			content.Add("scope", secret.GetProperty("Scope"));
+		}
 		if (redirectUri != null)
 		{
 			content.Add("redirect_uri", redirectUri);
 		}
-
 		try
 		{
 			var httpClient = _httpClientFactory.CreateClient(nameof(OAuth2TokenService));
 			httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-			httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", Base64Encode(clientId + ":" + clientSecret));
 			
 			var response = await httpClient.PostAsync(secret.GetProperty("AccessTokenUrl"), new FormUrlEncodedContent(content));
 			var json = await response.Content.ReadAsStringAsync();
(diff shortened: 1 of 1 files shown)


sfmskywalker (approved):
Very nice!

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
```diff
@@ -48,23 +48,25 @@ public async Task<TokenResponse> GetToken(Secret secret, string? authCode, strin
 			{ "grant_type", secret.GetProperty("GrantType") },
 			{ "client_id", clientId },
 			{ "client_secret", clientSecret },
-			{ "scope", secret.GetProperty("Scope") }
+			{ "offline_access ", "true" }
```
kevinchalet, line 51: Hey 👋🏻 

FYI, `offline_access` is not a standard parameter (and not used by OpenIddict unless you add custom logic to handle it). It also seems there's an extra space at the end of the parameter name so it's likely something you'll want to fix/remove 😃

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
```diff
@@ -48,23 +48,25 @@ public async Task<TokenResponse> GetToken(Secret secret, string? authCode, strin
...
 		if (redirectUri != null)
 		{
 			content.Add("redirect_uri", redirectUri);
 		}
-
 		try
 		{
 			var httpClient = _httpClientFactory.CreateClient(nameof(OAuth2TokenService));
 			httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-			httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", Base64Encode(clientId + ":" + clientSecret));
```
kevinchalet, line 67: Note: OpenIddict supports sending the client credentials in the `Authorization` header (basic authentication) or in the request form, but as per the spec, it's illegal to send credentials using both methods (which is likely why you were getting an error).

That said, if this client is expected to be used with other vendors that don't support sending the credentials in the request form, this change might break them. You may want to add an option to control how the credentials are sent 😃

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
tanelkuhi: Damn, you are right. "offline_access" is probably not required indeed, but was something I added when trying to get refresh tokens to work. I will create a new PR to remove this.

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
tanelkuhi: Yep, I indeed got an error, because the credentials were specified in 2 ways. Do you think every there are vendors, that support only credentials in request form and now in headers?

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
mohdali: @tanelkuhi If you create a PR with the required corrections we may be able to include it in a 2.11.2 release soon

PR #3970, opened 2023-04-27
Make it choosable where client id and secret are put and improve expi…

tanelkuhi:
Make it choosable where client id and secret are put and improve expiration check

### DIFF
--- a/src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/oauth2.secret.ts
+++ b/src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/oauth2.secret.ts
@@ -78,13 +78,37 @@ export const OAuth2: Secret = {
       disableWorkflowProviderSelection: false,
       isBrowsable: true,
       isReadOnly: false,
-      label: 'Scope',
-      name: 'Scope',
+      label: 'Credentials location',
+      name: 'CredentialsLocation',
       order: 4,
       supportedSyntaxes: ['Literal'],
       type: 'System.String',
-      uiHint: 'single-line',
+      uiHint: 'dropdown',
+      options: {
+        isFlagsEnum: false,
+        items: [
+          {
+            text: 'Authorization header',
+            value: 'header',
+          },
+          {
+            text: 'Form data',
+            value: 'form',
+          }
+        ]
+      }
     },
+    {
+      disableWorkflowProviderSelection: false,
+      isBrowsable: true,
+      isReadOnly: false,
+      label: 'Scope',
+      name: 'Scope',
+      order: 5,
+      supportedSyntaxes: ['Literal'],
+      type: 'System.String',
+      uiHint: 'single-line',
+    }
   ],
   type: 'OAuth2',
 };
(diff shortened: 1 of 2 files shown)


src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/oauth2.secret.ts
```diff
@@ -74,17 +74,41 @@ export const OAuth2: Secret = {
       type: 'System.String',
       uiHint: 'single-line',
     },
+    {
+      disableWorkflowProviderSelection: false,
+      isBrowsable: true,
+      isReadOnly: false,
+      label: 'Credentials location',
+      name: 'CredentialsLocation',
```
kevinchalet, line 82: Note: in case you'd want to use the standard term, it's officially called "token endpoint client authentication method".

src/designer/elsa-workflows-studio/src/modules/credential-manager/models/secrets/oauth2.secret.ts
```diff
@@ -74,17 +74,41 @@ export const OAuth2: Secret = {
...
+      order: 4,
+      supportedSyntaxes: ['Literal'],
+      type: 'System.String',
+      uiHint: 'dropdown',
+      options: {
+        isFlagsEnum: false,
+        items: [
+          {
+            text: 'Authorization header',
+            value: 'header',
```
kevinchalet, line 92: If you want to use the standard values, it's `client_secret_basic` for that.

src/modules/secrets/Elsa.Secrets.Http/Services/OAuth2TokenService.cs
```diff
@@ -67,6 +70,10 @@ public async Task<TokenResponse> GetToken(Secret secret, string? authCode, strin
 		{
 			var httpClient = _httpClientFactory.CreateClient(nameof(OAuth2TokenService));
 			httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+			if (credentialsLocation == "header")
+			{
+				httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", Base64Encode(clientId + ":" + clientSecret));
```
kevinchalet, line 75: Note: while many servers implement it this way, your encoding mechanism is not standard: you must formURL-encode the client identifier and secret before base64-encoding them (i.e do `Uri.EscapeDataString(value).Replace("%20", "+")` on each value).

Depending on the OAuth 2.0 services you want to support, you may want to make that configurable.

tanelkuhi: I added escaping, and since it is standard, and seems to be working with google, I wouldn't make it configurable. That secret modal is getting a bit bloated already, so if someone requires it, they can expand upon this feature.
About your second comment, I guess the one who setups the credentials, would just choose the `client_secret_post` option and not set secret, wouldn't they?

sfmskywalker (approved):

Issue #3984, opened 2023-05-01
HTTP Endpoint Authorize setting is not taken into effect


sfmskywalker:
## Repro steps

1. Create new workflow definition
2. Create HttpEndpoint and HttpResponse (simple as can be)
3. Do NOT set Authorize checkbox to true. Only provide Path and CanStartWorkflow
4. Write silly response: {status: "OK"}
5. PUBLISH
6. Execute Workflow through Postman WITHOUT bearer token.
7. Go back to workflow definition, check Authorize checkbox
8. PUBLISH
9. Execute Workflow through Postman WITHOUT bearer token
10. You are still able to execute the workflow

## Symptoms

The Authorize Checkbox in the trigger seems not to be updated in the DB. When we export this definition, we actually see the Authorize value of the HttpEndpoint is set to True. In the UI, it is set to true, but still, it does not trigger the IHttpEndpoitnAuthorizationHandler

In order to work around this, and make it persisted to the DB (at least we suspect that is the issue), perform the following steps:

1. Go back to WorkflowDefinition
2. Add a Policy: set input type to JavaScript and write return "default-policy"
3. PUBLISH
4. Execute through postman WITHOUTH bearer token. Expected response: 404
5. Go back to Workflow definition, remove the value in Policy input field, and set the input type of the form input field back to Default (so that it is no longer on JavaScript)
6. PUBLISH
7. Execute through postman WITHOUT bearer token. Expected response: 401
8. Now add a JWT Bearer Token and execute Workflow. Expected response: {status: "OK"}

PR #3985, opened 2023-05-01
Fix HTTP triggers not being updated

sfmskywalker:
This PR fixes an issue where triggers would not be updated under certain conditions.

In the case of HTTP Endpoint, the issue was that the `Authorize` field is excluded from the Hash computation (as it should) - but because trigger comparison relied on this hash, changing only the Authorize value would not change the hash, and therefore the trigger would not be replaced.


sfmskywalker:
Fixes #3984

PR #3995, opened 2023-05-03
Added option for override existing published workflow

MariusVuscanNx:
We wanted to use the workflow definition publisher in our custom workflow definition provider implementation.

Knowing that the workflow definition provider is executed on the application startup, using the workflow definition publisher in its current state will result in a new version of every workflow every time we restart the application.

This PR solves that by enriching the workflow definition publisher with an extra option to avoid the creation of a new version.

### DIFF
--- a/src/bundles/Elsa.AllInOne.Web/Program.cs
+++ b/src/bundles/Elsa.AllInOne.Web/Program.cs
@@ -4,6 +4,7 @@ using Elsa.EntityFrameworkCore.Modules.Management;
 using Elsa.EntityFrameworkCore.Modules.Runtime;
 using Elsa.Webhooks.Extensions;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Cors.Infrastructure;
 
 var builder = WebApplication.CreateBuilder(args);
 builder.WebHost.UseStaticWebAssets();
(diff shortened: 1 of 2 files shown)


sfmskywalker:
I'd like to have a better understanding of your use case of publishing workflows during application startup. Is this because you have JSON files that you need to be part of the system? In which case, have you considered using the blob storage provider (`FluentStorageWorkflowDefinitionProvider`) that can read JSON files from e.g. a local folder, blob storage, etc.?

MariusVuscanNx:
Yes, we have workflows as json files in the blob storage and we retrieve those using our custom made workflow provider. 
I checked the FluentStorageWorkflowDefinitionProvider and it doesn't fully apply in our case as:
- The FluentStorageWorkflowDefinitionProvider expects that each workflow is in a separate file. In our case all the workflows are in the same file.
- We do some extra processing of the json file (compression), so in the provider we need to decompress the file at the beginning.

sfmskywalker:
Gotcha. have you considered implementing a custom `IWorkflowDefinitionProvider` that performs the extra processing you need? In doing so, you wouldn't have to explicitly import the workflow definitions anymore.

MariusVuscanNx:
Yes, it was implemented using a custom IWorkflowDefinitionProvider.

I checked FluentStorageWorkflowDefinitionProvider code and we found out that it has issues with importing sub-workflows and workflows that contain sub-workflows. First, the subworkflows, were imported without the inputs, outputs and the IsUsableAsActivty properties set. We tried setting those but, we still have issues (When you open the main workflow, you get Not found activity). I created an issue for this: https://github.com/elsa-workflows/elsa-core/issues/3997

So, in that's way we cannot remove the code for manually importing the workflow. Therefore, our options would be to fix this issue with the sub-workflows or manually import the workflows in the provider.

sfmskywalker:
I see, thanks for the clarification!

This makes me realize that we need to fix this for the workflow provider mechanism, regardless of the importer changes you are suggesting via this PR.

I created a separate issue for that: https://github.com/elsa-workflows/elsa-core/issues/4011

For this PR, I will take another look soon. If this is somehow blocking you, please use the importer code in your own project until the workflow importer and/or the provider issue is handled.

PR #4033, opened 2023-05-12
Handle exceptions generated from ActivityTypeProviders

mohdali:
If an ActivityType provider generates an exception it will cause the /activities endpoint to fail and the missing triggers will cause the runtime to stop.

This attempts to handle those exceptions that may be caused by external dependencies and improve the engine robustness.


sfmskywalker (approved):
Thanks!

PR #4064, opened 2023-05-24
Allow the user to select the start activty when calling through the api

MariusVuscanNx:
Allow the user to select which activity should start with from the ones marked with 'Can start workflow'. This is useful when calling the workflow through the /execute or /dispatch endpoint.


sfmskywalker (approved):
Thank you!

PR #4105, opened 2023-06-05
Changed consumer/message processing to synchronized mode - former async (#4095)

### LINKED ISSUE #4095: Elsa 2: Kafka Activity module is preparing consumed message in an asynchronous way
The Elsa.Activities.Kafka.Helpers.Consumer class is using the observer pattern to notificate about received messages from Kafka.
The "notification" process is working asynchronous (fire and forget method) which is causing an issue, if there is a lag in the kafka queue or a bulk processing is intended.

In other words; the consumer is trying to preparing all (outstanding) queue messages in parallel and this can cause in extreme cases the exhausting of all pooled database connections.

To prevent this kind of issues, the process between receiving and preparing should be in an await mode, and the process should stop/retry if preparation of a message will fail. Otherwise, the message which is not processed will be lost.

dao74:
resolves #4095 

The asynchronous processing of incoming message can exhaust the database connection pool if there is a (big) lag in the kafka queue.
On other hand, the ACK of each message should perform after workflow is successfully created (or retried when the process fails) to not lose the kafka message.
This PR is fixing the synchronize processing and ACK issue.

Note:
Changing to synchronize will slowdown the whole process for larger amount of kafka messages. Therefore, the consumer should be increased - in the same count like the partition size of the topic.
Therefore the partition size should be configurable in the Kafka Activity.


dao74:
@dotnet-policy-service agree

sfmskywalker (approved):

dwoldo:
This also resolves #4192

Snotax:
@dao74  Where did you even encounter your original error, we couldn't reproduce this?

dao74:
Hello all, are we talking about side affects?

We are good with this fix, I don't remember the details of our problem but
would like to know what is wrong and why you need the revert.

Snotax:
We are experience an unusable increase in time used to handle incoming messages. Like multiple seconds per message.

Snotax:
We are experiencing an unusable increase in time used to handle incoming messages. Like multiple seconds per message.

dao74:
The reason for the slowness is not our fix; please
check TriggerWorkflowsAsync method in the Worker class.

You will find the following code which throttles the process flow:

*await Task.Delay(_delay, cancellationToken);*

We were not sure why it is necessary to have a sleep between the messages
therefore we didn' touch this part.
But in our internal tests and (load tests) we commented it out, with
satisfied results.

Please check whether we can waive the Task.Delay line.

Am Mi., 11. Okt. 2023 um 19:37 Uhr schrieb Yannick Laubscher <
***@***.***>:

Snotax:
I know this line. But the increase in consumation time is significantly higher with your change.

dao74:
> I know this line. But the increase in consumation time is significantly higher with your change.

Of course, process time might be increased because of the sync mode (instead of async) but I am not sure how significant - we didn't compared or made a benchmark. The async processing brings other problems, therefore - in my opinion - it must be stay in sync mode.

We made the performance tests (bulk processing) without the delay line and we were satisfied with the result. You should do the same before making a decision.

Snotax:
Ok, we will do some more performance testing next week. We'll keep you updated.

PR #4133, opened 2023-06-13
Fix - subscribers now Mqtt broker restart proof

mircotamburini:
After test we made this feature to be proof after mqtt broker restart

### DIFF
--- a/src/activities/Elsa.Activities.Mqtt/Services/MqttClientWrapper.cs
+++ b/src/activities/Elsa.Activities.Mqtt/Services/MqttClientWrapper.cs
@@ -25,15 +25,15 @@ namespace Elsa.Activities.Mqtt.Services
             _logger = logger;
         }
 
-        static bool subscribed = false;
-
-        public async Task SubscribeAsync(string topic)
+        
+        private async Task SubscribeAsync(string topic, Func<MqttApplicationMessage, Task> handler)
         {
             if (!Client.IsConnected)
             {
                 var opt = Options.GenerateMqttClientOptions();
                 Client.ConnectAsync(opt).Wait(); //Sync
-
+                _messageHandler = handler;
+                await Client.SubscribeAsync(topic, Options.QualityOfService);
                 Client.ApplicationMessageReceivedAsync += async e =>
                 {
                     if (_messageHandler != null)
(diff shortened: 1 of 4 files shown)


src/activities/Elsa.Activities.Mqtt/Services/MqttClientWrapper.cs
```diff
@@ -25,25 +25,22 @@ public MqttClientWrapper(IMqttClient client, Options.MqttClientOptions options,
...
+        
+        private async Task SubscribeAsync(string topic, Func<MqttApplicationMessage, Task> handler)
         {
             if (!Client.IsConnected)
             {
                 var opt = Options.GenerateMqttClientOptions();
                 Client.ConnectAsync(opt).Wait(); //Sync
-
+                _messageHandler = handler;
+                await Client.SubscribeAsync(topic, Options.QualityOfService);
```
sfmskywalker, line 36: Isn't it safer to subscribe to the topic **after** setting up the event handler, instead of the other way around?
In theory, it is possible that the client immediately receives an event after subscribing, at which point the next line where the event handler is assigned, hasn't executed yet. Admittedly, this seems like something that wouldn't actually happen in practice, but if it doesn't hurt, maybe better to first attach the handler, then subscribe to the topic.

src/activities/Elsa.Activities.Mqtt/Services/Worker.cs
```diff
@@ -30,6 +30,11 @@ public class Worker
             _receiverClient.SetMessageHandlerAsync(OnMessageReceived);
         }
 
+        public void Ping()
+        {
+            _receiverClient.SetMessageHandlerAsync(OnMessageReceived);
+        }
+
```
sfmskywalker, line 37: Out of curiosity, can you explain how this works? The method is called "ping", but it only sets a message handler. Does this mean that the receiver client is going to send some "ping" message under the hood?

mircotamburini: Yeah i had to adopt a compromise, because i have to leave opened subscribing (before i have issue with retain topics)  so instead to open and close, now i test if  Client.IsConnected

src/activities/Elsa.Activities.Mqtt/Services/MqttClientWrapper.cs
mircotamburini: you are right , im scared to leave handler binded but need to se handler before

PR #4149, opened 2023-06-19
UnscheduleCron triggers

sNakiex:
The Cron triggers were not unscheduled, this resulted in multiple executions upon the publishing and unpublishing of workflows.


sNakiex:
@dotnet-policy-service agree

sfmskywalker (approved):
Great catch, thank you!

sNakiex:
> small nit:
> 
> if you're calling tolist at the end (`filteredTriggers`), there's no need to call tolist in the intermediate collections (`timerTriggers` and `startAtTriggers`) since you only need to iterate through them once before collecting them in the final list.

Is it possible to enhance efficiency by completely removing the ToList operation? By eliminating it, the triggerList object could also be omitted. Unless I'm mistaken, this change should have no impact on the functionality.

PR #4158, opened 2023-06-22
Fixed the issue with setVariable execution in javascript activity

gurkanguran:


sNakiex:
Ran into this bug yesterday and tried your fix and it works!

sNakiex:
Is there a specific reason why the getVariable method doesn't use GetVariableInScope? I noticed that the typed version does use it. Wouldn't it make sense for both of them to access variables in the same way? This way, both methods would work exactly the same when retrieving a variable, making things consistent and easier to understand.

sfmskywalker (approved):

PR #4160, opened 2023-06-23
Cleanup schedulers

sNakiex:
As discussed in pull request #4149 The ToList and list variable are unnecessary.

So this pull request cleans up both DefaultTriggerScheduler and DefaultBookmarkScheduler by removing both the variable and the ToList actions.


sfmskywalker (approved):
Thank you!

PR #4199, opened 2023-07-05
Allow saving the Workflow Version Name for the instance list endpoint

brunomartinspro:
Currently, when using the endpoint /v1/workflow-instances the property "Name" is always null.

Added a mapping to the Workflow Factory that allows saving the "Name" of the executed workflow definition for retrieval later in the /v1/workflow-instances endpoint.

Example of the current output of the endpoint /v1/workflow-instances endpoint:
image

### DIFF
--- a/src/core/Elsa.Core/Services/Workflows/WorkflowFactory.cs
+++ b/src/core/Elsa.Core/Services/Workflows/WorkflowFactory.cs
@@ -28,6 +28,7 @@ namespace Elsa.Services.Workflows
             var workflowInstance = new WorkflowInstance
             {
                 Id = _idGenerator.Generate(),
+                Name = workflowBlueprint.Name,
                 DefinitionId = workflowBlueprint.Id,
                 TenantId = tenantId ?? workflowBlueprint.TenantId,
                 Version = workflowBlueprint.Version,


sfmskywalker:
Although I'm not necessarily against it, I am mildly concerned that it might give the impression that it is supposed to be set to the workflow blueprint name. 

The 'Name' property of the workflow instance was meant to be used as a free field for workflows that want to put in e.g. customer names, or document titles, or whatever else makes sense for a given workflow.

I do realize that by having the workflow instance `Name` property be literally called the same thing as on the workflow blueprint (also being `Name`) doesn't help my case. 😅 

Out of curiosity, do you need the Name property to be set to the workflow blueprint's name, or perhaps you (understandably!) concluded it must be a bug that the Name field wis being null?

Perhaps, after all, it makes sense for the Name property to default to the workflow blueprint's name, but I would love to hear your reflections on this.

brunomartinspro:
> Although I'm not necessarily against it, I am mildly concerned that it might give the impression that it is supposed to be set to the workflow blueprint name.
> 
> The 'Name' property of the workflow instance was meant to be used as a free field for workflows that want to put in e.g. customer names, or document titles, or whatever else makes sense for a given workflow.
> 
> I do realize that by having the workflow instance `Name` property be literally called the same thing as on the workflow blueprint (also being `Name`) doesn't help my case. 😅
> 
> Out of curiosity, do you need the Name property to be set to the workflow blueprint's name, or perhaps you (understandably!) concluded it must be a bug that the Name field wis being null?
> 
> Perhaps, after all, it makes sense for the Name property to default to the workflow blueprint's name, but I would love to hear your reflections on this.

The way i thought about this (and i can be completely missing something) for my business case was:

1. I create a workflow definition with a "Name" 
2. I execute this workflow definition and a Instance is generated 

I then show the instances using pagination in another screen using the endpoint /v1/workflow-instances.

i concluded it was a bug because the "Name" was always null, so i assumed (and i can be wrong) that the Instance record should  always have the Definition Name (that matches the executed version) so i can avoid doing another call to the /v1/workflow-registry.

Because i would have to paginate over the instances (/v1/workflow-instances) and then loop the result of the /v1/workflow-registry to populate the "Name" property for all the records to show it in a table, and this depending on the number of records can affect performance.

Should i create instead a "DefinitionName" property?

sfmskywalker:
I understand. The way the Elsa dashboard achieves displaying the workflow instance's definition name with great performance is by doing it in the following steps:

1. Load a page of workflow instances.
2. Project their definition IDs into a distinct list.
3. Select all WorkflowDefinitionSummary's matching the list of definition IDs.

This causes just two DB queries. Granted, that's not as optimal as just doing one DB call, but has the added benefit of including potentially more information about a workflow instance that comes from its definition.

That being said, if you do want to include the workflow definition name is part of the workflow instance's Name property, there are two ways to achieve this:

1. Implement a notification handler for the `WorkflowExecuting` notification. The notification object provides access to the blueprint and the instance, so you could set the workflow instance's Name property from this handler.
2. Use the Set Name activity to set the workflow instance name when the workflow starts. You can set the Value property to a JS or Liquid expression. For example: `workflowExecutionContext.WorkflowBlueprint.Name`

PR #4211, opened 2023-07-10
Secrets updates

johnwc:


johnwc:
@sfmskywalker We also have the following activity that we were going to contribute if you are interested. It allows you to create a workflow that the flow can be controlled based on what environment it is running in. Within the startup of the server, you call `elsa.SetEnvironment(EnvironmentNames.Production)` that sets the server's environment it is running in.

A good example for this, you can have a workflow with a cron schedule that you only want to run in Production or UAT. But you don't want to maintain two different workflows for each environment. The other start path can be a http endpoint to allow manually starting the workflow in the other environments.

image

johnwc:
Are you interested in the environment activity?

sfmskywalker:
I completely overlooked your message about the Environment activity, but that looks great! I would love for it to be part of the core set of activities 👍🏻

PR #4223, opened 2023-07-12
New workflow definitions backup/restore

johnwc:
Allows backing up and restoring entire workflow definitions list. Upon restoring determines if workflow definition already exists, publishing a new version if it does and is published or creating a new one if it does not.

### DIFF
--- a/src/designer/elsa-workflows-studio/src/components.d.ts
+++ b/src/designer/elsa-workflows-studio/src/components.d.ts
@@ -264,6 +264,7 @@ export namespace Components {
         "basePath": string;
         "culture": string;
         "history": RouterHistory;
+        "serverUrl": string;
     }
     interface ElsaStudioWorkflowInstancesList {
         "culture": string;
@@ -348,6 +349,7 @@ export namespace Components {
         "basePath": string;
         "culture": string;
         "history"?: RouterHistory;
+        "loadWorkflowDefinitions": () => Promise<void>;
         "serverUrl": string;
     }
     interface ElsaWorkflowFaultInformation {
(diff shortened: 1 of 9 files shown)


johnwc:
@sfmskywalker are you good with this addition? We desperately needed something like this to move our proven-out workflows from dev environment into prod environment.

sfmskywalker:
Probably yes, but I need a little time to check it out. I'll try tomorrow.

sfmskywalker (approved):
This is great, thank you!

johnwc:
@sfmskywalker when do you think it will be merged?

sfmskywalker:
Should have already merged it, but sometimes I move too fast ;)

johnwc:
@sfmskywalker So, after using this in our dev and uat environment, I am seeing an odd issue. I restored the backup into a server with existing workflows from previous restore. (Backed up dev, restored to uat, published a few in uat, edited dev workflows, backed up dev again, restored in uat again.)

This is now what I am seeing when viewing the list of definitions. Notice there is only 9 in the browser listed, but 15 in the api response. It seems all the workflows that were published during the restore are duplicated in the api response. With version 1 being `isPublished: true, isLatest: false` and version 2 `isPublished: false, isLatest: true`. What do you think is causing this?

image

johnwc:
@sfmskywalker any insight on this?

sfmskywalker:
First thing I would do is take a look at the WorkflowDefinitions DB table and inspect the IsLatest and IsPublished columns of a set of the same DefinitionID. Probably you will see that multiple records have IsLatest and or IsPublished are set to true. In which case I suspect that there is an issue with the Restore functionality. You should be able to reproduce this on a local machine with a debugger attached. The Publish functionality implements logic to ensure that there’s only one IsPublished and one IsLatest record within a group of a given DefinitionID. Perhaps this is missing in the restore functionality.

PR #4230, opened 2023-07-12
Add option to check custom header when receiving http endpoint request

tanelkuhi:
For HttpEndpoint activity we have to have the previous authorization work in parallel with the now new check for plain header (ApiKey or something like that). I feel like there could be a better solution, so I am open to suggestions.


sfmskywalker:
A more elegant approach might be to associate "http filters" with a given HTTP request, making it more modular and configurable. This could be achieved by examining how API gateways manage request filtering. However, I'm comfortable accepting this PR as it stands, since it introduces optional features that will be useful for others who may be seeking the same solution.

PR #4343, opened 2023-08-21
Change RunTask Payload from ActivityOutput to ActivityInput

SydoxX:
The `Payload` property of the `Elsa.Activities.Conductor.RunTask` is annotated as an `ActivityOutput`. This seems to be wrong as it is used in the `OnExecuteAsync` method, where it is always set to `null` with this annotation. Annotating it to be an `ActivityInput` solves that problem.


SydoxX:
@microsoft-github-policy-service agree company="Technische Informationssysteme GmbH"

sfmskywalker (approved):
Great catch, thanks for the PR!

PR #4363, opened 2023-08-25
Enrich trigger validation to take into consideration path and method …

MariusVuscanNx:
…uniquness

Before, there was a piece of code that ensured that the path was unique across workflows/triggers. That code was improved by taking into consideration the method. Now, you can have two workflows/triggers with the same path as long as they have different methods.


sfmskywalker (approved):
Fantastic, thanks @MariusVuscanNx !

PR #4370, opened 2023-08-29
Fixed IsNameUnique for InMemory WorkflowDefinition store

yspanj:
IWorkflowDefinitionStore GetIsNameUnique only inverted the result in the InMemoryStore, and it was inverted again in the endpoint. 

I changed the GetIsNameUnique to return false when the name already exists, and removed the invert in the endpoint. 

This should fix the issue seen in elsa-workflows/elsa-studio#29


sfmskywalker (approved):
Thanks!

PR #4398, opened 2023-09-05
Add PostgreSql dapper provider

herblinernexxbizz:


src/modules/Elsa.Dapper/Dialects/PostgreSqlDialect.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Elsa.Dapper.Dialects;
+
+/// <summary>
+/// Represents a SQL Server dialect.
+/// </summary>
+public class PostgreSqlDialect : SqlDialectBase
+{
+
+    /// <inheritdoc />
+    public virtual string Upsert(string table, string primaryKeyField, string[] fields)
```
sfmskywalker, line 13: A method with the same name also exists on the base class. Can we override that one instead?

sfmskywalker (changes requested):
Looks good! I think that all we need to do is override the `Upsert` method from the base class.

sfmskywalker (approved):

Issue #4415, opened 2023-09-08
Server unable to handle ANY requests when "Http Endpoint" Activity Path contains "//"


polferov:
# The Issue

Disclaimer: I use path and route interchangeably, as the HttpEndpoint activity calls it path and ASP.NET Core calls it route.

When creating an HttpEndpoint activity and entering a route that contains two consecutive slashes ("//"), e.g. when accidentally pasting (or getting via js) an absolute url (e.g. "https://abc.xyz"), an exception will be thrown on ANY request to the server.

The exception(s) are

```
Microsoft.AspNetCore.Routing.Patterns.RoutePatternException: The route template separator character '/' cannot appear consecutively. It must be separated by either a parameter or a literal value.
   at Microsoft.AspNetCore.Routing.Patterns.RoutePatternParser.Parse(String pattern)
   at Microsoft.AspNetCore.Routing.Template.TemplateParser.Parse(String routeTemplate)
```

and

```
System.ArgumentException: The route template separator character '/' cannot appear consecutively. It must be separated by either a parameter or a literal value. (Parameter 'routeTemplate')
 ---> Microsoft.AspNetCore.Routing.Patterns.RoutePatternException: The route template separator character '/' cannot appear consecutively. It must be separated by either a parameter or a literal value.
   at Microsoft.AspNetCore.Routing.Patterns.RoutePatternParser.Parse(String pattern)
   at Microsoft.AspNetCore.Routing.Template.TemplateParser.Parse(String routeTemplate)
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Routing.Template.TemplateParser.Parse(String routeTemplate)
   at Elsa.Activities.Http.Services.RouteMatcher.Match(String routeTemplate, String requestPath)
   at Elsa.Activities.Http.Middleware.HttpEndpointMiddleware.<>c__DisplayClass2_0.<InvokeAsync>b__0(String route)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
[...]

PR #4416, opened 2023-09-08
fix issues with double slashes

polferov:
#4415 

Fix issues described in issue using proposed solution

### DIFF
--- a/src/activities/Elsa.Activities.Http/Extensions/RouteTableExtensions.cs
+++ b/src/activities/Elsa.Activities.Http/Extensions/RouteTableExtensions.cs
@@ -15,13 +15,15 @@ public static class RouteTableExtensions
     
     public static void AddRoutes(this IRouteTable routeTable, IEnumerable<Trigger> triggers)
     {
-        var paths = Filter(triggers).Select(Deserialize).Select(x => x.Path).ToList();
+        var paths = Filter(triggers).Select(Deserialize).Select(x => x.Path)
+            .Where(x => !x.Contains("//")).ToList();
         routeTable.AddRange(paths);
     }
 
     public static void AddRoutes(this IRouteTable routeTable, IEnumerable<Bookmark> bookmarks)
     {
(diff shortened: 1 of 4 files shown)


polferov:
@dotnet-policy-service agree company="Fischer & Consultants"

sfmskywalker:
@polferov Apologies, I completely missed the fact that you already submitted a PR.

sfmskywalker:
Actually, never mind my last comment - I fixed it for V3, but you're fixing it for V2 👍🏻

src/activities/Elsa.Activities.Http/Extensions/RouteTableExtensions.cs
```diff
@@ -15,13 +15,15 @@ public static class RouteTableExtensions
...
+            .Where(x => !x.Contains("//")).ToList();
         routeTable.AddRange(paths);
     }
 
     public static void AddRoutes(this IRouteTable routeTable, IEnumerable<Bookmark> bookmarks)
     {
-        var paths = Filter(bookmarks).Select(Deserialize).Select(x => x.Path).ToList();
+        var paths = Filter(bookmarks).Select(Deserialize).Select(x => x.Path)
+            .Where(x => !x.Contains("//")).ToList();
         routeTable.AddRange(paths);
```
sfmskywalker, line 27: Although this will certainly work, it might be even better to do the filtering at the lowest level possible, which is inside the `Add` method of the `RouteTable` class itself - that way, all code paths leading into there will have this filtering logic applied. For example:

```csharp
public void Add(string path)
{
    if (path.Contains("//"))
    {
        _logger.LogWarning("Path cannot contain double slashes. Ignoring path: {Path}", path);
        return;
    }

    Routes.TryAdd(path, path);
}
```

polferov: I see. I falsely assumed that the (I)RouteTable was coming from ASP.NET Core itself. But since it is part of else, it makes sense to do it there.

sfmskywalker (approved):

PR #4428, opened 2023-09-12
Hangfire job names

johnwc:
* Updated hangfire to latest version.
* Applied job names for hangfire jobs.

### DIFF
--- a/src/core/Elsa.Abstractions/Extensions/CompositeActivityBlueprintExtensions.cs
+++ b/src/core/Elsa.Abstractions/Extensions/CompositeActivityBlueprintExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
 using Elsa.Models;
 using Elsa.Services.Models;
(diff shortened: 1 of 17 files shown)


src/core/Elsa.Abstractions/Extensions/CompositeActivityBlueprintExtensions.cs
```diff
@@ -9,6 +9,19 @@ public static class CompositeActivityBlueprintExtensions
...
+            {
+                var activity = workflowBlueprint?.GetActivity(activityId);
+                if (activity is not null)
+                    return $"{workflowBlueprint?.DisplayName ?? workflowBlueprint?.Name ?? workflowBlueprint?.Id}->{activity?.DisplayName ?? activity?.Name ?? activity?.Id}";
+            }
+            if (workflowBlueprint is not null)
+                return $"{workflowBlueprint?.DisplayName ?? workflowBlueprint?.Name ?? workflowBlueprint?.Id}";
+            return null;
+        }
+
```
sfmskywalker, line 24: Since this seems to be specific to generating a job name, perhaps we should move this extension method close to where it's used? Maybe somewhere in Elsa.Activities.Temporal.Common? Or is this used more broadly beyond the domain of scheduling and jobs?

johnwc: This is for creating a JobName for a blueprint, anything can use it to generate a job name. if I move it to the Elsa.Activities.Temporal.Common project, that will require the Elsa.Server.Hangfire project to have a dependency on a Elsa.Activities project. And that may not be desired by some.

johnwc: @sfmskywalker I rolled a null reference bugfix in this as well.

johnwc: When can this be merged?

sfmskywalker (approved):

src/core/Elsa.Abstractions/Extensions/CompositeActivityBlueprintExtensions.cs
sfmskywalker: Will merge now.

PR #4436, opened 2023-09-13
Incidents

sfmskywalker:
This PR changes the way faults are handled, by introducing a new concept called **Incidents**.

When a fault occurs, we regard it as an incident. Before this PR, a workflow would automatically enter the Faulted state, preventing it from continuing.

With this PR, this behavior becomes configurable through **Incident strategies**. Out of the box, there are two strategies provided:

- Fault Strategy - encodes the current behavior where the workflow enters the Faulted state and stops executing.
- Continue with Incidents strategy - encodes a new behavior where the workflow continues running.

The Elsa Studio UI has been updated with the following:
- The user can configure the desired incident strategy at the workflow definition level.
- The Workflow Instances list view displays the number of incidents that have occurred so far.


src/modules/Elsa.Http/Handlers/DefaultHttpEndpointWorkflowFaultHandler.cs
```diff
@@ -27,17 +27,12 @@ public virtual async ValueTask HandleAsync(HttpEndpointFaultedWorkflowContext co
     {
         var httpContext = context.HttpContext;
         var workflowState = context.WorkflowState;
-        var fault = workflowState.Fault;
+        var incidents = workflowState.Incidents;
```
MariusVuscanNx, line 30: The incidents variable does not seem to be used.

MariusVuscanNx (approved):

PR #4480, opened 2023-09-25
Missing WorkflowPassCompleted enum element in WorkflowPersistenceBehavior.cs

NightinGale997:
In Elsa.Abstractions there is enum for serialization of existing workflows in blob storage. WorkflowPassCompleted is missing there.

### DIFF
--- a/src/core/Elsa.Core/Handlers/PersistWorkflow.cs
+++ b/src/core/Elsa.Core/Handlers/PersistWorkflow.cs
@@ -27,7 +27,7 @@ namespace Elsa.Handlers
         {
             var behavior = notification.WorkflowExecutionContext.WorkflowBlueprint.PersistenceBehavior;
 
-            if (behavior == WorkflowPersistenceBehavior.ActivityExecuted || notification.ActivityExecutionContext.ActivityBlueprint.PersistWorkflow)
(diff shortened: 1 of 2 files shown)


NightinGale997:
@dotnet-policy-service agree

sfmskywalker:
Thanks for the PR! I'm wondering if it's enough to just add the enum? Should there be some logic somewhere that checks for this value and take appropriate action?

NightinGale997:
I found only one place where it would make sense to put logic behind it, last commit.

src/core/Elsa.Core/Handlers/PersistWorkflow.cs
```diff
@@ -27,7 +27,7 @@ public async Task Handle(WorkflowExecutionPassCompleted notification, Cancellati
         {
             var behavior = notification.WorkflowExecutionContext.WorkflowBlueprint.PersistenceBehavior;
 
-            if (behavior == WorkflowPersistenceBehavior.ActivityExecuted || notification.ActivityExecutionContext.ActivityBlueprint.PersistWorkflow)
+            if (behavior == WorkflowPersistenceBehavior.ActivityExecuted || behavior == WorkflowPersistenceBehavior.WorkflowPassCompleted || notification.ActivityExecutionContext.ActivityBlueprint.PersistWorkflow)
```
sfmskywalker, line 30: Aha. So, I think this may have been the reason I removed `WorkflowPassCompleted` in the past - it has effectively the same result as `ActivityExecuted`.

That being said, let's keep your changes in case we want to actually make a difference between the two behaviors.

sfmskywalker (approved):

PR #4521, opened 2023-10-11
3537 Switch to DefaultContractResolver

4lexKislitsyn:
Switch from `CamelCasePropertyNamesContractResolver` to `DefaultContractResolver`. `CamelCasePropertyNamesContractResolver` has a shared cache for contracts between instances. If any Dictionary<string, string> is previously serialized with settings to process dictionary keys, `ActivityDefinitionProperty.Expressions` will also be serialized with dictionary key processing. It's better to use DefaultContractResolver, which has a cache for each instance.


sfmskywalker (approved):

sfmskywalker:
Thanks for having figured that out, I had no idea about this.

PR #4522, opened 2023-10-11
[Kafka] Added possibility to add DefaultConnectionString

Snotax:
Added possibility to add DefaultConnectionString if connection string in trigger/activity is empty

### DIFF
--- a/src/activities/Elsa.Activities.Kafka/Activities/SendKafkaMessage/SendKafkaMessage.cs
+++ b/src/activities/Elsa.Activities.Kafka/Activities/SendKafkaMessage/SendKafkaMessage.cs
@@ -1,6 +1,3 @@
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using Confluent.Kafka;
 using Elsa.Activities.Kafka.Configuration;
 using Elsa.Activities.Kafka.Helpers;
 using Elsa.Activities.Kafka.Services;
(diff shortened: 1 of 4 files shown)


src/activities/Elsa.Activities.Kafka/Activities/SendKafkaMessage/SendKafkaMessage.cs
```diff
@@ -22,10 +22,12 @@ namespace Elsa.Activities.Kafka.Activities.SendKafkaMessage
     public class SendKafkaMessage : Activity
     {
         private readonly IMessageSenderClientFactory _messageSenderClientFactory;
+        private readonly KafkaOptions _kafkaOptions;
 
-        public SendKafkaMessage(IMessageSenderClientFactory messageSenderClientFactory)
+        public SendKafkaMessage(IMessageSenderClientFactory messageSenderClientFactory, KafkaOptions kafkaOptions)
```
sfmskywalker, line 27: Is the `KafkaOptions` type registered with DI somewhere? If not, I think this will cause an exception when trying to resolve the activity from the service container.

Snotax: KafkaOptions are already injected as a singleton and are currently used for default SASL credentials.

Issue #4299, opened 2023-08-08
Integrate Swagger for API Documentation in Elsa 3


sfmskywalker:
**Description**:
With our ongoing development in Elsa 3, which utilizes FastEndpoints, it has become increasingly essential to provide clear and concise API documentation for our end-users and developers. To achieve this, we should integrate Swagger to automatically generate this documentation.

**Acceptance Criteria**:
1. Swagger is successfully integrated with Elsa 3.
2. All existing API endpoints in Elsa 3 are documented within Swagger.
4. A dedicated URL (e.g., `/open-api`) is available for users to view the API documentation.
5. Documentation is clear, with accurate endpoint descriptions, request/response schemas, and example values.

**Additional Notes**:
1. FastEndpoints might have specific integration details or best practices for Swagger, so it's essential to ensure compatibility.

solutionmania:
I'm happy to have a go at this if you like?

sfmskywalker:
That would be fantastic!

PR #4525, opened 2023-10-11
Adds Swagger documentation generation and UI.

solutionmania:
This adds the Swagger documentation generation and Swagger UI middleware as a pair of extension methods (Issue #4299)

Once these are added to the application, you can view the API documentation by visiting: /swagger

At the moment I've hard-coded the version number at "3.0" - is there a better means of getting the version number of the API? The Elsa.Api.Common project's version number doesn't appear to be set.

Just submitting this PR at this stage to make sure this is heading in the right direction... I'll be adding some actual documentation to the API endpoints and updating this PR for further feedback / review.

Thanks!

### DIFF
--- a/src/bundles/Elsa.WorkflowServer.Web/Program.cs
+++ b/src/bundles/Elsa.WorkflowServer.Web/Program.cs
@@ -160,6 +160,8 @@ services
 
         // Initialize drop-ins.
         elsa.InstallDropIns(options => options.DropInRootDirectory = Path.Combine(Directory.GetCurrentDirectory(), "App_Data", "DropIns"));
+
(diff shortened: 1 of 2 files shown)


sfmskywalker (approved):

sfmskywalker:
This is great! 

> is there a better means of getting the version number of the API? The Elsa.Api.Common project's version number

The Elsa.Workflows.Api project contains an endpoint at /packages/version that reports the actual package version (which gets set during the build phase here on GitHub when building & publishing packages).

You could copy the same logic, but better would be to extract that logic to a shared service or utility class, and update both places (the /package/version endpoint and the swagger setup).

PR #4568, opened 2023-10-23
Add Authorization to File Http Response

sNakiex:
I needed to retrieve a file from a secure source, so I included the Authorization header. If there's a more elegant approach to achieve this please let me know.

### DIFF
--- a/src/modules/Elsa.Http/Activities/WriteFileHttpResponse.cs
+++ b/src/modules/Elsa.Http/Activities/WriteFileHttpResponse.cs
@@ -70,6 +70,16 @@ public class WriteFileHttpResponse : Activity
     [Input(Description = "The correlation ID of the download used to resume a download. If left empty, the x-download-id header will be used.")]
     public Input<string> DownloadCorrelationId { get; set; } = default!;
 
+    /// <summary>
+    /// The Authorization header value to send with the request.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


sfmskywalker:
Hi @sNakiex , I understand the reasoning, but I think that this addition exceeds the intended scope for the File HTTP Response activity. For your scenario, the recommendation is to first explicitly download the file using e.g. the SendHttpRequest activity, from which you can configure the Authorization header, and then use its output as the input for the File HTTP Response activity.

sNakiex:
Hello @sfmskywalker 
I had some issues with encoding using the approach you mention, that is why I opted to download the file directly.
Other bonus is that the document in question doesn't get saved to the instance which saves quite a lot of database storage.
But if this PR is unwanted feel free to reject. :-)

sfmskywalker:
I understand. So how about we fix the issues you mentioned instead.

1. Would you mind creating an issue for the encoding issue with steps to reproduce?
2. If the SendHttpRequest activity stores the downloaded document as part of the workflow instance, then that's definitely a problem, and a bug - it's not supposed to do that. Unless you're using a variable that is configured with "Workflow" storage, then it would do that. In that case, make sure to configure the variable to use "Memory" storage instead.

I will go ahead and close this PR, but I do want to give you a big Thank You for your contributions!

PR #4572, opened 2023-10-24
Kafka: Added Serialized Headers to WorkflowInput

Snotax:
Added serialized message headers to MessageReceivedInput --> WorkflowInput

### DIFF
--- a/src/activities/Elsa.Activities.Kafka/Models/MessageReceivedInput.cs
+++ b/src/activities/Elsa.Activities.Kafka/Models/MessageReceivedInput.cs
@@ -8,6 +8,8 @@ namespace Elsa.Activities.Kafka.Models
 {
     public class MessageReceivedInput
     {
+        public string? MessageHeaders { get; set; }
+
         public string? MessageString { get; set; }
 
         public byte[]? MessageBytes { get; set; }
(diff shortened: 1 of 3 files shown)


Snotax:
@sfmskywalker  Would be nice to get a release in the close future after this PR :D Thx

src/activities/Elsa.Activities.Kafka/Models/MessageReceivedInput.cs
```diff
@@ -8,6 +8,8 @@ namespace Elsa.Activities.Kafka.Models
 {
     public class MessageReceivedInput
     {
+        public string? MessageHeaders { get; set; }
```
sfmskywalker, line 11: Wouldn't it be more convenient to get the headers as a collection of headers instead of a serialized JSON string? Otherwise, the workflow activities first need to deserialize this, which also means they need to know what sort of JSON format to expect, like PascalCase, camelCase, etc.

Snotax: Sure, I'll try

Snotax: @sfmskywalker  I've changed it to Dictionary<string, string>.

sfmskywalker (approved):

PR #4607, opened 2023-11-07
V3 - Add to sample activity execution log pipeline and exception handler pipeline

b0rv3g4r4:
PR to add activity execution log and exception handling to pipeline

It can be confusing for people who rely on the example to enable workflow context.


src/samples/aspnet/Elsa.Samples.AspNet.WorkflowContexts/Program.cs
```diff
@@ -37,13 +37,18 @@
                 .UsePersistentVariables()
                 .UseBookmarkPersistence()
                 .UseWorkflowExecutionLogPersistence()
```
terry-delph, line 39: These two lines are repeated. Is this intentional?

b0rv3g4r4: Repeated?

sfmskywalker (approved):

PR #4616, opened 2023-11-12
AssemblyPathActivityProvider

IbbaAlberto:
Provides activity descriptors based on a list of NON-ELSA activity types finded into the same directory of the assembly.

### DIFF
--- a/src/modules/Elsa.Workflows.Management/Providers/AssemblyPathActivityProvider.cs
+++ b/src/modules/Elsa.Workflows.Management/Providers/AssemblyPathActivityProvider.cs
@@ -1,0 +1,63 @@
+﻿using Elsa.Workflows.Core.Contracts;
+using Elsa.Workflows.Core.Models;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Elsa.Workflows.Management.Providers;
+
+/// <summary>
+/// Provides activity descriptors based on a list of NON-ELSA activity types finded into the same directory of the assembly.
+/// </summary>
+public class AssemblyPathActivityProvider : IActivityProvider
+{
+    private readonly IActivityDescriber _activityDescriber;
+
+    public AssemblyPathActivityProvider(IActivityDescriber activityDescriber)
+    {
+        _activityDescriber = activityDescriber;
+    }
+
+    /// <inheritdoc />
+    public async ValueTask<IEnumerable<ActivityDescriptor>> GetDescriptorsAsync(CancellationToken cancellationToken = default)
+    {
+        var activityTypes = GetActivityTypes();
+        var descriptors = await DescribeActivityTypesAsync(activityTypes, cancellationToken).ToListAsync(cancellationToken);
+        return descriptors;
+    }
+
+    private async IAsyncEnumerable<ActivityDescriptor> DescribeActivityTypesAsync(IEnumerable<Type> activityTypes, [EnumeratorCancellation] CancellationToken cancellationToken)
+    {
+        foreach (var activityType in activityTypes)
+        {
+            var descriptor = await _activityDescriber.DescribeActivityAsync(activityType, cancellationToken);
+            yield return descriptor;
+        }
+    }
+
+    private IEnumerable<Type> GetActivityTypes()
+    {
+        var currentAssembly = Assembly.GetExecutingAssembly();
+
+        // Get the directory of the current assembly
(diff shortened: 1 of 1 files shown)


IbbaAlberto:
I'm proposing an activity provider that scan the assembly directory to provides all activity types that doesn't come from Elsa DLL's.

IbbaAlberto:
@dotnet-policy-service agree


Il giorno dom 12 nov 2023 alle ore 08:39 dotnet-policy-service[bot] <
***@***.***> ha scritto:

IbbaAlberto:
> Is the idea here that you can add assembly DLL files to the bin folder without referencing them from your main project? If that's the case, then I think the Elsa.DropIns module already provides this functionality. It basically reads the configured "drop-in" folder for assemblies, loads them in, and installs any activities it finds with the system. It even supports monitoring the directory at runtime, allowing you to add assemblies while the application is running without requiring a restart.

No, the Idea is to have an ActivityProvider that provides all non-ELSA activities coming from each namespace referred, just using one ".AddActivityProvider<AssemblyPathActivityProvider>()" instead of multiple ".AddActivitiesFrom<...>()". The purpose is to use it in project where you have multiple custom activities that could come from different namespaces.
At the moment I had to limit it to non-ELSA activities, but in the future I'd like to extend it to scan all project's namespaces and provide all, ELSA and non-ELSA, activities.

sfmskywalker:
I see, so this is a provider that provides all activities automatically without having to explicitly do e.g. `elsa.AddActivitiesFrom<TMarker>()`, is that correct?

mathijs-dumon:
Yes that is correct.

There is currently a workaround (filtering out assemblies starting with Elsa) because the activity registry does not like it when you try to register the same activity twice. Maybe a fix is warranted there, that would remove the need for this check.

For Context, Alberto and I are both working on the same project, trying to use Elsa as a replacement for Workflow Foundation.

sfmskywalker:
I see. Thanks for confirming. With regards to attempting to register the same activity twice, my concern is that this might lead to covering up a bug where there are two different activities but with the same name (however slim that change is due to name spacing). What we could do, however, is to add another method called e.g. TryAdd and or TryReplace. The former would do nothing in case the activity already exists, the latter would replace the existing activity with the new one (which potentially is the same activity anyway given your use case with this new activity provider).

Please let me know if you have any thoughts on this.

src/modules/Elsa.Workflows.Management/Providers/AssemblyPathActivityProvider.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Elsa.Workflows.Core.Contracts;
+using Elsa.Workflows.Core.Models;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Elsa.Workflows.Management.Providers;
+
+/// <summary>
+/// Provides activity descriptors based on a list of NON-ELSA activity types finded into the same directory of the assembly.
```
sfmskywalker, line 9: Based on the discussion, is it correct that this filtering out of Elsa activities is t prevent them from being added twice, which would lead into an exception? If yes, then let's update the service that invokes these providers by simply ignoring already-registered activities.

src/modules/Elsa.Workflows.Management/Providers/AssemblyPathActivityProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+    private IEnumerable<Type> GetActivityTypes()
+    {
+        var currentAssembly = Assembly.GetExecutingAssembly();
+
+        // Get the directory of the current assembly
+        var path = Path.GetDirectoryName(currentAssembly.Location);
+
+        // Get all the assembly files in the directory
+        var assemblyFiles = System.IO.Directory.GetFiles(path, "*.dll")
+            .Where(f => !f.Split('\\').ToList().Last().StartsWith("Elsa"));
```
sfmskywalker, line 46: Does this work for Linux-based systems? Would it be better to split on e.g. `Path. PathSeparator`?

sfmskywalker:
> What we could do, however, is to add another method called e.g. TryAdd and or TryReplace. The former would do nothing in case the activity already exists, the latter would replace the existing activity with the new one 

I see now that this would not fix the issue for this provider, since the only thing a provider is supposed to do is provide activities. So perhaps in this scenario, it's OK to ignore duplicate activities, and maybe log a warning.

PR #4623, opened 2023-11-15
V3 - Fix ParallelForEach Behaviour when empty items

### LINKED ISSUE #4608: V3 - Bad behavior in Parallel For Each
When the Input items its empty its block the workflow in "Suspended" status.
Captura de pantalla 2023-11-07 a las 21 00 12
Captura de pantalla 2023-11-07 a las 21 00 28

I think the behavior should be the same as that of the For Each activity

Captura de pantalla 2023-11-07 a las 21 01 21
Captura de pantalla 2023-11-07 a las 21 01 30

b0rv3g4r4:
This PR fix the ParallelForEach when the Items list of ParallelForEach activity is empty.

Actually the activity keeps suspended.

Fix #4608


sfmskywalker (approved):
Great catch!

PR #4630, opened 2023-11-16
Enhancement/dynamic options provider

jdevillard:
This PR modifies the Options Provider and add some Attributes to Input Properties. 
Theses Attributes will help UI to know if a provider provides Dynamic or Static Data and take decision to call the API when rendering the UI.
Attributes define if an input property is Refreshable, Depends On other Properties or is activable by other properties.

Some work will then be implemented on UI side with this PR.

### === auto-pr-body ===

 
Summary:
- Added a property named "isRefreshable" to `InputDescriptor` class.
- Added a read-only property "isRefreshable" to all implementing instances of `IActivityPropertyOptionsProvider`.
- Added properties "DependsOn" and "ActivatedBy" to `InputAttribute` class.
- Modified the `GetOptionsAsync` method of `Endpoint` class to return an `OptionsItems` model as the result.
- Added properties "DependsOn" and "ActivatedBy" to `InputDescriptor` class.
- Changed the return type of `GetOptionsAsync` from `IDictionary<string, object>?` to `OptionsProviderResult?` in the `IPropertyOptionsResolver` interface.
Refactoring Target:
- Create a separate base class to contain the shared logic between all implementing instances of `IActivityPropertyOptionsProvider`.
- Refactor `InputDescriptor` and `InputAttribute` classes to share logic when appropriate.
- Replace the `string[]?` properties in both `InputDescriptor` and `InputAttribute` classes with a list to eliminate the need for null checks.
- Refactor the `GetOptionsAsync` method to have an input for the `OptionsItems` model to eliminate the extra logic.
- Split `OptionsProviderResult` class into two classes `OptionsProviderMetadataResult` and `OptionsProviderItemsResult` for better scalability.
- Use a data type other than `string` to store the `IsRefreshable` property in `InputDescriptor` class.

### DIFF
--- a/src/clients/Elsa.Api.Client/Resources/ActivityDescriptors/Models/InputDescriptor.cs
+++ b/src/clients/Elsa.Api.Client/Resources/ActivityDescriptors/Models/InputDescriptor.cs
@@ -54,4 +54,20 @@ public class InputDescriptor : PropertyDescriptor
     /// If no driver is specified, the referenced memory block will remain in memory for as long as the expression execution context exists.
     /// </summary>
     public string? StorageDriverType { get; set; }
+
+    /// <summary>
+    /// True if the data provided is dynamic
+    /// </summary>
+    public bool IsRefreashable { get; set; }
+
+    /// <summary>
+    /// The property Name on which this input depends. 
+    /// </summary>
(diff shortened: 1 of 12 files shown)


jdevillard:
close as done with #4688

PR #4634, opened 2023-11-17
fix: workflow definition editor autoLayout

Dharmang92:


Dharmang92:
It seems that after the connection model has been updated into source and target object which contains the activity and port, Antv layout library throws error "Maximum call stack size exceeded" due to it being in a recursive call.

src: https://github.dev/antvis/layout/blob/0.3.11/src/util/math.ts : getEdgeTerminal()

Discord Post: https://discord.com/channels/814605913783795763/1174736804235186186

sfmskywalker (approved):

sfmskywalker:
Many thanks for the PR! 👍🏻

PR #4635, opened 2023-11-17
Feature/change singleton to scope

jeanbaptistedalle:
Change the lifetime of dependancies to use scoped one to prepare the multitenant pull request (to avoid sharing most dependancies like store, dbContext, etc. I kept some in singleton like memory store, registries to avoid loose datas that are added once in hosted services.


sfmskywalker (changes requested):
Looks good to me. I won't block the PR because of the System.* using recording, but please let me know what you think about the other suggestion about moving the scope out of the foreach loop.

Thanks for the great work!

sfmskywalker (approved):

PR #4638, opened 2023-11-17
Features : Allow Customization of DbContext  Schema

jdevillard:
This PR allow you to define a custom Schema Name for the `DbContext` **And** take this information into account in Migration.

for example : 
```csharp
    elsa.UseWorkflowManagement(management => { management.UseEntityFrameworkCore(ef =>
        ef.UseSqlServer("Server=(localdb)\\MSSQLLocalDb;Initial Catalog=Elsav3CustomSchema;Integrated Security=True;",
        new ElsaDbContextOptions()
        {
            SchemaName = "CustomSchemaName",
        }));
    });
```

If not defined, the default Schema Name will be `Elsa`
This change is available for the following provider : 
- SqlServer
- Mysql
- PostgreSql

Sqlite does'nt seems to take into account any schema name.

### === auto-pr-body ===

Summary:
This pull request changes the hardcoded schema "Elsa" to "_schema.Schema" in CreateIndex(), CreateTable() and DropTable() methods, as well as adding a constructor to the Initial.cs file and changing the schema in 11 associated indexes. 

List of changes: 
- Replaced "Elsa" schema with _schema.Schema in the migration constructors 
- Replaced hardcoded string "Elsa" with _schema.Schema for 11 indexes
- Added a constructor to Initial.cs to accept the _schema.Schema variable 

Refactoring Suggestions: 
- Use constants to refer to the schema name to reduce repetition and make the code more readable and maintainable 
- Utilize a loop to traverse the array of indexes which were changed instead of manually updating each one 
- Consolidate code blocks to reduce complexity and improve readability of the code

### DIFF
--- a/src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20231015122151_Initial.cs
+++ b/src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20231015122151_Initial.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Elsa.EntityFrameworkCore.Common.Contracts;
 using Microsoft.EntityFrameworkCore.Migrations;
 
 #nullable disable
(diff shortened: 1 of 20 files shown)


src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20231015122151_Initial.cs
```diff
@@ -8,18 +9,23 @@ namespace Elsa.EntityFrameworkCore.MySql.Migrations.Alterations
     /// <inheritdoc />
     public partial class Initial : Migration
     {
+        private readonly IElsaDbContextSchema _schema;
+        public Initial(IElsaDbContextSchema schema)
+        {
+            _schema = schema ?? throw new ArgumentNullException(nameof(schema));
+        }
```
sfmskywalker, line 16: Is this generated, or do we need to manually do this after generating migrations?

sfmskywalker (approved):

src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20231015122151_Initial.cs
jdevillard: For now , we have to add this manually .

Maybe we can generate this automatically , I know we can hook some element in the generation, but afaik empty constructor is the basic usage in EF so I don't know if we can hook this element in the generation, I've to check this  and I'll propose a new pr if any

sfmskywalker: OK, thanks for clarifying.

PR #4688, opened 2023-12-08
Add a more generic UIHandler to customize how inputAttributes can be handle by UI

jdevillard:
The idea is to give more flexibility between InputDescriptor and UI Implementation to customiser Input Field rendering.

For this, instead of using specific properties in the InputDescriptor , the idea is to use a `Dictionary<String,Object>` where : 
- `key`: represent the UiHandler Name
- `Object` : represent an serializable object that can be use by the custom UI handler to customize the rendering (eg: implement the dynamic options provider, dependency between properties etc...)


sfmskywalker:
I like where you're going with this, please continue 👍🏻

sfmskywalker (approved):

PR #4716, opened 2023-12-20
Elsa2 dotnet8 migration

HubbuHubbu:
Added .NET 8 Support by adding target framework in *.csproj files and added corresponding references.
- open issue: as of today there is only a Beta Version 8 of Pomelo.EntityFrameworkCore.MySql available. Will have to wait before merge

### DIFF
--- a/src/activities/Elsa.Activities.Http.OpenApi/Elsa.Activities.Http.OpenApi.csproj
+++ b/src/activities/Elsa.Activities.Http.OpenApi/Elsa.Activities.Http.OpenApi.csproj
@@ -4,7 +4,7 @@
     <Import Project="..\..\..\configureawait.props" />
 
     <PropertyGroup>
-        <TargetFrameworks>net5.0;net6.0;net7.0</TargetFrameworks>
+        <TargetFrameworks>net5.0;net6.0;net7.0;net8.0</TargetFrameworks>
         <Description>
             Elsa is a set of workflow libraries and tools that enable lean and mean workflowing capabilities in any .NET Core application.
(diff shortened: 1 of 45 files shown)


mohdali:
Hi,

I noticed you have closed this PR. Is it due to MySQL driver?

I think MySQL driver might take a while till it is released. At least that's what happened last time with .NET 7. So we might need to go with the preview version for the time being until it gets released which might take a month or two.

HubbuHubbu:
Hi,

actually it was a first test and playing around with github.
But also there are several reasons why I closed this PR:
- Contributer License Agreement: have to clarify this with my company after holidays
- Testing: I want to create nuget packages from the migrated version first locally and test them (what is the process to create them? Any script for this available?)
- I think the migraton is not complete
- Is it still necessary to support .net 5/6/7 in this migration?

mohdali:
Thank you for clarificaiton.

> * Contributer License Agreement: have to clarify this with my company after holidays
Noted.
> * Testing: I want to create nuget packages from the migrated version first locally and test them (what is the process to create them? Any script for this available?)
I usually try to run the monolith sample with the target framework. This should identify any compilation errors and hopefully no runtime errors occur. Sometimes you get runtime errors due to breaking changes as well. In that case we need to go a step deeper in the code and add conditional preprocessors e.g. `#if NET7.0 #endif`
> * I think the migraton is not complete
ok
> * Is it still necessary to support .net 5/6/7 in this migration?
.NET 5 is out of support but there might be users who have not yet migrated. 6/7 are still under support and we need to keep them. So, no removal at this stage, I think.

PR #4739, opened 2024-01-02
Feature/multitenancy

jeanbaptistedalle:


src/samples/aspnet/Elsa.Samples.AspNet.Tenants.External/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+ConfigurationManager configuration = builder.Configuration;
+
+var identitySection = configuration.GetSection("Identity");
+
+// Add services to the container.
+builder.Services.AddControllers();
+builder.Services.AddElsa(elsa =>
+{
+    ElsaDbContextOptions dbContextOptions = EfCoreDbContextBuilder.BuildDbContextTenantOption();
+    string postgresConnectionString = configuration.GetConnectionString("Postgres")!;
```
sfmskywalker, line 26: I think it would be user-friendly to provide examples using SQLite, as it doesn't necessitate any external dependencies. This approach could make the initial setup and learning curve easier for users. Of course, we should continue to use PostgreSQL in examples specifically designed to demonstrate its usage. This way, we can cater to both simplicity and specificity in our examples.

src/samples/aspnet/Elsa.Samples.AspNet.Tenants.External/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    elsa.UseWorkflowManagement(management => management.UseEntityFrameworkCore(ef => ef.UsePostgreSql(postgresConnectionString, dbContextOptions)));
+    elsa.UseWorkflowRuntime(runtime => runtime.UseEntityFrameworkCore(ef => ef.UsePostgreSql(postgresConnectionString, dbContextOptions)));
+
+    elsa.UseIdentity(options => identitySection.Bind(options));
+
+    elsa.UseTenants(configuration => configuration
+            .UseExternalTenantProvider(options => identitySection.Bind(options))
+            .UseEfcoreStrategies());
```
sfmskywalker, line 42: I suggest using the casing `UseEFCoreStrategies` for this identifier. Since "EF" stands for "Entity Framework," and our naming convention typically capitalises each letter in two-letter abbreviations, this format aligns well with our standards. It's consistent with how we handle other abbreviations like "IO" for "Input Output" and "UI" for "User Interface."

src/modules/Elsa.Tenants/Elsa.Tenants.csproj
```diff
@@ -0,0 +1,29 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="7.0.13" />
```
sfmskywalker, line 11: I noticed that the Tenants module currently has a direct dependency on EF Core. Could you clarify if this is a deliberate design choice? Additionally, I'm curious about our strategy for integrating multi-tenancy with other database providers like MongoDB, Dapper, and potentially custom ones. How do we plan to accommodate these varied environments while maintaining the module's versatility?

src/modules/Elsa.Tenants/Features/TenantsFeature.cs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public TenantsFeature UseEfcoreStrategies()
+    {
+        Services
+            .AddScoped<IDbContextStrategy, MustHaveTenantIdBeforeSavingStrategy>()
+        ;
+
+        return this;
+    }
```
sfmskywalker, line 96: I think it would be better to have EF Core specific code live in the EF Core project (Elsa.EntityFrameworkCore.Common perhaps?). This way, we don't need to update the Elsa.Tenants module when a new persistence provider is introduced; instead, the multi-tenancy enhancements can be implemented there. What do you think?

src/common/Elsa.Mediator/Middleware/Notification/Components/NotificationHandlerInvokerMiddleware.cs
```diff
@@ -38,6 +38,8 @@ public async ValueTask InvokeAsync(NotificationContext context)
         var handlers = _notificationHandlers.Where(x => handlerType.IsInstanceOfType(x)).DistinctBy(x => x.GetType()).ToArray();
         var strategyContext = new NotificationStrategyContext(notification, handlers, _logger, _serviceProvider, context.CancellationToken);
 
+        //TODO JBD Set tenantId here or in every handler ?
+
```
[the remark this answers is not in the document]
jeanbaptistedalle: That's an old TODO I forgot to clean, sorry 😅

src/modules/Elsa.Tenants/Elsa.Tenants.csproj
jeanbaptistedalle: Indeed, that's not a good design choice I will split that to keep the tenant project agnostic of the persistence provider

src/modules/Elsa.Tenants/Features/TenantsFeature.cs
jeanbaptistedalle: I'm agree with you, I'm currently make the changes

src/modules/Elsa.Common/Extensions/ServiceScopeFactoryExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Elsa.Common.Contracts;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Elsa.Common.Extensions;
+public static class ServiceScopeFactoryExtensions
```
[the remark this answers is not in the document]
jeanbaptistedalle: Moved to an ITenantServiceScopeFactory like suggested

sfmskywalker:
Note to self:

- Introduced a breaking change to simplify StoredBookmark entity by making its PK's name consistent with other entities.
  - Release notes should include SQL scripts to copy BookmarkId values to new Id column

sfmskywalker (approved):

sfmskywalker:
Overall, the core multi tenancy features look good. There are a few more items I'd like to add still:

- Tenant resolver based on site domain (origin)
- Tenant resolver based on path prefix.
- MongoDB + Dapper support

PR #4750, opened 2024-01-04
New options to control retry logic for transient failures

NightWuYo:
Add a new option to control the number of automatic retries for transient failures.

Currently when a workflow is triggered from Elsa Studio, it will retry for 3 times automatically if failed, and each time it will start from the the beginning. This means some activities may be run for up to 4 times (for example, when the workflow has two activities, and the first one activity always gets success and the second one is always failed, the first one will be run for 4 times successfully).

Fixes elsa-workflows/elsa-studio#108


sfmskywalker:
I was hoping that maybe we could expose the configuration lambda to the calling code so that we can not only change the retry count, but also e.g. the time between retry attempts, and perhaps even add additional strategies if the user wants to?

NightWuYo:
@sfmskywalker I have introduced "Func<int, TimeSpan> SleepDurationProvider" option.

I think the RetryCount and SleepDurationProvider should be enough.

sfmskywalker (approved):

PR #4766, opened 2024-01-07
Use correct dialect for SqlServerDbConnectionProvider

### LINKED ISSUE #4723: SqlServerDbConnectionProvider using Sqlite dialect
We found this line appears to be incorrect when using the SqlServerDbConnectionProvider with Dapper:
https://github.com/elsa-workflows/elsa-core/blob/6a5835e51ac2ffc70a77011149fa9c5ceab79328/src/modules/Elsa.Dapper/Services/SqlServerDbConnectionProvider.cs#L46

The fix should be to use the SqlServerDialect instead.

NightWuYo:
Fixes #4723


sfmskywalker (approved):
Very nice catch, thank you! 👍🏻

PR #4771, opened 2024-01-08
Changed Service Lifetime of RunTaskDispatcher

yspanj:
Following the guide for External Application Integration from the documentation an exceptions happens from the RunTask activity.

Changing the lifetime of RunTaskDispatcher to scoped resolved this.


yspanj:
I just noticed that this change came from #4753 so just reverting it is probably not the correct resolution.

PR #4772, opened 2024-01-08
Add requester info into WorkflowExecutionContext

NightWuYo:
Add Requester property to WorkflowExecutionContext, along with some other classes.

This new property can be used to save the requester information: who or which service triggered this workflow.

In particular, when a workflow is triggered from Studio, the user (who logged in) identity name will be set. This gives custom activities the ability to know which user has triggered the workflow (when from Studio), which helps:
- Audit purpose
- Permission control
- Custom logic based on the request user


sfmskywalker:
The functional requirement makes sense, but we will have to think about the right design.

jdevillard:
Hello @NightWuYo , 

Thanks for your PR, as discuss on discord and as @sfmskywalker  said, this functional requirement will be very useful for futur usage of Elsa. However, the idea is not to just embed a requester property but really a `Identity` Object that can be use for any kind of validation rules, choice, or process in a Workflow. 

For that, we need to think about the design and the right proposition on this new object. 

This is in the Pipe. We can track the issue with #4793

NightWuYo:
Good to know that this functional requirement will be implemented.
Hopefully it will be ready soon.

PR #4779, opened 2024-01-09
Use NET 8 SDK artifacts output

lahma:
Use the new [NET 8 SDK artifacts output layout](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output) feature which consolidates `bin` and `obj` folders under root level `artifacts` directory. Need to have `Directory.Build.props` to root level for proper configuration (you might want to consider such standard way over common.props too). For a solution with large amount of projects this can help both with I/O performance and cleanup (deleting `obj` and `bin` folders as they are in single place).

To remove old cruft, might be good idea to run locally in solution root directory:

```powershell
Get-ChildItem .\ -include bin,obj -Recurse | foreach ($_) { remove-item $_.fullname -Force -Recurse }
```

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+
+  <PropertyGroup>
+    <UseArtifactsOutput>true</UseArtifactsOutput>
+  </PropertyGroup>
+
+</Project>


sfmskywalker (approved):
Hey there, thanks for the tips and the contribution! I will check it out, as well as NUKE. We are currently using GH Actions, and we should enable it for PR's as well 👍🏻

lahma:
Thanks for accepting the PR, with such large codebase with multiple projects it's now easier to have all compilation artifacts under single directory.

I would be happy to submit a NUKE proposal for your review, no pressure to accept but would show how things would be different (for better or worse).

lahma:
Great, I'll try to whip something up,  might take a  bit though 🙂

sfmskywalker:
Awesome, thank you. Of course, no rush :)

markjones-WK:
fyi, i just pulled a fresh copy of the srouce from elsa-core-main and was unable to build as I was getting artifact errors.  I deleted the new Directory.Build.props file in the root as restarted VS and was able to build.  Thi smay need more testing.  I switched to th3 3.0.3 branch that does not have this file, and all is good.

sfmskywalker:
Thanks for testing and sharing your findings @markjones-WK. I haven't had a chance yet to try for myself, but will report here my own findings when I have them.

markjones-WK:
Thanks.  Just curious.  Why would it be pushed to the main branch and not the 3.0.3 branch or another branch other than main?

sfmskywalker:
The 3.0.3 branch is for the upcoming patch release and ideally not for larger changes. The main branch is for larger changes and additions from which 3.1 will be released around March.

lahma:
@sfmskywalker can you please revert this for now, I somehow missed the incompatibility between root outputs and ProtoGen tasks, they seem to be quite strict about being under project folder. I'll make amends with a PR to build the PRs and then maybe try this again when CI judge rules the code as compatible...

markjones-WK:
> The 3.0.3 branch is for the upcoming patch release and ideally not for larger changes. The main branch is for larger changes and additions from which 3.1 will be released around March.

I get it, but isn't the Main branch supposed for the current release branch (3.0.2)?  As you mentioned, this PR was not tested and is now asking to be reverted, so if I wanted the released 3.0.2 source code, where would I go to get it?

sfmskywalker:
> @sfmskywalker can you please revert this for now, I somehow missed the incompatibility between root outputs and ProtoGen tasks, they seem to be quite strict about being under project folder. I'll make amends with a PR to build the PRs and then maybe try this again when CI judge rules the code as compatible...

Of course, no worries. Thank you for having figured that one out 👍🏻

sfmskywalker:
@markjones-WK The main branch should contain the latest and greatest changes from which we do minor releases (3.1 being the first upcoming one). I am experimenting to see if we can provided patches (revision releases) via tags only. Not sure if it will work out, but so far it seems OK. If you want to get the source code for the 3.0.2 release, checkout [the commit with that tag](42db09c689c0e353fdaa030160e0aa372603cfb5).

lahma:
I've opened https://github.com/asynkron/protoactor-dotnet/issues/2095 to track the problematic behavior.

sfmskywalker:
Thank you! I noticed it has been tagged as a bug by the maintainer 👍🏻

PR #4813, opened 2024-01-19
Workflow cancellation

raymonddenhaan:
Allows the cancellation of running and suspended workflows including any child activities.

Limitations:
- This feature has not yet been implemented in the user interface
- Dispatched child activities are not being cancelled


raymonddenhaan:
@dotnet-policy-service agree company="nexxbiz"

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.Cancel.cs
```diff
@@ -0,0 +1,43 @@
...
+    private void CancelWorkflow()
+    {
+        Bookmarks.Clear();
+        _completionCallbackEntries.Clear();
+
+        if (Status != WorkflowStatus.Running && SubStatus != WorkflowSubStatus.Faulted)
+            return;
+
+        AddExecutionLogEntry("Workflow cancelled");
+        // How to guarantee persistence of bookmarks??
```
sfmskywalker, line 33: I'm wondering if we even need to persist bookmarks at this point, since we're cancelling the workflow anyway. Do you have a scenario in mind where the persistence of bookmarks is important?

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowRuntime.cs
```diff
@@ -90,6 +103,70 @@ public async Task<WorkflowExecutionResult> StartWorkflowAsync(string definitionI
     {
         return await StartWorkflowAsync(definitionId, options);
     }
+    
+    /// <inheritdoc />
+    public async Task CancelWorkflowAsync(string workflowInstanceId, CancellationToken cancellationToken)
+    {
+        //Should we add a cancellationtoken?
```
sfmskywalker, line 110: I think so. After all, we might want to cancel the cancellation 😄 .

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowRuntime.cs
```diff
@@ -90,6 +103,70 @@ public async Task<WorkflowExecutionResult> StartWorkflowAsync(string definitionI
...
+
+            var workflowDefinition = await _workflowDefinitionService.FindAsync(workflowState.DefinitionId, VersionOptions.SpecificVersion(workflowState.DefinitionVersion), cancellationToken);
+
+            if (workflowDefinition == null)
+                throw new Exception("Workflow definition not found");
+
+            var workflow = await _workflowDefinitionService.MaterializeWorkflowAsync(workflowDefinition, cancellationToken);
+            workflowExecutionContext = await WorkflowExecutionContext.CreateAsync(_serviceProvider, workflow, workflowState, cancellationTokens: cancellationToken);
+
+            //Is this still necessary?
```
sfmskywalker, line 141: Good point. It looks like this is not necessary, given that the `await WorkflowExecutionContext.CreateAsync` call on line 139 already maps the state into the context.

sfmskywalker (approved):
Awesome PR - thank you. You provided an elegant solution to cancelling both suspended and actively running workflows and activities, and you  provided tests. In addition, I think that you increased the overall code quality 🔥

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.Cancel.cs
raymonddenhaan: Persistence of bookmarks in the case of cancellation would mean cleaning up the database, Not terribly important, but might be good for consistency.

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowRuntime.cs
raymonddenhaan: Removed the Apply call

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowRuntime.cs
raymonddenhaan: Removed the comment as cancelling the cancellation flow could have some unwanted effects.

PR #4815, opened 2024-01-19
Add NUKE build support

lahma:
This is the first step to add minimal NUKE build support. You're using some advanced GitHub Actions definitions, so it's not that easy to shift logic from them to the C# build. As first iteration I'm suggesting to move the build part to calling the NUKE infrastructure and generating the PR template that will be run against any PRs for this repo.

I removed the duplicate projects from the solution file as NUKE wasn't too happy about them. Also let Visual Studio open the file and update project GUIDs and fixed projects locations. After solution was saved Rider no longer complained about project loading problems either.


sfmskywalker (approved):
This looks awesome. Much appreciated!

PR #4828, opened 2024-01-23
Use Directory.Build.props

lahma:
By replacing `common.props` with `Directory.Build.props` build will adhere to the more modern way of doing things. `Directory.Build.props` is inherited to sub-directories and sets the properties. No need for or to remember a separate import.


sfmskywalker (approved):
I had seen build props before, but I didn't know about the convention that it is included in all projects like this without having to explicitly import them. Nice!

sfmskywalker:
I'll wait for you to mark the PR as Ready for review before merging.

lahma:
So I went further down the rabbit hole and removed the `framework.props` and `configureawait.props`. Now also these propertis are handled via the standard `Directory.Build.props`, inheriting defaults from parents. Now csproj's don't need to have `ImplicitUsings` or `Nullable` either, all inherited.

If `samples` directory would be on root level like `src` and `test`, wouldn't need to special case them in `src/Directory.Build.props`, but maybe a future improvement.

lahma:
OK so many files have changed, but hopefully they are quite consistent 🫠

dwoldo:
This approach is a great way to minimize those "dependency hell" scenarios. We've started using this in our projects, it makes managing versions much easier. Ps. You can override on a project-level by adding `VersionOverride` parameter:

see: [Overriding package versions](https://devblogs.microsoft.com/nuget/introducing-central-package-management/#:~:text=Overriding%20package%20versions)

lahma:
My hope is to tackle the versioning more deeply with a different PR converting to [Central Package Management,](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) which should improve things a bit too.

sfmskywalker:
> If samples directory would be on root level like src and test, wouldn't need to special case them in src/Directory.Build.props, but maybe a future improvement.

I agree, let's move them to the same level.

PR #4842, opened 2024-01-26
Bulk Dispatch Workflows Enhancements

sfmskywalker:
This PR updates the BulkDispatchWorkflows with support for async-enumerables and adds support for background execution of the activity itself. Before this PR, the activity would remain in the Suspended state when executed from the background due to the fact that its bookmarks were not synced with the persisted state.


src/modules/Elsa.Workflows.Core/Extensions/BackgroundActivityExecutionContextExtensions.cs
```diff
@@ -45,6 +45,22 @@ public static void SetBackgroundOutcomes(this ActivityExecutionContext activityE
...
+    /// </summary>
+    public static void SetBackgroundCompletion(this ActivityExecutionContext activityExecutionContext)
+    {
+        activityExecutionContext.SetProperty("BackgroundCompletion", true);
+    }
+    
+    /// <summary>
+    /// Gets the background outcomes.
+    /// </summary>
+    public static bool? GetBackgroundCompleted(this ActivityExecutionContext activityExecutionContext)
```
raymonddenhaan, line 59: GetBackgroundCompletion

src/modules/Elsa.Workflows.Core/Extensions/ItemSourceActivityExecutionContextExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <returns>An asynchronous enumerable of items from the source collection.</returns>
+    public static async IAsyncEnumerable<T> GetItemSource<T>(this ActivityExecutionContext context, Input<object> input)
+    {
+        var items = context.Get(input);
+
+        if (items == null)
+            yield break;
+        
+        var itemsType = items.GetType();
+        if (itemsType.Name == "AsyncEnumerableAdapter`1")
```
raymonddenhaan, line 27: Is there any way we can move away from the magic string here?

src/modules/Elsa.Workflows.Core/Extensions/BackgroundActivityExecutionContextExtensions.cs
sfmskywalker: You're right, that sounds more correct 👍🏻

PR #4850, opened 2024-01-30
Add support for list/array conversion in JavaScript evaluator

### LINKED ISSUE #4849: Enhance JS Evaluator for Array Function Support on .NET Lists
Currently, users working with dynamic objects in JavaScript within Elsa Workflows (such as `Dictionary` and `ExpandoObject`) face limitations when trying to use array functions like `length` and `filter`. These functions are not directly compatible with .NET List types. To address this issue, there is a proposal to update the JavaScript evaluator. The proposed update involves creating a modified copy of the input or variable value, where .NET List fields are converted to JavaScript array types.

### Considerations
- **Caveat**: This modification may impact use cases where users intend to use .NET List functionalities directly in JavaScript, such as adding items to a list. This is currently viewed as an edge case, but a workaround will be necessary.

### Goal
- To ensure a smoother and more intuitive experience when using array functions in JavaScript with dynamic objects in Elsa Workflows.

We welcome further input and discussion on this proposed enhancement.

sfmskywalker:
A `StringObjectDictionaryConverter` has been added, which converts all `IList` fields of an object to array fields. This enhances the interaction of JS expressions with list types, allowing common Array methods to be applied directly. Furthermore, defaults have been added to the `ExpressionEvaluatorOptions` parameters to improve usability. Lastly, adjustments were made to ensure variables are correctly set while evaluating expressions.

Closes #4849

### DIFF
--- a/src/bundles/Elsa.Server.Web/Program.cs
+++ b/src/bundles/Elsa.Server.Web/Program.cs
@@ -26,8 +26,8 @@ const bool useDapper = false;
 const bool useProtoActor = false;
 const bool useHangfire = false;
 const bool useQuartz = true;
-const bool useMassTransit = true;
-const bool useMassTransitAzureServiceBus = false;
+const bool useMassTransit = false;
+const bool useMassTransitAzureServiceBus = true;
 const bool useMassTransitRabbitMq = false;
 
 var builder = WebApplication.CreateBuilder(args);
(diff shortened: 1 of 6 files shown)


src/bundles/Elsa.Server.Web/Program.cs
```diff
@@ -26,8 +26,8 @@
 const bool useProtoActor = false;
 const bool useHangfire = false;
 const bool useQuartz = true;
-const bool useMassTransit = true;
-const bool useMassTransitAzureServiceBus = false;
+const bool useMassTransit = false;
+const bool useMassTransitAzureServiceBus = true;
```
raymonddenhaan, line 30: This will do nothing since the useMassTransit is already set to false.

raymonddenhaan (approved):

src/bundles/Elsa.Server.Web/Program.cs
sfmskywalker: You're correct.

PR #4856, opened 2024-01-31
Remove StringObjectDictionaryConverter and update JintJavaScriptEvaluator

sfmskywalker:
The StringObjectDictionaryConverter class was removed as it is no longer necessary. Instead, the JintJavaScriptEvaluator now wraps objects in ObjectWrapper instances and sets their prototype to Array.prototype if they are array-like. This allows for more convenient and intuitive use of Lists and arrays in JavaScript code within Elsa. A new integration test was created to validate this functionality.


test/integration/Elsa.IntegrationTests/Scenarios/JavaScriptListAndArray/Tests.cs
raymonddenhaan, line 1: It would be nice to test whether this works as expected with arrays, objects and primitives

raymonddenhaan (approved):

test/integration/Elsa.IntegrationTests/Scenarios/JavaScriptListAndArray/Tests.cs
sfmskywalker: Sure. What additional scenarios do you have in mind, specifically? The current test covers dealing with Lists as Arrays.

raymonddenhaan: Nested Lists, primitives and objects as the root object and maybe a test with an array instead of a list, just to see that this functionality doesn't somehow break that.

sfmskywalker: Ah, like that. Yes, makes sense 👍🏻

sfmskywalker: I added a few more cases that mixes primitives, objects and regular array: https://github.com/elsa-workflows/elsa-core/pull/4857

Do let me know if you see an opportunity for adding more cases.

raymonddenhaan: I think that with those cases you've covered most of the scenarios. Maybe some edge cases we're still missing, but it is good enough for now.

PR #4864, opened 2024-02-02
Fix schema for Quartz Postgres Store

mohdali:
The Postgres Migration for Quartz puts the tables under quartz schema. However, the job store does not have any option to provide schema definition. However, it is possible to specify in the table prefix.

Methods and file names had to be renamed to fix copy error. (might be a breaking change for existing Elsa 3.0).


sfmskywalker (approved):
Thanks @mohdali !

PR #4876, opened 2024-02-03
Add GZip compression for WorkflowInstance Data field on EF Core

### LINKED ISSUE #4875: Enable optional gzip compression when saving workflow instances
Introduce an optional feature for gzip compression in the workflow engine to enhance data storage efficiency.

This functionality should allow for the compression of workflow instances into gzip format prior to committing them to the database. 

Additionally, ensure that upon retrieval, the data is decompressed back to its original state. The activation of gzip compression should be configurable and determined at the application startup, providing flexibility to enable or disable this feature.

sfmskywalker:
This PR adds the ability to store workflow state in a compressed format for EF Core.

Future updates may include other persistence provider support such as Dapper and MongoDB.

Closes #4875


sfmskywalker:
I want to apply this also to the ActivityState field of the ActivityExecutionRecord table - for some customers, these fields can reach many megabytes (10MB for some scenarios that I noticed).

cristinamudura (approved):

src/modules/Elsa.EntityFrameworkCore/Modules/Runtime/ActivityExecutionLogStore.cs
```diff
@@ -58,42 +73,56 @@ public EFCoreActivityExecutionStore(EntityStore<RuntimeElsaDbContext, ActivityEx
 
     private async ValueTask OnSaveAsync(RuntimeElsaDbContext dbContext, ActivityExecutionRecord entity, CancellationToken cancellationToken)
     {
-        dbContext.Entry(entity).Property("SerializedActivityState").CurrentValue = entity.ActivityState != null ? await _safeSerializer.SerializeAsync(entity.ActivityState, cancellationToken) : default;
+        var compressionAlgorithm = _options.Value.CompressionAlgorithm ?? nameof(None);
+        var serializedActivityState = entity.ActivityState != null ? await _safeSerializer.SerializeAsync(entity.ActivityState, cancellationToken) : default;
+        var compressedSerializedActivityState = serializedActivityState != null ? await _compressionStrategyResolver.Resolve(compressionAlgorithm).CompressAsync(serializedActivityState, cancellationToken) : default;
```
raymonddenhaan, line 78: We do not have to resolve the strategy for each call unless we want to be able to change options while running.

migrations/efcore-3.0.sh
```diff
@@ -0,0 +1,38 @@
...
+        providerPath="../src/modules/Elsa.EntityFrameworkCore.$provider"
+        migrationsPath="Migrations/$module"
+    
+        echo "Updating migrations for $provider..."
+        echo "Provider path: ${providerPath:?}/${migrationsPath}"
+        echo "Migrations path: $migrationsPath"
+        echo "Connection string: ${connStrings[$provider]}"
+    
+        # 1. Delete the existing migrations folder
+        rm -rf "${providerPath:?}/${migrationsPath}"
```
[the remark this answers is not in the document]
sfmskywalker: True, we should never have to regenerate migrations from scratch.

lahma:
RavenDB folks did come compression algorithm comparison recently against JSON data: https://github.com/ravendb/ravendb/discussions/17678

src/modules/Elsa.EntityFrameworkCore/Modules/Runtime/ActivityExecutionLogStore.cs
sfmskywalker: True, we could store it as a field on this class, or introduce a "compression provider" type that resolves the active compression strategy once.

sfmskywalker: On second thought, each record may use a different algorithm, since the setting may change over time. Therefore it's important to resolve the correct algo per record. A small optimization can be done however by using a dictionary for quicker lookups.

sfmskywalker:
> RavenDB folks did come compression algorithm comparison recently against JSON data: [ravendb/ravendb#17678](https://github.com/ravendb/ravendb/discussions/17678)

Wow, ZSTD is not only significantly faster, but produces smaller output as well. 
Definitely gonna add that as a strategy and try it out 👍🏻

PR #4877, opened 2024-02-04
Convert to use Central Package Management

lahma:
* Now all versions are defined in root level `Directory.Packages.props`, some special casing based original `net6.0`, `net7.0`, `net8.0` checks
* Moved commont test package refererences (xunit, Microsoft test SDK, coverlet)  to `tests\Directory.build.props` which cleaned the test csprojs too
* Add GitHubActionsTestLogger which will show a better summary about tests and possible problems - [see the output at the bottom now](https://github.com/elsa-workflows/elsa-core/actions/runs/7773218049?pr=4877)


lahma:
@sfmskywalker I see that you removed the `v3*` branch selection, when you modify the build definition you need to run `build.cmd` locally afterwards in order for NUKE to regenerate the GH Actions workflow. This PR contains the updated workflow file was was generated when I executed NUKE.

lahma:
FastEndPoints had two different versions between NET6/7 and NET8: `5.19.2` and `5.21.2`. I didn't see a reason to have two separate for these and thus I used common version ref `5.21.2`. Can split back to framework dependant if needed.

sfmskywalker:
Dude this is amazing - a massive thank you! 🙌🏻

sfmskywalker:
> @sfmskywalker I see that you removed the `v3*` branch selection, when you modify the build definition you need to run `build.cmd` locally afterwards in order for NUKE to regenerate the GH Actions workflow. This PR contains the updated workflow file was was generated when I executed NUKE.

Will do! Thanks for pointing it out 👍🏻

sfmskywalker:
@lahma 
I should probably have asked this before merging the PR, but not knowing too much about CPM, do you know what effect this will have on user projects referencing a subset of Elsa packages - will they automatically inherit ALL package references in `Directory.Packages.props`? 🤔

sfmskywalker:
Ah I think I see. So projects still explicitly reference the packages they need, but without version number; that is all managed centrally now.

lahma:
> @lahma 
> I should probably have asked this before merging the PR, but not knowing too much about CPM, do you know what effect this will have on user projects referencing a subset of Elsa packages - will they automatically inherit ALL package references in `Directory.Packages.props`? 🤔 

It's the same csproj dictated dependencies. The versions are just managed centrally in one file. So if project X references Polly it will be part of nuget package deps, but a non-referenced Jint won't be.

lahma:
Rider and Visual Studio also have native support for updating these via Ui/finding updates. Of course the Condition-based versions need to be carefully updated - but now in one place!

Issue #4885, opened 2024-02-05
MongoDB : already a serializer registered


bbenameur:
Hello,
There are a Bug with elsa when using mongoDb only on the integration Tests,
On the integration tests as you know the program class will be called many time using CustomWebApplicationFactory,
and on the AddElsa service registry will register the serializer 
```csharp
builder.Services.AddElsa(elsa =>
	elsa
		.UseMongoDb(orchestratorOptions.PersistanceConnectionString!)
......
``` 
With 2 or more integration test i have this error:

Message: 
MongoDB.Bson.BsonSerializationException : There is already a serializer registered for type Object.

  Arborescence des appels de procédure: 
BsonSerializerRegistry.RegisterSerializer(Type type, IBsonSerializer serializer)
BsonSerializer.RegisterSerializer(Type type, IBsonSerializer serializer)
MongoDbFeature.RegisterSerializers()
MongoDbFeature.Apply()
Module.Apply()
ModuleExtensions.AddElsa(IServiceCollection services, Action`1 configure)
ElsaExtensions.AddElsa(WebApplicationBuilder builder, IConfiguration configuration) ligne 39
Program.Main(String[] args) ligne 15

When i see the code i see that you register the serializer  on the RegisterSerializers() method

```csharp
 private static void RegisterSerializers()
 {
     BsonSerializer.RegisterSerializer(typeof(object), new PolymorphicSerializer());
     BsonSerializer.RegisterSerializer(typeof(Type), new TypeSerializer());
     BsonSerializer.RegisterSerializer(typeof(Variable), new VariableSerializer());
     BsonSerializer.RegisterSerializer(typeof(Version), new VersionSerializer());
     BsonSerializer.RegisterSerializer(typeof(JsonElement), new JsonElementSerializer());
 }
```

to fix this can you please user  BsonSerializer.TryRegisterSerializer(typeof(object), new PolymorphicSerializer());
..
Thanks

sfmskywalker:
Closed via #4886

PR #4886, opened 2024-02-05
issues(4885)

bbenameur:
Hello,
See issue: 
https://github.com/elsa-workflows/elsa-core/issues/4885

Based on the context provided, it appears that we are encountering a serialization registration issue when running integration tests with Elsa and MongoDB. The error message indicates that a serializer for the Object type is being registered multiple times, which is not allowed by the MongoDB driver.

The code snippet provided suggests a solution to use BsonSerializer.TryRegisterSerializer instead of BsonSerializer.RegisterSerializer.

```csharp
private static void RegisterSerializers()
{
    BsonSerializer.TryRegisterSerializer(typeof(object), new PolymorphicSerializer());
    BsonSerializer.TryRegisterSerializer(typeof(Type), new TypeSerializer());
    BsonSerializer.TryRegisterSerializer(typeof(Variable), new VariableSerializer());
    BsonSerializer.TryRegisterSerializer(typeof(Version), new VersionSerializer());
    BsonSerializer.TryRegisterSerializer(typeof(JsonElement), new JsonElementSerializer());
}
```

### DIFF
--- a/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
+++ b/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
@@ -48,11 +48,11 @@ public class MongoDbFeature : FeatureBase
 
     private static void RegisterSerializers()
     {
-        BsonSerializer.RegisterSerializer(typeof(object), new PolymorphicSerializer());
-        BsonSerializer.RegisterSerializer(typeof(Type), new TypeSerializer());
-        BsonSerializer.RegisterSerializer(typeof(Variable), new VariableSerializer());
-        BsonSerializer.RegisterSerializer(typeof(Version), new VersionSerializer());
-        BsonSerializer.RegisterSerializer(typeof(JsonElement), new JsonElementSerializer());
+        BsonSerializer.TryRegisterSerializer(typeof(object), new PolymorphicSerializer());
+        BsonSerializer.TryRegisterSerializer(typeof(Type), new TypeSerializer());
+        BsonSerializer.TryRegisterSerializer(typeof(Variable), new VariableSerializer());
+        BsonSerializer.TryRegisterSerializer(typeof(Version), new VersionSerializer());
+        BsonSerializer.TryRegisterSerializer(typeof(JsonElement), new JsonElementSerializer());
     }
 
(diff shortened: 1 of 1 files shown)


bbenameur:
@dotnet-policy-service agree [company="Pluxee.fr"]

bbenameur:
@sfmskywalker

bbenameur:
@dotnet-policy-service agree company="pluxee"

sfmskywalker (approved):

bbenameur:
@dotnet-policy-service agree

hsnsalhi:
this doesn't resolve the problem since TryRegisterSerializer throws an exception in addition to returning false if it couldn't register the serializer.
I suggest to put the code in a try catch block in a separate method to ensure the registration of every serializer even if one of them couldn't be registered.

```cs
        private static void RegisterSerializers()
        {
            TryRegisterSerializerWithoutException(typeof(object), new PolymorphicSerializer());
            TryRegisterSerializerWithoutException(typeof(Type), new TypeSerializer());
            TryRegisterSerializerWithoutException(typeof(Variable), new VariableSerializer());
            TryRegisterSerializerWithoutException(typeof(Version), new VersionSerializer());
            TryRegisterSerializerWithoutException(typeof(JsonElement), new JsonElementSerializer());
        }


        private static void TryRegisterSerializerWithoutException(Type type, IBsonSerializer serializer)
        {
            try
            {
                BsonSerializer.TryRegisterSerializer(type, serializer);
            }
            catch (BsonSerializationException ex)
            {
                // should we log the exception ?
            }
        }
```

hsnsalhi:
BTW, you will face the same problem with `RegisterClassMap`. I suggest you to replace it by `TryRegisterClassMap` this method does not throw an exception if it cannot register the class map. so no need to put a try catch block for it.

```cs
        private static void RegisterClassMaps()
        {
            BsonClassMap.TryRegisterClassMap<StoredBookmark>(cm =>
            {
                cm.AutoMap(); // Automatically map other properties
                cm
                    .MapIdProperty(b => b.BookmarkId)
                    .SetSerializer(new StringSerializer(BsonType.String));
            });
        }

```

PR #4894, opened 2024-02-07
issues(4893): try register serializer or skip when already registred …

bbenameur:
Related to this fix https://github.com/elsa-workflows/elsa-core/pull/4886
it appeared that this fix will not resolve the probelem,
because even with a try it will always have a new instance which is totally different from what is registred
see code on MongoDB.Bson.Serialization

```csharp
 public bool TryRegisterSerializer(Type type, IBsonSerializer serializer)
 {
     if (type == null)
     {
         throw new ArgumentNullException(nameof(type));
     }
     if (serializer == null)
     {
         throw new ArgumentNullException(nameof(serializer));
     }
     EnsureRegisteringASerializerForThisTypeIsAllowed(type);

     if (_cache.TryAdd(type, serializer))
     {
         return true;
     }
     else
     {
         var existingSerializer = _cache[type];
         if (!existingSerializer.Equals(serializer)) // with new() ... it's always a different instance.
         {
             var message = $"There is already a different serializer registered for type {BsonUtils.GetFriendlyTypeName(type)}.";
             throw new BsonSerializationException(message);
         }
         return false;
     }
 }
```

To fix it:

```csharp

TryRegisterSerializerOrSkip(typeof(object), new PolymorphicSerializer());
TryRegisterSerializerOrSkip(typeof(Type), new TypeSerializer());
TryRegisterSerializerOrSkip(typeof(Variable), new VariableSerializer());
TryRegisterSerializerOrSkip(typeof(Version), new VersionSerializer());
TryRegisterSerializerOrSkip(typeof(JsonElement), new JsonElementSerializer());

private static void TryRegisterSerializerOrSkip(Type type, IBsonSerializer serializer)
{
	try
	{
		BsonSerializer.TryRegisterSerializer(type, serializer);
	}
	catch (BsonSerializationException ex)
	{
		// should we log the exception ?
	}
}

```

### DIFF
--- a/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
+++ b/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
@@ -48,16 +48,16 @@ public class MongoDbFeature : FeatureBase
 
     private static void RegisterSerializers()
     {
-        BsonSerializer.TryRegisterSerializer(typeof(object), new PolymorphicSerializer());
-        BsonSerializer.TryRegisterSerializer(typeof(Type), new TypeSerializer());
-        BsonSerializer.TryRegisterSerializer(typeof(Variable), new VariableSerializer());
-        BsonSerializer.TryRegisterSerializer(typeof(Version), new VersionSerializer());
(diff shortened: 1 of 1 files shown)


bbenameur:
Hello @sfmskywalker

Issue #4799, opened 2024-01-15
Allow configuration of LogRecord storage at differents levels


jdevillard:
In the actual implementation, `LogRecords` are done using differents interfaces : 
- `IActivityExecutionStore` that store serialized information about activity I/O 
- `IWorkflowExecutionLogStore` that store serialize information about Workflow and ActivityContext. (maybe with redondant information regarding the `IActivityExecutionStore`

This can lead to data store that grow rapidly if you done some activity with large transformation (large input/ouput property which will be serialized).

The idea is to give the ability to the user to configure the storage strategy at different level : 
- At App Level -> store : true/false
- At Workflow Level -> store : default (inherit from App Level) / true (override) / false (override)
- At Activity Level -> store : default (inherit from Workflow Level) / true (override) / false (override)

For UI Component : 
- App Level : N/A  this is configured at Startup
- Workflow level : dropdown in the property tab of the Workflow
- Activity level : new tab for persistence information, with a list of the I/O property and a drop down list for each which allow choosing the strategy.

jdevillard:
I think this issue can be closed ? 🧐

sfmskywalker:
@jdevillard Yes, thank you for this awesome contribution! 🙌🏻

PR #4911, opened 2024-02-08
LogRecord storage at differents levels

jdevillard:
This PR init the capability to choose if we want to store input properties and payload (output) of activity in the Activity Execution Log Record regarding the #4799 

This PR is necessary for the UI development : 
- https://github.com/elsa-workflows/elsa-studio/pull/136
- https://github.com/elsa-workflows/elsa-studio/issues/118


src/clients/Elsa.Api.Client/Resources/WorkflowDefinitions/Enums/PersistenceStrategy.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Elsa.Api.Client.Resources.WorkflowDefinitions.Enums;
+
+/// <summary>
+/// Define the Persistence strategy to store information
+/// </summary>
+public enum PersistenceStrategy
```
sfmskywalker, line 6: I am wondering if `LogPersistenceMode` might be a better name. It is more focussed at logging, and "mode" is an alternative to "strategy" - we already use "strategy" for types that implement the strategy pattern. "Persistence" on its own is too broad.

What do you think?

sfmskywalker (changes requested):
Looks good, just a small request to revisit the naming of "persistence strategy" etc. to be more specific in order to indicate that this is about persisting **log records**. Otherwise, it might be confusing where we think that this is about persistence in general.

src/clients/Elsa.Api.Client/Resources/WorkflowDefinitions/Enums/PersistenceStrategy.cs
jdevillard: Yes , agree ! I'll do the change

jdevillard: Change done , using the two commit, I also rename the json key. 
I'll update the Client code once this PR will be done.

sfmskywalker:
@jdevillard We will merge this once you give us the green light!

jdevillard:
Ok I found the issue about yesterday demo , 

it was one case when you revert Workflow Configuration to "Default" so the value of the property was LogPersistenceMode.Default and not LogPersistenceMode.Include or LogPersistenceMode.Exclude so the information was loosed when the algo have to record or not the data.

So now I think It's ok  for me !

thx a lot !

sfmskywalker (approved):

sfmskywalker:
Awesome, great job @jdevillard . Many thanks for this!

Issue #4907, opened 2024-02-08
Workflow-Server fails to start due to incorrect Cron expression in Cron activity


platschlena:
The workflow server fails to start due to an incorrect cron expression in the cron activity. 
Through user input, the system should not be hindered from starting up.
What would be the best course of action in this situation? 
We could take the initiative to implement a fix ourselves.

sfmskywalker:
Agreed 💯 

We still have to implement user input validation on the UI holistically, but even without that, bad user input should indeed not prevent the server from starting.

One possible solution to this is to simply catch the exception currently preventing the server from starting successfully and logging a warning.

If you want to propose a PR then that'd be great. Thank you!

platschlena:
> Agreed 💯
> 
> We still have to implement user input validation on the UI holistically, but even without that, bad user input should indeed not prevent the server from starting.
> 
> One possible solution to this is to simply catch the exception currently preventing the server from starting successfully and logging a warning.
> 
> If you want to propose a PR then that'd be great. Thank you!

I created a PR with a exception handling: 

https://github.com/elsa-workflows/elsa-core/pull/4917

sfmskywalker:
With your PR merged,I will go ahead and close this issue. Thank you for your contributions @platschlena !

platschlena:
> With your PR merged, I will go ahead and close this issue. Thank you for your contributions @platschlena !

Thabk you too @sfmskywalker :)

PR #4917, opened 2024-02-09
Fix elsa-workflows/elsa-core#4907

platschlena:
This PR is the Bugfix for this Issue : https://github.com/elsa-workflows/elsa-core/issues/4907

 Catching the exception currently preventing the server from starting successfully and logging a warning when a incorrect cron expression was added to the cron activity


platschlena:
@dotnet-policy-service agree

sfmskywalker (approved):

Issue #4918, opened 2024-02-09
Cannot resolve scoped service for MongoDb


bbenameur:
Hello,
I have a stranger exception when lunching demo project.
 On the main branch only,
Project  : Elsa.Server.Web
 - Configuration services:
 ```chsarp
AddElsa(elsa =>

{
    elsa
        .UseWorkflowManagement(management =>
        {
            management.UseMongoDb();
        })
        .UseMongoDb("xxxxxxxxxxxxxxx")
        .AddWorkflowsFrom<Program>()
        .AddActivitiesFrom<Program>()
        .UseWorkflowRuntime(f => f.UseMongoDb())
        .UseWorkflowsApi(f => f.AddFastEndpointsAssembly<Program>())
        .UseRealTimeWorkflows()
        .UseHttp();
});
```
Exception:

System.InvalidOperationException : 'Cannot resolve scoped service 'MongoDB.Driver.IMongoCollection`1[Elsa.Workflows.Management.Entities.WorkflowDefinition]' from root provider.'

Source exception: Elsa.MongoDb.Modules.Management; CreateIndices.cs
![Elsa-Bug-1](https://github.com/elsa-workflows/elsa-core/assets/32399944/7c2b629b-0c48-4da1-996c-0d52c64bbeb5)

PR #4924, opened 2024-02-09
issues(4918): fix resolve scoped service for Elsa.MongoDb.Modules

bbenameur:
Releated to this issues: https://github.com/elsa-workflows/elsa-core/issues/4918

### DIFF
--- a/src/modules/Elsa.MongoDb/Modules/Identity/CreateIndices.cs
+++ b/src/modules/Elsa.MongoDb/Modules/Identity/CreateIndices.cs
@@ -8,9 +8,9 @@ namespace Elsa.MongoDb.Modules.Identity;
 
 internal class CreateIndices : IHostedService
 {
-    private readonly IServiceProvider _serviceProvider;
+    private readonly IServiceScope _serviceScope;
 
-    public CreateIndices(IServiceProvider serviceProvider) => _serviceProvider = serviceProvider;
+    public CreateIndices(IServiceProvider serviceProvider) => _serviceScope = serviceProvider.CreateScope();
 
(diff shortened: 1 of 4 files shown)


sfmskywalker:
Your PR makes sense, and I think we can get by with a single scope that we create first and then pass down to the individual "create indices" methods. What do you think?

bbenameur:
I agree this is the correct way

src/modules/Elsa.MongoDb/Modules/Identity/CreateIndices.cs
```diff
@@ -8,9 +8,9 @@ namespace Elsa.MongoDb.Modules.Identity;
 
 internal class CreateIndices : IHostedService
 {
-    private readonly IServiceProvider _serviceProvider;
+    private readonly IServiceScope _serviceScope;
```
sfmskywalker, line 11: Although I am not 100% sure, but if this hosted service remains in memory, so does this child service scope. Therefore, perhaps it is better to not store the child scope asa private field, but rather as a temporary local variable that gets disposed when done with.

E.g.

```csharp
private readonly IServiceProvider _serviceProvider;

    public CreateIndices(IServiceProvider serviceProvider) => _serviceProvider = serviceProvider;

    public Task StartAsync(CancellationToken cancellationToken)
    {
       using var scope = _serviceProvider.CreateScope();
        return Task.WhenAll(
            CreateWorkflowDefinitionIndices(scope, cancellationToken),
            CreateWorkflowInstanceIndices(scope, cancellationToken));
    }
```

sfmskywalker (approved):
Awesome 👍🏻

PR #4926, opened 2024-02-10
Set the Timer Interval default value to match DefaultValue

### LINKED ISSUE #4818: Quartz Scheduler is not working in v3
In Elsa v3.0.3, I added below code, but the Timer is never triggered.

For Hangire, we have runtime.UseHangfireBackgroundActivityScheduler(), which uses Hangfire to schedule background activities.
But I did not find a similar funciton for Quartz.

`// Use timers.
elsa.UseQuartz();
elsa.UseScheduling(scheduling => scheduling.UseQuartzScheduler());`

mohdali:
Should fix #4818


sfmskywalker:
Thanks @mohdali !

mohdali:
> Thanks @mohdali , may I know why this would fix #4818 ? I have set the interval for the timer in the Studio, seems the default value will not be used.

The default value of `null` will cause Quartz scheduler to fail to start. Currently Elsa studio upates the value on change only. So even if you manually edit the value but kept it as 00:01:00 it will still be treated as null.

Otherwise, Quartz scheduling should work just fine.

PR #4941, opened 2024-02-14
Introduced messaging to running instances

raymonddenhaan:
This feature will allow communication across nodes in a multi-node environment. Enabling features such as cancelling dispatched workflows and caching implementations.


src/modules/Elsa.Hosting.Management/HostedServices/InstanceHeartbeatService.cs
```diff
@@ -0,0 +1,66 @@
...
+    private void WriteHeartbeat(object? state)
+    {
+        _ = Task.Run(async () => await WriteHeartbeatAsync());
+    }
+    
+    private async Task WriteHeartbeatAsync()
+    {
+        using var scope = _serviceProvider.CreateScope();
+        
+        var instanceNameRetriever = scope.ServiceProvider.GetRequiredService<IInstanceNameRetriever>();
```
sfmskywalker, line 56: Although `IInstanceNameRetriever` is a descriptive name, more conventional would be `IInstanceNameProvider` or `IInstanceNameAccessor`. Obviously these are subject to personal preference, but it would als be more in line with similar other services, such as `IHttpContextAccessor`.

src/modules/Elsa.Hosting.Management/Notifications/InstanceDeactivated.cs
```diff
@@ -0,0 +1,5 @@
+using Elsa.Mediator.Contracts;
+
+namespace Elsa.Hosting.Management.Notifications;
+
+public record InstanceDeactivated(string InstanceName) : INotification;
```
sfmskywalker, line 5: As a matter of personal preference, I think that `HeartbeatTimedOut` would has a better ring to it, and still makes sense since this event means that a given instance failed to send a heartbeat (i.e. it exceeded a certain time out threshold).

src/modules/Elsa.Hosting.Management/Options/HeartbeatSettings.cs
```diff
@@ -0,0 +1,7 @@
+namespace Elsa.Hosting.Management.Options;
+
+public class HeartbeatSettings
```
sfmskywalker, line 3: Although I prefer the use of the "Settings" suffix, everywhere else we use "Options" as the suffix - a choice that I regret in hindsight. On the one hand I would like to keep using "Settings" as introduced here, but on the other hand, it might be better to remain consistent and use "Options", and as a separate refactoring task, rename everything from ...Options to ...Settings. What do you think?

src/modules/Elsa.Hosting.Management/Options/HeartbeatSettings.cs
```diff
@@ -0,0 +1,7 @@
...
+{
+    public TimeSpan InstanceHeartbeatRhythm { get; set; } = TimeSpan.FromMinutes(1);
```
sfmskywalker, line 5: Although this is very descriptive, it has an unconventional sound to it. `HeartbeatInterval` sounds nicer and is simpler, too.

src/modules/Elsa.Hosting.Management/Options/HeartbeatSettings.cs
```diff
@@ -0,0 +1,7 @@
...
+    public TimeSpan InstanceDeactivatedPeriod { get; set; } = TimeSpan.FromHours(1);
```
sfmskywalker, line 6: Although this is very descriptive, it also sounds unconventional. `HeartbeatTimeout` sounds better because we're not talking about a _periodic_ event, but a _threshold_ and it's more concise, too. Plus, it has a nice coupling with the `HeartbeatInterval` name:

- HeartbeatInterval
- HeartbeatTimeout

src/modules/Elsa.MassTransit.AzureServiceBus/Features/AzureServiceBusFeature.cs
```diff
@@ -31,18 +53,81 @@ public override void Configure()
         {
             massTransitFeature.BusConfigurator = configure =>
             {
+                var consumers = massTransitFeature.GetConsumers().ToList();
+                var shortLivedConsumers = consumers
```
sfmskywalker, line 57: The term "temporary" or even "temporaryQueue" seems more concise.
E.g. `temporaryConsumers` or `temporaryQueueConsumers`. The latter is also more correct, given that it's not the consumers that are short-live/temporary, but the queues are.

src/modules/Elsa.MassTransit.AzureServiceBus/Features/AzureServiceBusFeature.cs
```diff
@@ -31,18 +53,81 @@ public override void Configure()
...
+
                     if (ConnectionString != null) 
                         serviceBus.Host(ConnectionString);
-                    
                     serviceBus.UseServiceBusMessageScheduler();
                     ConfigureServiceBus?.Invoke(serviceBus);
+
+                    foreach (var consumer in shortLivedConsumers)
+                    {
+                        serviceBus.ReceiveEndpoint($"Elsa-{instanceNameRetriever.GetName()}-{consumer.Name}", configurator =>
```
sfmskywalker, line 77: If this name becomes too long for subscriptions, we could consider generating a hash of fixed length. Or, a part of the endpoint could be hashed.

src/modules/Elsa.MassTransit.AzureServiceBus/Features/AzureServiceBusFeature.cs
```diff
@@ -31,18 +53,81 @@ public override void Configure()
...
+
+        Services.AddSingleton(new MessageTopologyProvider(subscriptionTopology));
+    }
+
+    /// <inheritdoc />
+    public override void Apply()
+    {
+        Services.Configure(AzureServiceBusOptions);
+        Services.AddScoped(ServiceBusAdministrationClientFactory);
+        Services.AddNotificationHandler<OrphanedSubscriptionRemover>();
```
sfmskywalker, line 124: When naming event/notification handlers, we either use the form `HeartbeatTimedOutHandler`, `OrphanedSubscriptionHandler`, `RemoveOrphanedSubscription` or even `RemoveOrphanedSubscriptionHandler`.

The form `OrphanedSubscriptionRemover` is used more commonly for services that represent a thing, while the other forms are more commonly used for classes that represent a handler of an event.

You could have a service that is called `OrphanedSubscriptionRemover` that is then used by a handler - if we needed this functionality to be reusable, for example.

src/modules/Elsa.MassTransit.RabbitMq/Features/RabbitMqServiceBusFeature.cs
```diff
@@ -32,18 +37,40 @@ public RabbitMqServiceBusFeature(IModule module) : base(module)
...
+                            configurator.QueueExpiration = options.ShortTermQueueLifetime ?? TimeSpan.FromHours(1);
+                            configurator.ConcurrentMessageLimit = options.ConcurrentMessageLimit;
+                            configurator.ConfigureConsumer<DispatchCancelWorkflowsRequestConsumer>(context);
+                        });
+                    }
 
-                serviceBus.ConfigureEndpoints(context, new KebabCaseEndpointNameFormatter("Elsa", false));
-            });
-        };
+                    serviceBus.ConfigureEndpoints(context, new KebabCaseEndpointNameFormatter("Elsa", false));
```
sfmskywalker, line 70: Some of these blocks might be moved to an extension - both RabbitMQ and Azure Service Bus configurators share a common ancestor. I did [something similar in the dispatch-channels branch](https://github.com/elsa-workflows/elsa-core/blob/feature/dispatch-channels/src/modules/Elsa.MassTransit/Extensions/ServiceBusChannelsConfigurationExtensions.cs) - that extension is used in both RabbitMQ and ASB configuration for MT.

src/modules/Elsa.MassTransit/Extensions/MassTransitFeatureExtensions.cs
```diff
@@ -18,27 +18,28 @@ public static class MassTransitFeatureExtensions
     /// <summary>
     /// Registers the specified type for MassTransit service bus consumer discovery.
     /// </summary>
-    public static MassTransitFeature AddConsumer<T>(this MassTransitFeature feature) where T : IConsumer => feature.AddConsumer(typeof(T));
+    public static MassTransitFeature AddConsumer<T>(this MassTransitFeature feature, string? name, bool isTemporary) where T : IConsumer => feature.AddConsumer(typeof(T), name, isTemporary);
```
sfmskywalker, line 21: Here you are using "temporary" instead of "short lived" 👍🏻

src/modules/Elsa.MassTransit/Options/MassTransitWorkflowDispatcherOptions.cs
```diff
@@ -5,6 +5,8 @@ namespace Elsa.MassTransit.Options;
 /// Provides options to the <see cref="DispatchWorkflowRequestConsumerDefinition"/>
 public class MassTransitWorkflowDispatcherOptions
 {
+    /// The TTL of queues that are seen as short lived (typically queues that are created per running instance).
+    public TimeSpan? ShortTermQueueLifetime { get; set; }
```
sfmskywalker, line 9: Some alternative (subjectively, better) options:

- `TemporaryQueueTtl`
- `TempQueueTtl`
- `TemporaryQueueLifetime`
- `TempQueueLifetime`

src/modules/Elsa.Workflows.Core/Contracts/IInstanceNameRetriever.cs
```diff
@@ -0,0 +1,12 @@
+namespace Elsa.Workflows.Contracts;
+
+/// <summary>
+/// Retrieves a name of the current instance.
+/// </summary>
+public interface IInstanceNameRetriever
```
sfmskywalker, line 6: This should be moved to a different layer, probably Runtime or even the new Hosting.
Rationale: Application instance names belong to the domain of infrastructure. Workflows Core is about workflows and activities.

sfmskywalker, line 6: Although with enough context, "instance" should be fine, but perhaps to make it extra clear, let's rename this to `IApplicationInstanceNameAccessor`.

src/modules/Elsa.Workflows.Runtime/Contracts/IKeyValueStore.cs
```diff
@@ -0,0 +1,32 @@
+using System.Linq.Expressions;
+using Elsa.Workflows.Runtime.Entities;
+using Elsa.Workflows.Runtime.Models;
+
+namespace Elsa.Workflows.Runtime.Contracts;
+
+/// <summary>
+/// Store that holds key value entities not fit to store in specific stores.
+/// </summary>
+public interface IKeyValueStore
```
sfmskywalker, line 10: Unless there's a technical reason this store must live in the Runtime layer, we should move this to its own module (e.g. `Elsa.KeyValues`). Other modules, such as `Elsa.Hosting.Management` are free to take on a dependency to this module.

sfmskywalker (changes requested):

src/bundles/Elsa.ServerAndStudio.Web/appsettings.json
```diff
@@ -61,6 +61,13 @@
       }
     ]
   },
+  "Heartbeat": {
+    "InstanceHeartbeatRhythm": "00:00:01:00",
+    "InstanceDeactivatedPeriod": "00:00:05:00",
+  },
+  "MassTransit": {
+    "ShortTermQueueLifetime": "00:00:05:00"
```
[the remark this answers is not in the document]
raymonddenhaan: I first named it temporary (as you found in a piece which I had not renamed), thought short-lived might be better.
I'll rename it to temporary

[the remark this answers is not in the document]
sfmskywalker: I agree with your initial instinct 😄

src/modules/Elsa.Hosting.Management/Options/HeartbeatSettings.cs
raymonddenhaan: I used settings since I saw another file use it as well. Since the vast majority is named Options let's keep it like that for now.
Although I do like the suffix settings a bit better I don't think it warrants a rename for all options. Especially not in a PR like this.

src/modules/Elsa.MassTransit/Options/MassTransitWorkflowDispatcherOptions.cs
raymonddenhaan: Changed it to TemporaryQueueTtl

src/modules/Elsa.Workflows.Runtime/Contracts/IKeyValueStore.cs
raymonddenhaan: While I agree that the Runtime project is not the right place, I'm not too sure about it having its own project. That could lead to a lot of very small projects in the future.

sfmskywalker (approved):

sfmskywalker (approved):

PR #4949, opened 2024-02-15
Implement dispatch channels

### LINKED ISSUE #4923: Workflow Dispatcher Service Bus Integration - Specific Channels
We need the ability to configure a specific channel for a dispatched workflows and all of its children.

### Tasks
- [x] Implement an options class to allow the application to configure available channels via configuration.
- [x] Update MassTransit transports (Memory, ASB and RabbitMQ) to create endpoints for each channel and associate them with the DispatchWorkflowRequestConsumer class.
- [x] Update DispatchWorkflow and BulkDispatchWorkflows activities to provide an option to the user from which they can select the desired channel (must be optional; the user is not required to select any channel and should be empty by default).

sfmskywalker:
This PR adds a feature that allows you to specify what channel to dispatch workflow execution requests to. 

Closes #4923


src/modules/Elsa.Workflows.Api/Endpoints/WorkflowDefinitions/Dispatch/Endpoint.cs
```diff
@@ -48,22 +38,23 @@ public override async Task HandleAsync(Request request, CancellationToken cancel
...
-        var correlationId = request.CorrelationId;
-        var triggerActivityId = request.TriggerActivityId;
-        var input = (IDictionary<string, object>?)request.Input;
+        var instanceId = request.InstanceId ?? identityGenerator.GenerateId();
         var dispatchRequest = new DispatchWorkflowDefinitionRequest
         {
             DefinitionId = definitionId,
             VersionOptions = versionOptions,
-            Input = input,
+            Input = (IDictionary<string, object>?)request.Input,
```
raymonddenhaan, line 46: Since this is an endpoint we should `handle` values besides the expected dictionary as well.

src/modules/Elsa.Workflows.Runtime/Requests/DispatchWorkflowDefinitionRequest.cs
```diff
@@ -60,5 +60,5 @@ public DispatchWorkflowDefinitionRequest(string definitionId, VersionOptions ver
     /// <summary>
     /// The ID of the activity that triggered the workflow.
     /// </summary>
-    public string? TriggerActivityId { get; init; }
+    public string? TriggerActivityId { get; set; }
```
raymonddenhaan, line 63: Why should we want to change this after initialization?

raymonddenhaan (approved):

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowDefinitions/Dispatch/Endpoint.cs
sfmskywalker: Agreed. In this case, since the dispatch request only supports dictionaries, this endpoint should validate the input type, and if it's anything but a dictionary, we respond with a 400 status code. Agreed?

src/modules/Elsa.Workflows.Runtime/Requests/DispatchWorkflowDefinitionRequest.cs
sfmskywalker: The rationale here is that this type has default constructor and a constructor that accepts the two required members.

This allows client code to instantiate a request object and provide the two required values via the constructor, without being forced to initialize the remaining properties, which are optional, via the initializer. For example, it allows for this usage pattern:

```csharp
var request = new DispatchWorkflowDefinitionRequest("my definition ID", VersionOptions.FromVersion(1));

// Configure other props.
request.TriggerActivityId = "activity-1";
```

In contrast, if the optional properties must be set during object init, the code would **have** to look like this:

```csharp
var request = new DispatchWorkflowDefinitionRequest("my definition ID", VersionOptions.FromVersion(1))
{
   TriggerActivityId = "activity-1"
}
```

Or this:

```csharp
var request = new DispatchWorkflowDefinitionRequest
{
   DefinitionId = "my definition ID",
   VersionOptions.FromVersion(1),
   TriggerActivityId = "activity-1"
}
```

All these options are fine, but subject to user preference.

raymonddenhaan: I would prefer it to still have it on init so the value can not be changed after initialization. This id will not change after the request object has been made. I those cases I prefer to make the properties immutable, regardless if they are optional.

PR #4957, opened 2024-02-18
Enhanced Workflow Instance Filters

sfmskywalker:
This PR enhances the ability to filter workflow instance search results. I includes the following filters:

- Has Incidents
- ID, Correlation ID, Definition ID, Name
- Date ranges: Created, Updated, Finished

The following filters already existed:

- Status
- Sub Status
- Workflow Definitions

### DIFF
--- a/src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
+++ b/src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
@@ -45,6 +45,11 @@ public class ListWorkflowInstancesRequest
     public int? Version { get; set; }
 
     /// <summary>
+    /// Gets or sets whether the workflow instances have incidents or not.
+    /// </summary>
+    public bool? HasIncidents { get; set; }
+
+    /// <summary>
     /// Gets or sets the status.
     /// </summary>
     public WorkflowStatus? Status { get; set; }
--- a/src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
+++ b/src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
@@ -13,64 +13,69 @@ public class ListWorkflowInstancesRequest
     /// Gets or sets the page number.
     /// </summary>
     public int? Page { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the page size.
     /// </summary>
     public int? PageSize { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the search term.
     /// </summary>
     public string? SearchTerm { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the workflow definition id.
     /// </summary>
     public string? DefinitionId { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the workflow definition ids.
     /// </summary>
-    [Query(CollectionFormat.Multi)] public ICollection<string>? DefinitionIds { get; set; }
-    
+    public ICollection<string>? DefinitionIds { get; set; }
+
     /// <summary>
     /// Gets or sets the correlation id.
     /// </summary>
     public string? CorrelationId { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the version.
     /// </summary>
     public int? Version { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the status.
     /// </summary>
     public WorkflowStatus? Status { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the statuses.
     /// </summary>
-    [Query(CollectionFormat.Multi)] public ICollection<WorkflowStatus>? Statuses { get; set; }
-    
+    public ICollection<WorkflowStatus>? Statuses { get; set; }
+
     /// <summary>
     /// Gets or sets the sub status.
     /// </summary>
     public WorkflowSubStatus? SubStatus { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the sub statuses.
     /// </summary>
-    [Query(CollectionFormat.Multi)] public ICollection<WorkflowSubStatus>? SubStatuses { get; set; }
+    public ICollection<WorkflowSubStatus>? SubStatuses { get; set; }
 
     /// <summary>
     /// Gets or sets the key to order by.
     /// </summary>
     public OrderByWorkflowInstance? OrderBy { get; set; }
-    
+
     /// <summary>
     /// Gets or sets the order direction.
     /// </summary>
     public OrderDirection? OrderDirection { get; set; }
+
+    /// <summary>
+    /// Gets or sets the timestamp filters.
+    /// </summary>
+    public ICollection<TimestampFilter>? TimestampFilters { get; set; }
 }
--- a/src/clients/Elsa.Api.Client/Shared/Models/TimestampFilter.cs
+++ b/src/clients/Elsa.Api.Client/Shared/Models/TimestampFilter.cs
@@ -1,0 +1,26 @@
+using Elsa.Api.Client.Shared.Enums;
+using JetBrains.Annotations;
+
+namespace Elsa.Api.Client.Shared.Models;
+
+/// <summary>
+/// Represents a timestamp filter used for filtering data based on a specified timestamp column and operator.
+/// </summary>
+[UsedImplicitly]
+public class TimestampFilter
+{
+    /// <summary>
+    /// Gets or sets the column to filter by.
+    /// </summary>
+    public string Column { get; set; } = default!;
+
+    /// <summary>
+    /// Gets or sets the operator to use for filtering.
+    /// </summary>
+    public TimestampFilterOperator Operator { get; set; }
+
+    /// <summary>
+    /// Gets or sets the timestamp to filter by.
+    /// </summary>
+    public DateTimeOffset Timestamp { get; set; }
+}
--- a/src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
+++ b/src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
@@ -14,8 +14,8 @@ internal class List(IWorkflowInstanceStore store) : ElsaEndpoint<Request, Respon
 {
     public override void Configure()
     {
-        Get("/workflow-instances");
-        Post("/workflow-instances");
+        Verbs(FastEndpoints.Http.GET, FastEndpoints.Http.POST);
+        Routes("/workflow-instances");
         ConfigurePermissions("read:workflow-instances");
     }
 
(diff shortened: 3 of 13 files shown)


src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
```diff
@@ -13,64 +13,74 @@ public class ListWorkflowInstancesRequest
...
     /// <summary>
     /// Gets or sets the version.
     /// </summary>
     public int? Version { get; set; }
-    
+
+    /// <summary>
+    /// Gets or sets whether the workflow instances have incidents or not.
+    /// </summary>
+    public bool? HasIncidents { get; set; }
```
bbenameur, line 50: I'm not sur but we can have instances incidents just with status, failed or other ?

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
```diff
@@ -33,15 +27,15 @@ public override async Task<Response> ExecuteAsync(Request request, CancellationT
...
             WorkflowStatus = request.Status,
             WorkflowSubStatus = request.SubStatus,
-            WorkflowStatuses = request.Statuses?.Select(Enum.Parse<WorkflowStatus>).ToList(),
-            WorkflowSubStatuses = request.SubStatuses?.Select(Enum.Parse<WorkflowSubStatus>).ToList(),
-            ToCreatedAt = request.ToCreatedAt,
-            FromCreatedAt = request.FromCreatedAt,
+            WorkflowStatuses = request.Statuses?.Any() == true ? request.Statuses : null,
+            WorkflowSubStatuses = request.SubStatuses?.Any() == true ? request.SubStatuses : null,
+            HasIncidents = request.HasIncidents,
+            TimestampFilters = request.TimestampFilters?.Any() == true ? request.TimestampFilters : null,
```
bbenameur, line 38: request.TimestampFilters?.Any() is enough not need to request.TimestampFilters?.Any() == true

src/clients/Elsa.Api.Client/Shared/Models/TimestampFilter.cs
```diff
@@ -0,0 +1,26 @@
...
+/// <summary>
+/// Represents a timestamp filter used for filtering data based on a specified timestamp column and operator.
+/// </summary>
+[UsedImplicitly]
+public class TimestampFilter
+{
+    /// <summary>
+    /// Gets or sets the column to filter by.
+    /// </summary>
+    public string Column { get; set; } = default!;
```
bbenameur, line 15: Should not be an enum like 
{
CreatedAt,
UpdatedAt,
FinishedAt,
}

or add a validator ?

src/modules/Elsa.Workflows.Management/Filters/WorkflowInstanceFilter.cs
```diff
@@ -106,19 +111,47 @@ public IQueryable<WorkflowInstance> Apply(IQueryable<WorkflowInstance> query)
...
         if (filter.WorkflowSubStatuses != null) query = query.Where(x => filter.WorkflowSubStatuses.Contains(x.SubStatus));
-        if (filter.ToCreatedAt is not null) query = query.Where(x => filter.ToCreatedAt >= x.CreatedAt);
-        if (filter.FromCreatedAt is not null) query = query.Where(x => filter.FromCreatedAt <= x.CreatedAt);
+        if (filter.HasIncidents != null) query = filter.HasIncidents == true ? query.Where(x => x.IncidentCount > 0) : query.Where(x => x.IncidentCount == 0);
+
+        if (TimestampFilters != null)
+        {
+            foreach (TimestampFilter timestampFilter in TimestampFilters)
+            {
+                var column = timestampFilter.Column;
```
bbenameur, line 120: One of the properties of **TimestampFilter** can be null and have a bug for Eg:
with : 
?TimestampFilters=[{ "CreatedAt", "Operator": 1, "Timestamp": "2024-02-16T00:00:00"}]

src/modules/Elsa.Workflows.Management/Models/TimestampFilter.cs
```diff
@@ -0,0 +1,26 @@
...
+/// <summary>
+/// Represents a timestamp filter used for filtering data based on a specified timestamp column and operator.
+/// </summary>
+[UsedImplicitly]
+public class TimestampFilter
+{
+    /// <summary>
+    /// Gets or sets the column to filter by.
+    /// </summary>
+    public string Column { get; set; } = default!;
```
bbenameur, line 15:
```suggestion
    public required string Column { get; set; } = default!;
```

src/modules/Elsa.Workflows.Management/Models/TimestampFilter.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// <summary>
+    /// Gets or sets the operator to use for filtering.
+    /// </summary>
+    public TimestampFilterOperator Operator { get; set; }
```
bbenameur, line 20:
```suggestion
    public required TimestampFilterOperator Operator { get; set; }
```

src/modules/Elsa.Workflows.Management/Models/TimestampFilter.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// <summary>
+    /// Gets or sets the timestamp to filter by.
+    /// </summary>
+    public DateTimeOffset Timestamp { get; set; }
```
bbenameur, line 25:
```suggestion
    public required DateTimeOffset Timestamp { get; set; }
```

src/modules/Elsa.Workflows.Management/Models/TimestampFilter.cs
bbenameur: required if for:
if TimestampFilter not null should have all properties not null

src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Requests/ListWorkflowInstancesRequest.cs
sfmskywalker: It's a good question. Depending on the incident handling strategy applied to the workflow definition, it is possible that a workflow has incidents (faults), without the workflow itself being in the Faulted state.

There are currently two incident handling strategies:

- Fault: whenever an activity faults, the entire workflow faults. This is the default strategy.
- Continue with Incidents: keep the workflow in the Running state, even when an activity faults.

If we only had the first strategy, then you are correct and we would not need the "Has Incidents" filter.
But, since we also support the second strategy (and potentially others), the workflow status can remain to be Running, but with incidents; this is why we added this filter.

src/clients/Elsa.Api.Client/Shared/Models/TimestampFilter.cs
sfmskywalker: I considered adding an enum instead of a string, but figured that we might keep the `TimestampFilter` more general purpose, instead of only applying to the `WorkflowInstance` entity.

But I agree with your second suggestion and add a validator at the WorkflowInstance List API endpoint and return a 400 Bad Request in case the client sends an invalid column name (instead of a 500 internal server error that would happen right now) 👍🏻

src/modules/Elsa.Workflows.Management/Models/TimestampFilter.cs
sfmskywalker: I wasn't aware of this capability - very nice 👍🏻

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
sfmskywalker: I wish you were right, but unfortunately that doesn't work on nullable booleans:

image

src/modules/Elsa.Workflows.Management/Filters/WorkflowInstanceFilter.cs
sfmskywalker: Should be good now with input validation 👍🏻

sfmskywalker:
@bbenameur Thanks for the code review, I appreciate it!

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
```diff
@@ -6,65 +6,106 @@
...
         };
 
         var summaries = await FindAsync(request, filter, pageArgs, cancellationToken);
-        return new Response(summaries.Items, summaries.TotalCount);
+        var response = new Response(summaries.Items, summaries.TotalCount);
+        await SendOkAsync(response, cancellationToken);
+    }
+
+    private async Task<bool> ValidateInputAsync(Request request, CancellationToken cancellationToken)
```
raymonddenhaan, line 53: Is there a specific reason you are not using the validation engine of fastendpoints?

src/modules/Elsa.Workflows.Management/Filters/WorkflowInstanceFilter.cs
```diff
@@ -106,19 +111,47 @@ public IQueryable<WorkflowInstance> Apply(IQueryable<WorkflowInstance> query)
...
         if (!string.IsNullOrWhiteSpace(searchTerm))
         {
             query =
                 from instance in query
                 where instance.Name!.Contains(searchTerm)
-                      || instance.DefinitionVersionId.Equals(searchTerm)
-                      || instance.Id.Equals(searchTerm)
-                      || instance.DefinitionId.Equals(searchTerm)
-                      || instance.CorrelationId!.Equals(searchTerm)
+                      || instance.DefinitionVersionId.Contains(searchTerm)
```
raymonddenhaan, line 151: Should we only be able to search on an (part of) id or do we also want to be able to search on the name of the workflow definition?

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
sfmskywalker: I wasn't sure how to do it for the TimestampFilters collection, but maybe I missed something obvious. If you have a suggestion then it'd be great to move it into its own DTO validator 👍🏻

src/modules/Elsa.Workflows.Management/Filters/WorkflowInstanceFilter.cs
sfmskywalker: It would be nice to be able to also search on a part of the workflow definition name, but I wasn't sure if it is worth the added complexity to achieve that. I think that it would require to do a preliminary search on the workflow definitions, then selecting the matching definition IDs, and adding those to the workflow instance filter. If you think it is worthwhile, we can do it. It would certainly offer a very nice experience to the user.

raymonddenhaan: A version where we can reduce the complexity is where we can use the search in the workflow definition selection box and use that to select the definitions we are interested in.

sfmskywalker: Aha - so basically grab a set of matching workflow definition IDs when the user enters a search term in the text field, and including that as part of the request. That's pretty smart! 😎 I will send another PR with that improvement 👍🏻

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/List/Endpoint.cs
raymonddenhaan: You can use custom functions for validation as well. I plan to have a PR up later today in which I use this

sfmskywalker: I see - I will give that a try, thank you 👍🏻

PR #4965, opened 2024-02-19
Added workflow instance and bulk cancellation to the UI

raymonddenhaan:
With this change it is possible to cancel a single workflow, select a number of workflows to cancel and cancel all instances of a definition.


src/modules/Elsa.ProtoActor/Services/ProtoActorWorkflowRuntime.cs
```diff
@@ -249,10 +249,51 @@ public async Task<WorkflowExecutionResult> ExecuteWorkflowAsync(WorkflowMatch ma
...
+        if (instance is null)
+            return new CancellationResult(false, FailureReason.NotFound);
+        
         var client = _cluster.GetNamedWorkflowGrain(workflowInstanceId);
-        await client.Cancel(cancellationToken);
+        var result = await client.Cancel(cancellationToken);
+        return new CancellationResult(result?.Result ?? false);
+    }
+
+    public async Task<int> CancelWorkflowAsync(IEnumerable<string> workflowInstanceIds, CancellationToken cancellationToken = default)
```
sfmskywalker, line 268: We should probably rename this to `CancelWorkflowsAsync` (plural), given that it supports the cancellation of a list of them.

src/modules/Elsa.ProtoActor/Services/ProtoActorWorkflowRuntime.cs
```diff
@@ -249,10 +249,51 @@ public async Task<WorkflowExecutionResult> ExecuteWorkflowAsync(WorkflowMatch ma
...
+    public async Task<int> CancelWorkflowByDefinitionAsync(string definitionId, VersionOptions versionOptions, CancellationToken cancellationToken = default)
+    {
+        // Shouldn't we get possible multiple definitions here?
+        var workflowDefinition = await _workflowDefinitionService.FindAsync(definitionId, versionOptions, cancellationToken);
+        if (workflowDefinition is null)
+            return 0;
+
+        var result = await CancelWorkflowByDefinitionVersionAsync(workflowDefinition.Id, cancellationToken);
+        return result;
     }
```
sfmskywalker, line 297: Since these 3 methods seem to be mostly conveniences to client code, it seems to me that it would be better to extract these as extension methods on IWorkflowRuntime, so that the DefaultWorkflowRuntime automatically gets these methods too without having to implement them twice.

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/Cancel/Endpoint.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        _workflowRuntime = workflowRuntime;
+    }
+
+    public override void Configure()
+    {
+        Post("/cancel/workflow-instances/{id}");
+        ConfigurePermissions("cancel:workflow-instances");
+        //Allows for post with empty body
+        Description(x => x.Accepts<Request>("*/*"), clearDefaults: true);
```
sfmskywalker, line 24: Nice find 👍🏻

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/Cancel/Endpoint.cs
```diff
@@ -0,0 +1,45 @@
...
+        else if (result.Reason == FailureReason.NotFound)
+            await SendNotFoundAsync(cancellationToken);
+        else
+        {
+            if (result.Reason == FailureReason.InvalidState)
+                AddError("Instance is in finished state.");
+            else
+                AddError("Unable to access instance.");
+            
+            await SendErrorsAsync(StatusCodes.Status409Conflict, cancellationToken);
```
sfmskywalker, line 42: I am having second thoughts if we should return a bad request response. Let's consider the following scenario:

The user decides that they want to abort a bulk operation and cancel everything. Some of the dispatched workflows may have already completes. The user doesn't want their API call to fail because of that.

Perhaps it makes sense to just respond with a "report" (as a simple JSON model) of what got cancelled and what was already cancelled.

What do you think?

sfmskywalker (changes requested):

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/Cancel/Endpoint.cs
raymonddenhaan: The bad request is only returned on this endpoint (for a single instance cancellation. If the given instance can be found it will give a succeeded result regardless what happens with the child workflows.
In the case of multiple instance cancellation only the amount of workflows that should be cancelled is reported (no bad requests). The only way to get a non 200 on that endpoint is if the validation fails or if there is an uncaught exception.

sfmskywalker: Gotcha. then I agree 👍🏻

src/clients/Elsa.Api.Client/Resources/WorkflowInstances/Contracts/IWorkflowInstancesApi.cs
```diff
@@ -62,6 +62,22 @@ public interface IWorkflowInstancesApi
...
+    /// <param name="cancellationToken">The cancellation token.</param>
+    [Post("/cancel/workflow-instances/{id}")]
+    Task CancelAsync(string id, CancellationToken cancellationToken = default);
+    
+    /// <summary>
+    /// Cancels a list of workflow instances.
+    /// </summary>
+    /// <param name="request">The request containing the selection of workflow instances to cancel.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    [Post("/bulk-actions/cancel/workflow-instances")]
```
bbenameur, line 78: I thinks that is not important to specefie that is a bulk action,
[Post("/cancel/workflow-instances")] is enough.

raymonddenhaan: Agreed, I'll remove the bulk-actions part from the url (and other existing urls)

sfmskywalker (approved):

PR #4978, opened 2024-02-22
Fix incorrect comment information in EFCoreWorkflowInboxMessageStore

faith0831:


faith0831:
@dotnet-policy-service agree

Issue #4810, opened 2024-01-19
An exception occurred while executing CleanupExpiredMessages with MySQL database in Elsa3


faith0831:
Here is the exception information:

 Elsa.Workflows.Runtime.HostedServices.WorkflowInboxCleanupHostedService[0]
      Entering expired workflow inbox messages cleanup service loop
fail: Microsoft.EntityFrameworkCore.Database.Command[20102]
      Failed executing DbCommand (1ms) [Parameters=[@__now_0='?' (DbType = DateTimeOffset), @__p_2='?' (DbType = Int32), @__p_1='?' (DbType = Int32)], CommandType='Text', CommandTimeout='30']
      DELETE `w`
      FROM `WorkflowInboxMessages` AS `w`
      WHERE `w`.`Id` IN (
          SELECT `w0`.`Id`
          FROM `WorkflowInboxMessages` AS `w0`
          WHERE `w0`.`ExpiresAt` <= @__now_0
          LIMIT @__p_2 OFFSET @__p_1
      )
fail: Elsa.Workflows.Runtime.HostedServices.WorkflowInboxCleanupHostedService[0]
      An error occurred while cleaning up expired workflow inbox messages
      MySqlConnector.MySqlException (0x80004005): This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'
         at MySqlConnector.Core.ServerSession.ReceiveReplyAsync(IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/Core/ServerSession.cs:line 936
         at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in /_/src/MySqlConnector/Core/ResultSet.cs:line 37
         at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 130
         at MySqlConnector.MySqlDataReader.InitAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, Activity activity, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 483
[...]

PR #4980, opened 2024-02-22
Fix CleanupExpiredMessages Error (#4810)

faith0831:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/4980)

### DIFF
--- a/src/modules/Elsa.EntityFrameworkCore/Modules/Runtime/WorkflowInboxStore.cs
+++ b/src/modules/Elsa.EntityFrameworkCore/Modules/Runtime/WorkflowInboxStore.cs
@@ -49,7 +49,7 @@ public class EFCoreWorkflowInboxMessageStore : IWorkflowInboxMessageStore
     /// <inheritdoc />
     public async ValueTask<long> DeleteManyAsync(WorkflowInboxMessageFilter filter, PageArgs? pageArgs = default, CancellationToken cancellationToken = default)
     {
-        return await _store.DeleteWhereAsync(q => Paginate(filter.Apply(q, _systemClock.UtcNow), pageArgs), cancellationToken);
+        return await _store.DeleteWhereAsync(q => filter.Apply(q, _systemClock.UtcNow), cancellationToken);
     }
 
     private ValueTask OnSaveAsync(RuntimeElsaDbContext dbContext, WorkflowInboxMessage entity, CancellationToken cancellationToken)


sfmskywalker:
Hello @faith0831 , thank you for submitting a PR. Can you explain what the issue is and how this PR fixes that? I noticed that it removes the use of the pagination args, which might have other consequences.

mohdali:
> Hello @faith0831 , thank you for submitting a PR. Can you explain what the issue is and how this PR fixes that? I noticed that it removes the use of the pagination args, which might have other consequences.

It is related to #4810 for MySQL as LIMIT is not supported in sub queries. However, I'm not sure if removing the pagination is the way to go as it seems to be possible to replace the subquery with a LEFT JOIN. I didn't get the chance to test that though.

@faith0831 is it possible to use a LEFT JOIN instead of getting rid of pagination completely?

sfmskywalker:
@mohdali I see, thanks for the clarification. We shouldn't drop paging support because it might negatively impact other users that use other database servers and where doing a batched deletion of workflow inbox messages is preferred (in case of high numbers of records).

Additionally, if the paging SQL clauses cause an issue for MySQL for the WorkflowInbox table, it will also be an issue for all other DB tables, which means that this PR might "fix" one affected area, it will not fix the root issue.

faith0831:
> > Hello @faith0831 , thank you for submitting a PR. Can you explain what the issue is and how this PR fixes that? I noticed that it removes the use of the pagination args, which might have other consequences.您好，感谢您提交 PR。你能解释一下问题是什么以及这个 PR 是如何解决这个问题的吗？我注意到它删除了分页参数的使用，这可能会产生其他后果。
> 
> @sfmskywalker It is related to #4810 for MySQL as LIMIT is not supported in sub queries. However, I'm not sure if removing the pagination is the way to go as it seems to be possible to replace the subquery with a LEFT JOIN. I didn't get the chance to test that though.它与 MySQL 的 #4810 相关，因为子查询不支持 LIMIT。但是，我不确定删除分页是否是要走的路，因为似乎可以用 LEFT JOIN 替换子查询。不过，我没有机会测试它。
> 
> @faith0831 is it possible to use a LEFT JOIN instead of getting rid of pagination completely?是否可以使用 LEFT JOIN 而不是完全摆脱分页？

@mohdali  Thank you for your assistance, I will continue to try to solve this problem.

faith0831:
> @mohdali I see, thanks for the clarification. We shouldn't drop paging support because it might negatively impact other users that use other database servers and where doing a batched deletion of workflow inbox messages is preferred (in case of high numbers of records).
> 
> Additionally, if the paging SQL clauses cause an issue for MySQL for the WorkflowInbox table, it will also be an issue for all other DB tables, which means that this PR might "fix" one affected area, it will not fix the root issue.

Thank you for your reply @sfmskywalker, this PR indeed has the issues you mentioned.

sfmskywalker:
@faith0831 @mohdali Should we keep this PR open, change it to Draft, or abandon?

mohdali:
@sfmskywalker we need to change how the MySql query for clearing the inbox is executed. I didn't get the time to check the possible solution. I think this PR can be changed to draft.

mohdali:
@sfmskywalker In order to resolve this issue we need to change the `DeleteManyAsync` query specifically for MySQL to avoid using pagination. I think it is still possible to batch the deletion in MySQL but needs to be done using a subquery instead.

In that case, what is the recommended approach? Show we create a new `IWorkflowInboMessageStore` in `Elsa.EntitiyFrameworkCore.MySql` and let it hide the default store? We only need to add that store if the Workflow Inbox feature is enabled though (not sure what is the correct pattern here).

I think in general there will be some exceptions like this depending on each DB capabilities, we need to have a clear pattern to handle those cases.

raymonddenhaan:
Closing this PR since the `IWorkflowInboxMessageStore` is no longer used in the 3.3 preview

Issue #4968, opened 2024-02-20
Elsa 3.0.5 - Programmatic workflow with MassTransit MessageReceived activity not starting workflow


ckeller81:
**Summary:**
When using Elsa with Elsa.MassTransit and Elsa.MassTransit.RabbitMq and using the MessageReceived activity in a programmatically defined workflow as a start activity / trigger, the workflow is never started.

**Used Version**
All packages are from NuGet in the version 3.0.5, the example project does not have any other packages referenced.

**Example-Code:**
https://github.com/ckeller81/elsa-masstransit-test

**Reproduce steps:**
- Clone repo
- start RabbitMq from ./rabbitmq using 'docker-compose up'
- start project from ./src/ElsaMassTransitTest.csproj using 'dotnet run .\ElsaMassTransitTest.csproj'
- call http://localhost:5188/ to start a workflow from a queue-message
- call http://localhost:5188/simple to start a console-log workflow

**Expected behaviour:**
- the call to http://localhost:5188/simple writes "Hello world from simple workflow" to the console
- the call to http://localhost:5188/ writes "Hello world ... from Bus" to the console

**Actual behaviour:**
- :white_check_mark: the call to http://localhost:5188/simple writes "Hello world from simple workflow" to the console
- :x: the call to http://localhost:5188/ does **not** execute the workflow and does **not** write anything to the console

**Workaround**
Use elsa without MassTransit, consume the bus-message in a own consumer and start the workflow from this consumer.

**Thoughts**
After some hours of research, I found out that the queue-message is consumed correctly, but it seems that the workflow-start trigger is either not finding the workflow, or the bookmark. But this was a bit too deep in the package. 😄

RyoukoKonpaku:
From my investigation on making this work on my end, it's because the `Type` property doesn't match with the incoming payload type which is used on the [WorkflowMessageConsumer](https://github.com/elsa-workflows/elsa-core/blob/662a0767269f7b623297daae94bea80249a6e2d7/src/modules/Elsa.MassTransit/Consumers/WorkflowMessageConsumer.cs#L30) to dispatch which activity gets triggered, so it never gets executed. Designer created triggers fine as it assigns the type property to the message type.

You can workaround it by overriding that property to match what it does when creating a trigger from the designer. 

```cs
new MessageReceived
{
    CanStartWorkflow = true,
    Type = ActivityTypeNameHelper.GenerateTypeName(typeof(YourMessageType)),
[...]

PR #4981, opened 2024-02-22
Example: added MassTransit triggered workflow example-project (#4968)

ckeller81:
Added example of MassTransit-triggered programmatic workflow as proposed in #4968 

To test, just start the example-application and execute a POST to /elsa/api/messages

**Additional fix:**
Added .UseMassTransitDispatcher() to the MassTransitActivities example, as the example was throwing an exception during startup.


sfmskywalker (approved):
Very nice 🤩

PR #4983, opened 2024-02-22
Implement distributed lock for registry population

### LINKED ISSUE #4982: 'WorkflowDefinition' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked
Sometimes, after restarting the pods (or pod crash) the pod fails to start with the following exceptions
'WorkflowDefinition' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked

sfmskywalker:
Added distributed lock in 'PopulateRegistriesHostedService' to prevent concurrent registry updates. Also implemented a semaphore in 'DefaultWorkflowDefinitionStorePopulator' to control access to shared resources during add or update operations. This change helps to ensure the integrity and consistency of the workflow registries.

May fix #4982

### DIFF
--- a/src/modules/Elsa.Workflows.Runtime/HostedServices/PopulateRegistriesHostedService.cs
+++ b/src/modules/Elsa.Workflows.Runtime/HostedServices/PopulateRegistriesHostedService.cs
@@ -1,9 +1,11 @@
 using Elsa.Workflows.Contracts;
 using Elsa.Workflows.Runtime.Contracts;
+using Elsa.Workflows.Runtime.Options;
 using JetBrains.Annotations;
 using Medallion.Threading;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
 
 namespace Elsa.Workflows.Runtime.HostedServices;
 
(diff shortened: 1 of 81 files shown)


src/modules/Elsa.Workflows.Runtime/HostedServices/PopulateRegistriesHostedService.cs
```diff
@@ -24,8 +27,12 @@ public PopulateRegistriesHostedService(IServiceScopeFactory scopeFactory)
     public async Task StartAsync(CancellationToken cancellationToken)
     {
         using var scope = _scopeFactory.CreateScope();
-        var registriesPopulator = scope.ServiceProvider.GetRequiredService<IRegistriesPopulator>();
-        await registriesPopulator.PopulateAsync(cancellationToken);
+        var distributedLockProvider = scope.ServiceProvider.GetRequiredService<IDistributedLockProvider>();
+        await using (await distributedLockProvider.AcquireLockAsync("PopulateRegistries", TimeSpan.FromMinutes(10), cancellationToken))
```
raymonddenhaan, line 31: Do we want instances to possibly wait 10min before timing out?

raymonddenhaan (approved):

src/modules/Elsa.Workflows.Runtime/HostedServices/PopulateRegistriesHostedService.cs
sfmskywalker: I'm thinking of it like this: each pod **must** go through this step, but they can only go through it one at a time. If it so happens to be that a pod B has to wait for pod A, there's no telling how long the process might take. Ideally, just a few milliseconds to a few seconds - perhaps one or two minutes if someone went crazy with the number of workflow definitions as activities. Regardless how long, pod B must wait. Obviously, 10 minutes is a very long time, which also means it should be relatively safe to set it to 10 minutes - if pod A takes longer than that, probably something isn't going very well over there. But instead of waiting for pod A indefinitely, the 10 minutes is an ultimate timeout to ensure pod B crashes with a timeout (which, by the way, could also happen if the shared resource to acquire a lock is unreachable).

That being said, it would be better for the user to configure this timeout, with a sensible default - which I think should be around 10 minutes, given that it is **very** unlikely that this step takes more than a couple of seconds, let alone 10 minutes, even with a large number of workflows in the system.

sfmskywalker: I updated this section to leverage the Distributed Locking Timeout option which can be configured from the application itself.

PR #5004, opened 2024-02-26
Respond with 500 in case of Incidents

### LINKED ISSUE #5003: HTTP Endpoint Activity Incorrectly Responds with 200 OK on Incidents with "Continue with Incidents" Strategy
**Issue Summary:**
When configuring a workflow to act as an API endpoint via the HTTP Endpoint activity, it's observed that the server's response behavior is inconsistent when handling incidents. Specifically, if a fault occurs during the execution of a request, the server correctly returns a 500 Internal Server Error status code when the Incident Handling Strategy is set to "Fault." However, altering this strategy to "Continue with Incidents" results in the server responding with a 200 OK status code, despite the occurrence of an incident.

**Expected Behavior:**
The server should respond with a 500 Internal Server Error status code when an incident occurs during the HTTP request that invokes the workflow, regardless of the Incident Handling Strategy configuration ("Fault" or "Continue with Incidents").

**Steps to Reproduce:**
1. Create a workflow with an HTTP Endpoint activity designed to serve as an API endpoint.
2. Configure the Incident Handling Strategy to "Fault" and trigger a fault; observe the correct 500 Internal Server Error response.
3. Change the Incident Handling Strategy to "Continue with Incidents" and trigger a fault again.
4. Note that the server responds with a 200 OK status code instead of the expected 500 Internal Server Error.

**Impact:**
This behavior can lead to incorrect handling of errors by clients consuming the API, as they may interpret the 200 OK response as a successful request, despite the occurrence of an incident within the workflow execution.

**Suggested Fix:**
Adjust the HTTP Endpoint activity's error handling mechanism to ensure that a 500 Internal Server Error is returned when an incident occurs, even if the workflow's Incident Handling Strategy is set to "Continue with Incidents".

sfmskywalker:
Fixes #5003

### DIFF
--- a/src/modules/Elsa.Dapper/Services/Store.cs
+++ b/src/modules/Elsa.Dapper/Services/Store.cs
@@ -245,6 +245,11 @@ public class Store<T> where T : notnull
     {
         var query = _dbConnectionProvider.CreateQuery().Delete(TableName);
         filter(query);
+        
+        // If there are no conditions, we don't want to delete all records.
+        if (!query.Parameters.ParameterNames.Any())
+            return 0;
+        
         using var connection = _dbConnectionProvider.GetConnection();
         return await query.ExecuteAsync(connection);
     }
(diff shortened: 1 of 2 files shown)


src/modules/Elsa.Dapper/Services/Store.cs
```diff
@@ -245,6 +245,11 @@ public async Task<long> DeleteAsync(Action<ParameterizedQuery> filter, Cancellat
     {
         var query = _dbConnectionProvider.CreateQuery().Delete(TableName);
         filter(query);
+        
+        // If there are no conditions, we don't want to delete all records.
+        if (!query.Parameters.ParameterNames.Any())
```
raymonddenhaan, line 250: Do we want to support a delete all statement?

raymonddenhaan (approved):

src/modules/Elsa.Dapper/Services/Store.cs
sfmskywalker: Probably eventually, yes, but I couldn't think of a decent solution on the spot, and I don't think we have a current use case for it that warrants additional work on the Dapper provider.

PR #5008, opened 2024-02-27
Fix Security Bug: No permissions applied to version endpoints

IsaacHayes1995:
List,Delete, and Revert endpoints for editing endpoints by version were set to allow anonymous, meaning no authorisation policies were applied.

I have changed these to apply permissions as per the rest of the API.


IsaacHayes1995:
@dotnet-policy-service agree

sfmskywalker (approved):

sfmskywalker:
Great catch @IsaacHayes1995 , thank you for raising this issue + PR 🙏🏻

PR #5026, opened 2024-03-03
fix missing metadata when persisting bookmark

carbonhe:


carbonhe:
@dotnet-policy-service agree

carbonhe:
@sfmskywalker  Can you merge this pr? This is a very simple bug fix and no other side effects. So that I can revert my local change

sfmskywalker:
Done. Thanks for the PR!

PR #5032, opened 2024-03-04
Alterations with Queries

### LINKED ISSUE #5006: Alteration improvements - Query Parameters
Currently, the Alterations feature allows the user to provide a list of workflow instance IDs that need to be altered.
In order to allow the client application to specify specific ranges of workflows that match some query, we currently have to do this in three steps:

1. Query the desired set of workflow instances using the /workflow-instances API endpoint (which supports specifying queries);
2. Project the received set of workflow instances into a list of workflow instance IDs;
3. Submit a new Alteration Plan and include the list of workflow instance IDs.

Although this works, there are at least two potential issues:

1. We need to somehow project the workflow instances to a list of IDs, perhaps using another tool or script.
2. If we need to process hundreds of thousands of workflow instances, the Alteration Plan that will be stored in the database will have a very large column.

We should enhance this as follows:
- In addition to accept a list of workflow instance IDs, the Alterations API should accept query parameters:
   - Correlation IDs
   - Definition Version IDs
   - Definition IDs + Version
   - Has Incidents
   - Date ranges:
        - Created on / before / after, etc.
        - Updated on / before / after, etc.
   - Activity IDs
      - and Activity Status
- The specified query params should be stored as part of the Alteration Plan. These params are then used when the plan is processed.
   - When a plan is processed, it should execute the query and retrieve the results in batches.
   - For each matching workflow instance in the batch, a new Alteration Job record is created. This is already the case, but the current implementation only considers the Workflow Instance IDs list.

sfmskywalker:
This PR adds the ability to send query parameters when submitting an alteration plan.

It also leverages ELSA itself to coordinate the various steps to process & execute the alteration plan. This led to the introduction of the "Is System Workflow" flag, which further enables other "system" workflows to be provided by modules and allowing tools to exclude these workflows from the UI.

Closes #5006


src/modules/Elsa.Workflows.Runtime/Filters/ActivityExecutionRecordFilter.cs
```diff
@@ -46,12 +46,37 @@ public class ActivityExecutionRecordFilter
...
 
     /// <summary>
     /// Whether to include completed activity execution records. If not specified, all activity execution records will be included.
     /// </summary>
     public bool? Completed { get; set; }
     
+    /// <summary>
+    /// Returns true if the filter is empty.
+    /// </summary>
+    public bool IsEmpty => Id == null && Ids == null && WorkflowInstanceId == null && WorkflowInstanceIds == null && ActivityId == null && ActivityIds == null && ActivityNodeId == null && ActivityNodeIds == null && Name == null && Names == null && Status == null && Statuses == null && Completed == null;
```
raymonddenhaan, line 78: To me this would be more readable if each property is on a separate line

src/modules/Elsa.MassTransit/Features/MassTransitFeature.cs
```diff
@@ -53,21 +55,42 @@ public override void Apply()
...
+            var temporaryConsumers = consumers
+                .Where(c => c.IsTemporary)
+                .ToList();
 
-                if (PrefetchCount != null) configurator.PrefetchCount = PrefetchCount.Value;
+            configure.UsingInMemory((context, busFactoryConfigurator) =>
+            {
+                var options = context.GetRequiredService<IOptions<MassTransitWorkflowDispatcherOptions>>().Value;
+
+                foreach (var consumer in temporaryConsumers)
```
raymonddenhaan, line 67: How will this work when we have the ASB enabled? Does this config get overridden or will it create both InMemory and ASB Endpoints?

raymonddenhaan (approved):

src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20240312145108_V3_1.cs
```diff
@@ -10,6 +10,12 @@ public partial class V3_1 : Migration
         /// <inheritdoc />
         protected override void Up(MigrationBuilder migrationBuilder)
         {
+            migrationBuilder.RenameColumn(
```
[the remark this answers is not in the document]
sfmskywalker: I don't think so - nobody is using alterations right now. We are only  just starting to test it out, so there's no data - or if there is, it would be insignificant.

src/modules/Elsa.MassTransit/Features/MassTransitFeature.cs
sfmskywalker: Notice that this logic sits inside of a local function, that gets assigned to the BusConfigurator lambda property conditionally:

image

That being said, I'll see if I can just move it outside of this method and make it a separate, private method for improved readability.

PR #5036, opened 2024-03-04
Update VersionOptions.LatestAndPublished to set correct property

pramsperger:
`VersionOptions.LatestAndPublished` was incorrectly setting the `LatestOrPublished` property to true rather than the `LatestAndPublished` property.


sfmskywalker (approved):
Great catch 👍🏻

PR #5039, opened 2024-03-05
Added Properties field to ActivityExecutionRecord

### LINKED ISSUE #5038: Implement Alternative Storage Target for Persistent Variables Post-ExecutionContext Eviction
### Description:
Following the recent update where `ActivityExecutionContexts` are no longer persisted for completed activity instances, a critical issue has emerged regarding the persistence of variables. Previously, these variables were stored within `ActivityExecutionContext` objects, leveraging the "Workflow Storage" driver. However, with the eviction of completed activity execution contexts, this method is no longer viable.

We need to explore and implement an alternative storage solution for persistent variables that maintains the integrity and accessibility of these variables throughout the workflow lifecycle. This solution should seamlessly integrate with the current workflow engine design, ensuring that variables remain accessible and mutable as required by the workflow logic.

**Key Points:**
- Identify a new storage target for persistent variables, considering the limitations imposed by the eviction of completed `ActivityExecutionContexts`.
- Ensure the new storage mechanism aligns with the workflow's performance and scalability requirements.
- Maintain or enhance the current level of accessibility and mutability for persistent variables within workflows.

This change is crucial for maintaining the robustness and reliability of workflows that rely on persistent variables for their logic and state management.

sfmskywalker:
This PR adds a `Properties` field to the `ActivityExecutionRecord` entity, allowing the persistence of `ActivityExecutionContext.Properties`.

Fixes #5038


MariusVuscanNx (approved):

src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20240305212213_V3_1.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    /// <inheritdoc />
+    public partial class V3_1 : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "Status",
+                schema: "Elsa",
```
rubenkonica, line 15: Add custom schema as the Initials migrations

sfmskywalker: Hi, thanks for the reminder. We will do this before the 3.1 release. Given that it’s a manual process and we might face more migration updates, it’s more efficient to do it once rather than multiple times.

PR #5053, opened 2024-03-08
Update VersionOptions.cs

pietervoorschot:
When setting the VersionOptions with the TryParse function to LatestAndPublished, it sets IsLatestOrPublished to true instead of IsLatestAndPublished.


pietervoorschot:
@dotnet-policy-service agree

sfmskywalker (approved):
Nice catch - thanks!

Issue #4784, opened 2024-01-10
Fail to specify input for "For Each" Activity in Elsa Studio


teddyli:
I'm following the below articals to setup Elsa Server and Elsa Studio and already upgraded to latest version 3.0.2.
https://v3.elsaworkflows.io/docs/installation/elsa-server
https://v3.elsaworkflows.io/docs/installation/elsa-studio-blazorserver

However I found in the Elsa Studio, the drop-down-list of ForEach Activity Input is always nothing, so I'm not able to specify the array to loop, even though I have define a few array-typed variables in the workflow.

![image](https://github.com/elsa-workflows/elsa-core/assets/865755/9b4ea44c-767f-46c7-9165-a90a23e15cb4)

PR #5062, opened 2024-03-12
Correct ForEach : The activity ForEach take one Variable in enter to …

DelarreAlexis:
Currently, the ForEach activity cannot be used because we cannot choose the input; we always end up with an empty list. To correct this issue, I have decided to use an input of the Variable type, which allows me to avoid having an empty list.

https://github.com/elsa-workflows/elsa-core/issues/4784

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Activities/ForEach.cs
+++ b/src/modules/Elsa.Workflows.Core/Activities/ForEach.cs
@@ -1,4 +1,8 @@
 using Elsa.Workflows.Attributes;
+using Elsa.Extensions;
+using Elsa.Workflows.Contracts;
+using Elsa.Workflows.Memory;
+using Elsa.Workflows.Models;
 
 namespace Elsa.Workflows.Activities;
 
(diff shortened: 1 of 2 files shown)


DelarreAlexis:
@dotnet-policy-service agree company="Stereograph"

sfmskywalker:
Hi @DelarreAlexis,

Thank you for your contribution through the pull request. While your solution meets your specific needs, it introduces a couple of concerns:

1. It restricts us to predefined variables, deviating from our functional specifications that allow for dynamic expressions.
2. It leads to redundant code, which we aim to minimize.

However, we acknowledge the suboptimal user experience of an empty dropdown list and agree it necessitates improvement. The optimal solution would involve integrating a variable picker UI as the default for the ForEach activity's Items input. This approach maintains flexibility for dynamic expressions and reduces code redundancy.

Furthermore, this variable picker UI could potentially benefit other activities requiring similar input methods, enhancing overall usability and consistency.

We appreciate your understanding and look forward to your thoughts on this suggested direction.

sfmskywalker:
@DelarreAlexis I will proceed to closing this PR. Although we couldn't accept your PR this time, I look forward to your future PRs. Thank you.

PR #5074, opened 2024-03-14
Made loading workflows more robust

### LINKED ISSUE #5033: ON CONFLICT DO UPDATE command cannot affect row a second time while loading WorkflowDefinitions
This is a continuation of bug #4982.
Due to changes made to the BulkUpsertAsync method we get a different exception, however the cause seems to be the same.

raymonddenhaan:
Fixes #5033 and other issues with importing workflow definitions through background processes.

- Prevents trying to add multiple versions with the Id
- Prevents multiple definitions with status IsLatest
- Makes most recent workflow (based on version) leading
- Fixes version comparison that caused issues in earlier attempts of this fix (Version from Identity compared to Version from root)

### DIFF
--- a/src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
+++ b/src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
@@ -92,6 +92,9 @@ public class DefaultWorkflowDefinitionStorePopulator : IWorkflowDefinitionStoreP
         var workflow = materializedWorkflow.Workflow;
         var definitionId = workflow.Identity.DefinitionId;
 
+        var existingWorkflowLatest = false;
+        var existingWorkflowPublished = false;
+
         // Serialize materializer context.
         var materializerContext = materializedWorkflow.MaterializerContext;
         var materializerContextJson = materializerContext != null ? _payloadSerializer.Serialize(materializerContext) : default;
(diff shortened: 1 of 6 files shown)


src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
```diff
@@ -202,6 +166,67 @@ private async Task AddOrUpdateCoreAsync(MaterializedWorkflow materializedWorkflo
...
+            var latestWorkflowDefinitions = (await _workflowDefinitionStore.FindManyAsync(filter, cancellationToken)).ToList();
+
+            // If the latest definitions contains definitions with the same ID then we need to replace them with the latest workflow definitions.
+            SyncExistingCopies(latestWorkflowDefinitions, workflowDefinitionsToSave);
+
+            foreach (var latestWorkflowDefinition in latestWorkflowDefinitions)
+            {
+                if (latestWorkflowDefinition.Version > workflow.Identity.Version)
+                {
+                    _logger.LogWarning("A more recent version of the workflow has been found, overwriting the IsLatest property on the workflow.");
```
sfmskywalker, line 190: Log event messages should be fragments, not sentences. Avoid a trailing period/full stop.

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
```diff
@@ -202,6 +166,67 @@ private async Task AddOrUpdateCoreAsync(MaterializedWorkflow materializedWorkflo
...
+                var publishedWorkflowDefinitions = (await _workflowDefinitionStore.FindManyAsync(filter, cancellationToken)).ToList();
+
+                // If the published workflow definitions contains definitions with the same ID as definitions in the latest workflow definitions, then we need to replace them with the latest workflow definitions.
+                SyncExistingCopies(publishedWorkflowDefinitions, workflowDefinitionsToSave);
+
+                foreach (var publishedWorkflowDefinition in publishedWorkflowDefinitions)
+                {
+                    if (publishedWorkflowDefinition.Version > workflow.Identity.Version)
+                    {
+                        _logger.LogWarning("A more recent version of the workflow has been found to be published, overwriting the IsPublished property on the workflow.");
```
sfmskywalker, line 220: Log event messages should be fragments, not sentences. Avoid a trailing period/full stop.

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
```diff
@@ -114,47 +117,8 @@ private async Task AddOrUpdateCoreAsync(MaterializedWorkflow materializedWorkflo
...
-            SyncExistingCopies(publishedWorkflowDefinitions, workflowDefinitionsToSave);
-
-            foreach (var publishedWorkflowDefinition in publishedWorkflowDefinitions)
-            {
-                publishedWorkflowDefinition.IsPublished = false;
-                workflowDefinitionsToSave.Add(publishedWorkflowDefinition);
-            }
-        }
+        await UpdateIsLatest();
+        await UpdateIsPublished();
```
sfmskywalker, line 121: As an optimization, do you think it would be possible to do these two calls after all workflows have been imported? Perhaps, while a workflow is being imported, you should capture information like "this workflow is published" in local variables so that you can be specific about what workflow definition versions must be published.

sfmskywalker: As discussed offline with @raymonddenhaan , we will revisit this later.

sfmskywalker (approved):

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
raymonddenhaan: Created enhancement #5075 for this.

PR #5084, opened 2024-03-18
Add schema for migration v3_1 #5022

### LINKED ISSUE #5022: Wrong schema in V3_1 SqlServer migrations
There is a bug in the new migrations to the new version in SqlServer. 
The initial migrations use the configured schema and the new ones use the Elsa default schema.

jdevillard:
add schema for migration 

should fix #5022

### DIFF
--- a/src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20240312145108_V3_1.cs
+++ b/src/modules/Elsa.EntityFrameworkCore.MySql/Migrations/Alterations/20240312145108_V3_1.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.EntityFrameworkCore.Migrations;
+﻿using Elsa.EntityFrameworkCore.Common.Contracts;
+using Microsoft.EntityFrameworkCore.Migrations;
 
 #nullable disable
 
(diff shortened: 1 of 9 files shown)


raymonddenhaan:
While this does fix the issue, if we recreate the migration again this issue will resurface. I think it was @sfmskywalker idea to fix this just before creating the 3.1 release to prevent having to do this each and every time we are creating migrations in the preview versions.

sfmskywalker:
Thanks @jdevillard !

@raymonddenhaan is right, let's do this right before the - otherwise we have have to do this again (there may still be schema changes underway).

jdevillard:
I agree with you. But I don't have find anything to change add the ctor of the Generated Migration Class or change the Base Class. 

I only find CSharpMigrationsGenerator which (if I'm not wrong) could be fine to change the 
`schema: "Elsa"` by `schema: _schema.Schema,`.

After we can have different options : 
- add a new class before each migration (because the migration is `partial class`) which can be possible because we now the name of the migration
- look up the CS file to find the migration and use Roslyn to modify the .cs and add a constructor (maybe more difficult)

and then use the `CSharpMigrationsGenerator ` to get th right value in the migration operation.

what do you think ? prefer?

sfmskywalker:
@jdevillard If theres a way to generate the desired migrations, then that would be absolutely fantastic.

jdevillard:
I'm currently working on a tooling that we can use on top of Ef core tooling to generate migration and use Roslyn to change the result .cs file.

The core code is well advanced, need to fight with Whitespace and Tabulation to get a beautiful code generated ^^ .

jdevillard:
Tool is near the end ! need to polish one more thing (and write the readme) and publish everything to github/nuget

sfmskywalker:
@jdevillard Given that we need to do a final update of the migrations before the imminent 3.1 release, should we close this PR?

jdevillard:
Yes @sfmskywalker lets close this pr and change the code using the tools discussed in discord ;)

sfmskywalker:
That tool you created is awesome 🙏🏻 Saves us a ton of time!

PR #5095, opened 2024-03-19
Implemented distributed cache refresh for workflow definitions

raymonddenhaan:
This PR will ensure that all instances will keep an up-to-date cache ActivityRegistry


src/modules/Elsa.Workflows.Management/Handlers/RefreshActivityRegistryHandler.cs
```diff
@@ -42,6 +37,16 @@ public async Task HandleAsync(WorkflowDefinitionPublished notification, Cancella
...
+    public async Task HandleAsync(WorkflowDefinitionVersionDeleted notification, CancellationToken cancellationToken) => await RefreshAsync(cancellationToken);
+
+    /// <inheritdoc />
+    public async Task HandleAsync(WorkflowDefinitionVersionsDeleted notification, CancellationToken cancellationToken) => await RefreshAsync(cancellationToken);
     
-    private async Task RefreshAsync(CancellationToken cancellationToken) => await _activityRegistryPopulator.PopulateRegistryAsync(typeof(WorkflowDefinitionActivityProvider), cancellationToken);
+    private async Task RefreshAsync(CancellationToken cancellationToken)
+    {
+        await workflowDefinitionDispatcher.DispatchAsync(new RefreshWorkflowDefinitionsRequest(), cancellationToken);
+        await activityRegistryPopulator.PopulateRegistryAsync(typeof(WorkflowDefinitionActivityProvider), cancellationToken);
```
sfmskywalker, line 50: The sender of this request will also receive it, which means it will re-populate the registry twice. Maybe awe might consider introducing MT middleware that includes a SenderID that stores the current application instance ID, so that it can then simply not deliver the message to the sender's consumer class. Populating the registry can potentially be a resource-intensive operation when a system has a large number of workflows, so it might be a worthwhile optimization. In any case, this would be out of scope for this PR, but would love to hear your thoughts.

sfmskywalker (changes requested):

src/modules/Elsa.Workflows.Management/Handlers/RefreshActivityRegistryHandler.cs
raymonddenhaan: That might be an elegant and generic way to prevent duplicate execution. I wouldn't be surprised if MT is already adding the SenderID out of the box. We would need to make ignoring the message from the send optional though

sfmskywalker (approved):

PR #5097, opened 2024-03-19
Add alteration notifications and management

### LINKED ISSUE #5096: Alterations API: Submitted planned is stuck on status: Dispatching
### Repro Steps

1. create an alteration plan
2. submit plan
3. get Plan id
4. check status of plan id

### Symptoms

Plan never changes status: is stuck on dispatching / tried with different filters.

```json
{
    "plan": {
        "alterations": [
            {
                "type": "ModifyVariable",
                "variableId": "2bbea8ff54191747",
                "value": "100"
            },
            {
                "type": "Migrate",
                "targetVersion": 9
            },
            {
                "type": "ScheduleActivity",
                "activityId": "d49e835cb0a87ce9"
            }
        ],
        "workflowInstanceFilter": {
            "workflowInstanceIds": [
                "c44f1565f93dd77"
            ],
            "correlationIds": [],
            "timestampFilters": [],
            "definitionVersionIds": [],
            "activityFilters": []
        },
        "status": "Dispatching",
        "createdAt": "2024-03-19T14:51:05.325469+00:00",
        "id": "1c9354f5892addf1"
    },
    "jobs": []
}
```

sfmskywalker:
Implement notification publishing for completed alteration plans and add related handlers and management logic to coordinate alterations workflow triggering and completion status updates.

Fixes #5096

### DIFF
--- a/src/modules/Elsa.Alterations/Services/AlterationPlanManager.cs
+++ b/src/modules/Elsa.Alterations/Services/AlterationPlanManager.cs
@@ -1,0 +1,43 @@
+using Elsa.Alterations.Core.Contracts;
+using Elsa.Alterations.Core.Entities;
+using Elsa.Alterations.Core.Enums;
+using Elsa.Alterations.Core.Filters;
+using Elsa.Alterations.Core.Notifications;
+using Elsa.Common.Contracts;
+using Elsa.Mediator.Contracts;
+
+namespace Elsa.Alterations.Services;
+
+/// <inheritdoc />
+public class AlterationPlanManager(IAlterationPlanStore planStore, IAlterationJobStore jobStore, ISystemClock systemClock, INotificationSender notificationSender) : IAlterationPlanManager
+{
+    /// <inheritdoc />
+    public async Task<AlterationPlan?> GetPlanAsync(string planId, CancellationToken cancellationToken = default)
+    {
+        var planFilter = new AlterationPlanFilter { Id = planId };
+        return await planStore.FindAsync(planFilter, cancellationToken);
+    }
+
+    /// <inheritdoc />
+    public async Task<bool> GetIsAllJobsCompletedAsync(string planId, CancellationToken cancellationToken = default)
+    {
+        // Check if all jobs are completed.
+        var jobFilter = new AlterationJobFilter
+        {
+            PlanId = planId,
+            Statuses = new[] { AlterationJobStatus.Pending, AlterationJobStatus.Running }
+        };
+        
+        return await jobStore.CountAsync(jobFilter, cancellationToken) == 0;
+    }
+
+    /// <inheritdoc />
+    public async Task CompletePlanAsync(AlterationPlan plan, CancellationToken cancellationToken = default)
+    {
+        plan.Status = AlterationPlanStatus.Completed;
+        plan.CompletedAt = systemClock.UtcNow;
+        
+        await planStore.SaveAsync(plan, cancellationToken);
+        await notificationSender.SendAsync(new AlterationPlanCompleted(plan), cancellationToken);
+    }
+}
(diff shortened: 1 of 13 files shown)


MariusVuscanNx (approved):

src/modules/Elsa.Alterations/Workflows/ExecuteAlterationPlanWorkflow.cs
```diff
@@ -32,8 +33,15 @@ protected override void Build(IWorkflowBuilder builder)
                     Result = new(planId)
                 },
                 new Correlate(planId),
-                new GenerateAlterationJobs(planId),
-                new DispatchAlterationJobs(planId),
+                new GenerateAlterationJobs(planId)
+                {
+                    Result = new(jobCount)
+                },
+                new If(context => jobCount.Get(context) > 0)
```
raymonddenhaan, line 40: Can we check if the correct jobs have been created instead of just relying on the count?

src/modules/Elsa.Alterations/Services/AlterationPlanManager.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        var planFilter = new AlterationPlanFilter { Id = planId };
+        return await planStore.FindAsync(planFilter, cancellationToken);
+    }
+
+    /// <inheritdoc />
+    public async Task<bool> GetIsAllJobsCompletedAsync(string planId, CancellationToken cancellationToken = default)
+    {
+        // Check if all jobs are completed.
+        var jobFilter = new AlterationJobFilter
```
raymonddenhaan, line 25: What if the jobs do not take much time and some of them are already completed?

raymonddenhaan: Or failed?

sfmskywalker: f some jobs have completed while others are still running, then this method will return false, since not all jobs have completed yet, given that there is a positive number of jobs with a status of Running or Pending.

But I am probably missing your point, so please elaborate if I did.

src/modules/Elsa.Alterations/Workflows/ExecuteAlterationPlanWorkflow.cs
sfmskywalker: Why?

src/modules/Elsa.Alterations/Services/AlterationPlanManager.cs
raymonddenhaan: Nevermind, I read this completely the wrong way.

src/modules/Elsa.Alterations/Workflows/ExecuteAlterationPlanWorkflow.cs
raymonddenhaan: If we can guarantee that there will be no duplicates created than the count should be enough. However if we can not guarantee it we should check to see if the correct plans have been created.

raymonddenhaan (approved):

src/modules/Elsa.Alterations/Workflows/ExecuteAlterationPlanWorkflow.cs
sfmskywalker: This logic is about determining if the Plan has any jobs or not. Because of no jobs were created, the plan should complete immediately. If at least one job was created, then we will wait for it to complete. Even if duplicate jobs are created (of which I have seen no evidence or suspect logic), we only care about the number of jobs in order to know if we need to complete the plan or wait for the jobs to complete.

Given this context, do you still see a potential problem with this logic?

Issue #5071, opened 2024-03-14
Add dry run option to alterations


raymonddenhaan:
Add the possibility for a dry run to the alterations API making it possible to check which workflows are going to be affected by the created query.

PR #5110, opened 2024-03-20
Add DryRun endpoint and workflow instance finder

### LINKED ISSUE #5109: Activity Filter in API Not Filtering as Expected
We've encountered an issue where the `ActivityFilter` does not seem to be functioning as expected. Despite applying filters intended to yield no results under certain conditions, plans are being executed for instances that should ostensibly be filtered out.

### Initial Conditions:
- There are two workflow instances.

### Steps to Reproduce:
1. Create a plan with the following filter criteria:
   - `hasIncidents`: true (Expecting 2 instances to match this condition)
   - A filter with arbitrary values for an activity, which should theoretically result in 0 matching instances:
     ```json
     "filter": {
       "hasIncidents": true,
       "activityFilters": [
         {
           "id": "1",
           "nodeId": "1",
           "name": "Ijustmadethisup",
           "status": "Pending"
         }
       ]
     }
     ```

2. Expectation: No workflow instances should be found due to the specified `activityFilters`.

3. Outcome: Despite the expectation, the plan is executed for the two instances.

### Observed Behavior:
Upon reviewing the executed plan via `{{elsaurl}}elsa/api/alterations/227522d98102f026`, it's observed that the plan proceeded with alterations on both workflow instances that were supposed to be excluded by the filter.

```json
"workflowInstanceFilter": {
  "hasIncidents": true,
  "activityFilters": [
    {
      "id": "1",
      "nodeId": "1",
      "name": "Ijustmadethisup",
      "status": "Pending"
    }
  ]
}
```

The status of both jobs is marked as "Completed", indicating that the filtering did not work as intended.

### Expected Behavior:
The `ActivityFilter` should accurately filter out workflow instances that do not meet the specified criteria, in this case, not matching the arbitrary values specified for an activity.

### Additional Context:
- This issue directly impacts the ability to selectively perform operations on workflow instances, potentially affecting performance and the accuracy of workflow alterations.
- Further investigation into how filters are applied and processed in the API might be necessary to identify the root cause of this discrepancy.

sfmskywalker:
Added a new DryRun endpoint for executing alteration plans and implemented the IWorkflowInstanceFinder interface to search for workflow instances. Refactored existing code to use new workflow instance finder service, resulting in cleaner activity implementation.

Fixes #5109

### DIFF
--- a/src/modules/Elsa.Alterations.Core/Services/WorkflowInstanceFinder.cs
+++ b/src/modules/Elsa.Alterations.Core/Services/WorkflowInstanceFinder.cs
@@ -1,0 +1,54 @@
+using Elsa.Alterations.Core.Contracts;
+using Elsa.Alterations.Core.Models;
+using Elsa.Workflows.Management.Contracts;
+using Elsa.Workflows.Management.Filters;
+using Elsa.Workflows.Runtime.Contracts;
+using Elsa.Workflows.Runtime.Filters;
+
+namespace Elsa.Alterations.Core.Services;
+
+/// <inheritdoc />
+public class WorkflowInstanceFinder(IWorkflowInstanceStore workflowInstanceStore, IActivityExecutionStore activityExecutionStore) : IWorkflowInstanceFinder
+{
+    /// <inheritdoc />
+    public async Task<IEnumerable<string>> FindAsync(AlterationWorkflowInstanceFilter filter, CancellationToken cancellationToken = default)
+    {
+        var workflowInstanceFilter = new WorkflowInstanceFilter
+        {
+            Ids = filter.WorkflowInstanceIds?.ToList(),
+            DefinitionVersionIds = filter.DefinitionVersionIds?.ToList(),
+            CorrelationIds = filter.CorrelationIds?.ToList(),
+            HasIncidents = filter.HasIncidents,
+            IsSystem = filter.IsSystem,
+            TimestampFilters = filter.TimestampFilters?.ToList(),
+        };
+        var activityExecutionFilters = filter.ActivityFilters?.Select(x => new ActivityExecutionRecordFilter
+        {
+            ActivityId = x.ActivityId,
+            Id = x.ActivityInstanceId,
+            ActivityNodeId = x.NodeId,
+            Name = x.Name,
+            Status = x.Status,
+        }).ToList();
+
+        var workflowInstanceIds = workflowInstanceFilter.IsEmpty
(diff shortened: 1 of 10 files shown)


src/modules/Elsa.Alterations.Core/Services/WorkflowInstanceFinder.cs
```diff
@@ -0,0 +1,74 @@
...
+            Id = x.ActivityInstanceId,
+            ActivityNodeId = x.NodeId,
+            Name = x.Name,
+            Status = x.Status,
+        }).ToList();
+
+        var workflowInstanceFilterIsEmpty = WorkflowFilterIsEmpty(workflowInstanceFilter);
+
+        var workflowInstanceIds = workflowInstanceFilterIsEmpty
+            ? Enumerable.Empty<string>().ToHashSet()
```
raymonddenhaan, line 39: Do we give some feedback to the user that they actually need to enter filter criteria or is this implied from the empty list that is returned?

src/modules/Elsa.Alterations.Core/Services/WorkflowInstanceFinder.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (workflowInstanceFilterIsEmpty)
+                workflowInstanceIds = matchingWorkflowInstanceIds;
+            else
+                workflowInstanceIds.IntersectWith(matchingWorkflowInstanceIds);
+        }
+
+        return workflowInstanceIds;
+    }
+
+    private bool WorkflowFilterIsEmpty(WorkflowInstanceFilter filter)
```
raymonddenhaan, line 59: Why remove this from the filter itself? There we would be able to re use the logic

src/modules/Elsa.Alterations/Endpoints/Alterations/DryRun/Endpoint.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <inheritdoc />
+    public override void Configure()
+    {
+        Post("/alterations/dry-run");
+        ConfigurePermissions("run:alterations");
+    }
+
+    /// <inheritdoc />
+    public override async Task HandleAsync(AlterationWorkflowInstanceFilter filter, CancellationToken cancellationToken)
```
raymonddenhaan, line 22: Wouldn't it be easier to add dryrun as an optional parameter to the alteration endpoint? Now we need to ensure these 2 endpoints stay the same.

raymonddenhaan:
This closes #5071

PR #5128, opened 2024-03-22
#5107 Add framework reference & package removal

### LINKED ISSUE #5107: Elsa.FileStorage should be using a Framework Reference instead of a deprecated package
The Elsa.FileStorage package has a package reference for Microsoft.AspNetCore.Http however that package is deprecated and instead a framework reference should be being used as is the case in Elsa.SasTokens

thompson-tomo:
This change introduces a framework reference to the file storage project so that the deprecated package can be removed.

Closes #5107


sfmskywalker (approved):
Awesome 👍🏻

Issue #5031, opened 2024-03-04
Refactor IWorkflowDispatcher to Avoid Large Message Exceptions


raymonddenhaan:
## Description
 
ELSA currently faces a risk of exceptions due to message sizes exceeding the limits supported by message brokers. This issue is primarily caused by large input sent to dispatch a workflow. To mitigate this, we propose a change in the way input are handled.
 
### Proposed Changes:
- **Separate input from Dispatch Messages:** Before dispatching a workflow, input should be saved to the database instead of being included in the messages sent through brokers.
- **Modify Dispatch Message Content:** Dispatch messages should only contain the `WorkflowInstanceId` and the `BookmarkId`.
- **Update IWorkflowDispatcher Implementation:** Implement a new version of `IWorkflowDispatcher` that is responsible for saving input to the database and creating the simplified dispatch message. This implementation should include a transport mechanism (either for background processing or for integration with MassTransit) for sending the dispatch message.
 
### Impact:
- **Workflow Statuses and Handling:** No changes are required for workflow statuses or the handling of input variables. However, care must be taken to ensure that input variables are merged correctly with the workflow instances.
 
This approach aims to enhance the reliability of ELSA by preventing message size exceptions and improving the efficiency of message handling through brokers.

Issue #4835, opened 2024-01-24
 Controlled Commit States


sfmskywalker:
**Objective:**
To enhance the current workflow execution process by introducing the concept of "commit points". This will allow for more frequent and controlled persistence of workflow state to the database, beyond the existing default behaviour of persisting only upon completion or when blocked.

**Background:**
Currently, workflows are executed in-memory and are persisted to the configured data store only upon completion or in scenarios where they can no longer proceed (e.g., due to blocking activities). This approach, while efficient, lacks the flexibility to capture interim states of the workflow during its execution, potentially leading to data loss or inconsistency in case of unexpected interruptions.

**Requirements:**

1. **Configurable Commit Points for Workflows:**
   - Introduce a setting at the workflow level to specify commit behaviour. This includes:
     - Persisting the workflow state to the database before execution begins.
     - Persisting after every activity within the workflow.
     - Persisting after activities designated as "commit points".

2. **Commit Settings for Individual Activities:**
   - Implement an option for each activity within a workflow where users can specify:
     - Committing the workflow state to the database before the execution of the activity.
     - Committing after the activity has been executed.

**Expected Benefits:**
This feature will provide users with greater control over the persistence of workflow states, enhancing reliability and data integrity. It allows for recovery of workflows to a more recent state in case of interruptions, thereby reducing the risk of data loss and improving overall system resilience.

PR #5136, opened 2024-03-26
Feature/cancel dispatched workflows

### LINKED ISSUE #4177: Workflow Cancellation
Hello,
We are using Elsa 2 to construct workflows with activities that access various external resources sequentially in an asynchronous manner. As a precaution, we aim to cancel any workflow instance that exceeds a predefined time limit (e.g., if an API call takes too long).

I believe that using the built-in workflow cancellation or faulting features is not suitable in this case since they do not immediately terminate the workflow execution.

Another option is to utilize a `CancellationToken`. To provide a `CancellationToken`, we need to implement the `IHandleMessages<ExecuteWorkflowDefinitionRequest>` interface and replace its default implementation of `ExecuteWorkflowDefinitionRequestConsumer` in the dependency injection container.

However, canceling the token will also terminate the post-processing of the workflow instance, including notifications such as `WorkflowExecuted`, `WorkflowFaulted`, and `WorkflowExecutionFinished`. Consequently, the workflow instance will become 'stuck' in a running state.

Is there a way to cancel only the 'user-defined' section of the workflow without affecting the pre- and post-processing stages?

Thank you in advance!

raymonddenhaan:
Ensures workflows that are dispatched gets cancelled.

There are 2 known issues:
- Dispatch workflows messages that have not been processed will still execute after a cancellation
- Dispatched workflows that have not been saved to the database (have not been suspended) will not be cancelled.

Both issues will be picked up through issues #5031 and #4835 respectively

Closes #4177


src/modules/Elsa.Common/Extensions/EnumerableExtensions.cs
```diff
@@ -58,4 +58,21 @@ public static Page<T> Paginate<T>(this IEnumerable<T> enumerable, PageArgs? page
     /// <typeparam name="T">The type of the items in the lists.</typeparam>
     /// <returns>True if the lists are equal, otherwise false.</returns>
     public static bool IsEqualTo<T>(this IEnumerable<T> list1, IEnumerable<T> list2) => list1.OrderBy(g => g).SequenceEqual(list2.OrderBy(g => g));
+
+    /// <summary>
+    /// Splits an enumerable into batches of a specified size.
+    /// </summary>
+    public static IEnumerable<IEnumerable<T>> ToBatches<T>(this IEnumerable<T> enumerable, int pageLimit)
```
sfmskywalker, line 65: Given the name of this method, it sounds like you could use the built-in Chunk extension.

sfmskywalker (changes requested):

sfmskywalker (approved):

PR #5142, opened 2024-03-27
Fix Alteration Runner

### LINKED ISSUE #5140: CancelActivity in Alteration API Fails to Update Activity State to Cancelled
When attempting to cancel an activity using the Alteration API, the API returns a successful message, indicating that the `CancelActivity` operation succeeded. However, upon inspecting the workflow instance, the state of the activity remains `Running` instead of updating to the expected `Cancelled` state.

### Steps to Reproduce

1. Start a workflow containing an activity with its state set to `Suspended`.
2. Use the Alteration API to cancel this activity by sending the following payload:
   ```json
   {
     "alterations": [
       {
         "type": "CancelActivity",
         "activityInstanceId": "87d5edfb7f30a1c5"
       }
     ],
     "workflowInstanceIds": ["3504d24c287e902a"]
   }
   ```
3. The API returns a success message, indicating the cancellation was processed:
   ```json
   {
     "results": [
       {
         "workflowInstanceId": "3504d24c287e902a",
         "log": {
           "logEntries": [
             {
               "message": "CancelActivity succeeded",
               "logLevel": "Information",
               "timestamp": "2024-03-19T08:57:28.2839419+00:00"
             }
           ]
         },
         "workflowHasScheduledWork": false,
         "isSuccessful": true
       }
     ]
   }
   ```
4. Upon checking the workflow instance, the activity's state is incorrectly listed as `Running` instead of `Cancelled`.

### Expected Behavior

After calling `CancelActivity` via the Alteration API, the targeted activity's state should be updated to `Cancelled`.

### Actual Behavior

The activity's state remains `Running` despite the successful execution of a `CancelActivity` operation through the Alteration API.

This issue leads to inconsistencies within the workflow's state management and can affect the execution flow and the overall reliability of the workflow management system.

---
**Note**: It is crucial to ensure that state transitions through the Alteration API are accurately reflected within workflow instances to maintain the integrity and accuracy of workflow state tracking.

sfmskywalker:
This PR fixes the Alteration runner by ensuring it runs through the workflow execution middleware pipeline, which is responsible for persisting activity execution logs, workflow execution logs and persistent variables.

Fixes #5140


MariusVuscanNx (approved):

src/modules/Elsa.Alterations/Middleware/Workflows/RunAlterationsMiddleware.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private async Task RunAsync(WorkflowExecutionContext workflowExecutionContext, IEnumerable<IAlteration> alterations, AlterationLog log, CancellationToken cancellationToken = default)
+    {
+        var commitActions = new List<Func<Task>>();
+
+        foreach (var alteration in alterations)
+        {
+            // Find handlers.
+            var supportedHandlers = handlers.Where(x => x.CanHandle(alteration)).ToList();
```
raymonddenhaan, line 32: Do we want to support having multiple handlers for the same type of alteration? I can imagine sequencing could have a major influence if we do.

src/modules/Elsa.Workflows.Core/Pipelines/WorkflowExecution/WorkflowExecutionMiddlewareExtensions.cs
```diff
@@ -11,16 +12,50 @@ public static class WorkflowExecutionMiddlewareExtensions
     /// <summary>
     /// Installs the specified middleware component into the pipeline being built.
     /// </summary>
-    public static IWorkflowExecutionPipelineBuilder UseMiddleware<TMiddleware>(this IWorkflowExecutionPipelineBuilder pipelineBuilder, params object[] args) where TMiddleware: IWorkflowExecutionMiddleware
+    public static IWorkflowExecutionPipelineBuilder UseMiddleware<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] TMiddleware>(
```
raymonddenhaan, line 15: We are using DynamicallyAccessedMembers here on all functions, however this is only directly the case in the last function.

src/modules/Elsa.Alterations/AlterationHandlers/CancelActivityHandler.cs
```diff
@@ -49,32 +34,19 @@ protected override ValueTask HandleAsync(AlterationContext context, CancelActivi
...
-        await SaveActivityExecutionRecordAsync(activityExecutionContext);
-    }
-
-    private async Task RemoveBookmarksAsync(ActivityExecutionContext activityExecutionContext)
+    private async Task CancelAsync(IEnumerable<ActivityExecutionContext> activityExecutionContexts)
     {
-        var filter = new BookmarkFilter { ActivityInstanceId = activityExecutionContext.Id };
-        await _bookmarkManager.DeleteManyAsync(filter, activityExecutionContext.CancellationToken);
+        foreach (var activityExecutionContext in activityExecutionContexts)
+            await CancelAsync(activityExecutionContext);
```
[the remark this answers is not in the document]
sfmskywalker: I think the method was already there, but with more statements originally. But, there's no reason to keep it.

src/modules/Elsa.Alterations/Middleware/Workflows/RunAlterationsMiddleware.cs
sfmskywalker: That's indeed the idea, although I haven't thought of a real use case for it yet. Can you think of a scenario where the order of handlers would have an impact? Because then we might also can think of a real use case, but handle it in a way where one can control the order of things. If not, then we could consider just supporting one handler per type.

sfmskywalker: Given that this is not a functional change - just moved code, I will resolve this comment, but please do continue the conversation.

src/modules/Elsa.Workflows.Core/Pipelines/WorkflowExecution/WorkflowExecutionMiddlewareExtensions.cs
sfmskywalker: Indeed. It would seem that this has a viral effect. If we remove the attribute from this method, for example, the compiler will generate the following warning:

image

Adding the attribute solves this warning.

PR #5149, opened 2024-03-28
Fix concurrency issues with distributed messages

raymonddenhaan:
When MassTransit is enabled the delay with processing the distributed workflow definition messages could result in accessing the ActivityDescriptor dictionary while it is being updated.


raymonddenhaan:
Different approach of #5141

sfmskywalker (approved):

PR #5181, opened 2024-04-04
Improved the returned count of cancelled workflows

raymonddenhaan:
- Removed dispatched child workflows from the count
- Removed finished workflows from the count
- Prevented dispatching a cancellation request for finished workflows as they can not be cancelled

It is still not 100% accurate, however it is less confusing to the user.
Not 100% since we can not guarantee that all dispatched cancellation requests also lead to a cancelled workflow. (Could be finished by the time we actually try to cancel it)


sfmskywalker:
Perhaps we should consider adding a "Canceling" status to workflow instances. This way, we could immediately set the status of the selected workflow instances to `Canceling` - which would be reflected by the UI.
The UI could then potentially receive realtime updates via SignalR (or simple polling) as workflow instances become canceled.

sfmskywalker (approved):

PR #5182, opened 2024-04-04
Enhance BulkDispatchWorkflows Activity with Support for Multiple Input Keys and Customizable Input Key

### LINKED ISSUE #5132: Enhance BulkDispatchWorkflows to Support Dynamic Input Mapping
The `BulkDispatchWorkflows` activity currently allows users to specify a list of objects, dispatching a specified workflow for each item in the list, with the current item sent as an input named "Item". This approach requires the dispatched workflows to anticipate receiving an input named "Item", which couples the child workflows too closely to the `BulkDispatchWorkflows` activity's implementation.

### Proposed Enhancement
To decouple the parent and child workflows, we propose updating the `BulkDispatchWorkflows` activity to treat each object in the list as a dictionary of key/value pairs. Each key/value pair should map directly to an input expected by the child workflow, allowing for a more dynamic and flexible input mechanism. This change will enable the parent workflow to dispatch child workflows without enforcing a strict contract on the input naming convention.

### Example
Consider the following list of objects passed to `BulkDispatchWorkflows`:

```javascript
[
  {
    "Input1": "Alice",
    "Input2": { "customerId": "1" }
  },
  {
    "Input1": "Bob",
    "Input2": { "customerId": "2" }
  }
]
```

With the proposed changes, two child workflows would be dispatched, each receiving inputs `Input1` and `Input2`, enhancing flexibility and decoupling the workflows.

### Benefits
- **Decouples Parent and Child Workflows**: Reduces the dependency on a specific input naming convention, making the architecture more robust.
- **Enhances Flexibility**: Allows the parent workflow to dynamically define the inputs for child workflows based on the specific requirements of each child.
- **Simplifies Workflow Design**: Makes it easier to design child workflows that can be reused in different contexts without modification.

This enhancement aims to improve the usability and flexibility of the `BulkDispatchWorkflows` activity, facilitating a more dynamic and decoupled workflow design.

sfmskywalker:
This PR introduces significant improvements to the BulkDispatchWorkflows activity by enabling the support for multiple input keys. Specifically, it allows users to provide a list of items, with each item being a dictionary. This change permits each dictionary key to act as the input name for the dispatched workflow, providing users with precise control over the inputs of the child workflow.

Furthermore, this update includes the flexibility to configure a different key for the workflow input, moving beyond the previously hardcoded "Item" key. This enhancement ensures backward compatibility for users who prefer not to adopt the "list of dictionaries" method, maintaining the activity's utility in existing workflows without requiring any modifications.

Sample Items JS expression:

```javascript
return [{
    Order: { 
        id: 1,
        customerId: 1
    }
},
{
    Order: { 
        id: 2,
        customerId: 2
    }
}];
```

The dispatched workflows will each now receive an input called "Order".

Fixes #5132

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
+++ b/src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
@@ -65,6 +65,15 @@ public static class DictionaryExtensions
         dictionary.Add(key, value);
         return dictionary;
     }
+    
+    /// <summary>
+    /// Merges the specified dictionary with the other dictionary.
+    /// </summary>
+    public static void Merge(this IDictionary<string, object> dictionary, IDictionary<string, object> other)
+    {
+        foreach (var (key, value) in other)
+            dictionary[key] = value;
+    }
 
     private static T? ConvertValue<T>(object? value) => value.ConvertTo<T>();
 }
--- a/src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
+++ b/src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
@@ -68,6 +68,7 @@ public static class DictionaryExtensions
     
     /// <summary>
     /// Merges the specified dictionary with the other dictionary.
+    /// When a key exists in both dictionaries, the value in the other dictionary will overwrite the value in the specified dictionary.
     /// </summary>
     public static void Merge(this IDictionary<string, object> dictionary, IDictionary<string, object> other)
     {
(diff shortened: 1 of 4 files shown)


src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
```diff
@@ -65,6 +65,15 @@ public static class DictionaryExtensions
...
         return dictionary;
     }
+    
+    /// <summary>
+    /// Merges the specified dictionary with the other dictionary.
+    /// </summary>
+    public static void Merge(this IDictionary<string, object> dictionary, IDictionary<string, object> other)
+    {
+        foreach (var (key, value) in other)
+            dictionary[key] = value;
```
raymonddenhaan, line 75: Do we want to overwrite in case of duplicate values?
Whatever we choose for this, let's make it clear in the comments/remarks

raymonddenhaan (approved):

src/modules/Elsa.Workflows.Core/Extensions/DictionaryExtensions.cs
sfmskywalker: Yes, and good idea 👍🏻

PR #5188, opened 2024-04-05
feat(liquid): delete json filter and add filters already provided by …

jeanbaptistedalle:
feat(liquid): delete json filter and add filters already provided by fluid library


test/integration/Elsa.Workflows.IntegrationTests/Scenarios/HttpRequestWithLiquid/HttpRequestWithLiquidWorkflow.cs
```diff
@@ -0,0 +1,45 @@
...
+                    Url = new Input<Uri?>(new Uri("https://fakestoreapi.com/products?limit=5")),
+                    ParsedContent = new Output<object?>(products)
+                },
+                new WriteLine(new Expression("Liquid", "First product id: {{ Variables.Products[0].id }}")),
+                new WriteLine(new Expression("Liquid", "First product price rounded: {{ Variables.Products[0].price | round }}")),
+                new WriteLine(new Expression("Liquid", "First product as json: {{ Variables.Products[0] | json }}")),
+                new WriteLine(new Expression("Liquid", "Second product id: {{ Variables.Products[1].id }}")),
+                new FlowSendHttpRequest
+                {
+                    Url = new Input<Uri?>(new Expression("Liquid", "https://fakestoreapi.com/products/{{ Variables.Products[1].id }}")),
```
sfmskywalker, line 37: I am not sure it's such a good idea to make HTTP requests to external services during integration tests, as it might cause the test to fail in case the external site is not available, which is a dependency we shouldn't want here.

As an alternative, perhaps replace this activity with a RunJavaScript activity that produces the JSON payload as its output?

I can push that change if you want after merging the PR, just let me know.

src/modules/Elsa.Liquid/Extensions/TemplateContextExtensions.cs
```diff
@@ -18,5 +19,13 @@ internal static void AddFilters(this TemplateContext templateContext, FluidOptio
...
             });
         }
+        if (options.AddFluidFilters)
+        {
+            templateContext.Options.Filters
+                .WithArrayFilters()
+                .WithStringFilters()
+                .WithNumberFilters()
+                .WithMiscFilters();
+        }
```
sfmskywalker, line 29: Assuming there could be more filters, coming from Fluid and potentially other packages, a more extensible way would be to replace the options with a delegate that defaults to installing these filters, while allowing the application to override the delegate to e.g. reduce or increase the set of filters.

I can push that change if you want after merging the PR, just let me know.

sfmskywalker (approved):
Thanks @jeanbaptistedalle ! I added some suggestions, but please feel free to ignore them and let me know if we should proceed to take this PR.

sfmskywalker (approved):
Thank you!

Issue #5207, opened 2024-04-09
Batch publishing definitions which reference eachother can lead to not found activities in the designer.


raymonddenhaan:
This happens especially when having the MassTransit dispatcher enabled for the workflow management feature.
The publish action fails for some of the workflows due to an exception while updating the ActivityRegistry.

![image](https://github.com/elsa-workflows/elsa-core/assets/155616759/b1cfe742-156b-4427-bd5d-5987c46728f7)

raymonddenhaan:
Fixed in #5208

PR #5208, opened 2024-04-09
Changed behaviour on publishing workflows

raymonddenhaan:
Instead of repopulation the entire ActivityRegistry, only add the new workflow definition.


raymonddenhaan:
Fixes #5207

PR #5218, opened 2024-04-12
Fix LogPersistenceMode default mapping logic

### LINKED ISSUE #5217: [BUG] Workflow Instance Data Missing Due to Default Log Persistence Mode Misinterpretation
When initiating a workflow instance that involves multiple steps, it is observed that the expected data and logs are not present in the journal and activity instances. This issue stems from a misinterpretation regarding the default setting for the Log Persistence Mode.

## Expected Behavior

The expectation is that starting a workflow would automatically include detailed logs and data in the journal and associated activity instances, allowing for easy monitoring and debugging.

## Current Behavior

Due to a misinterpretation, the Default Log Persistence Mode is interpreted as `Exclude` by default, rather than `Include`. This results in missing crucial logs and data that are essential for effective workflow management and analysis.

## Steps to Reproduce

1. Configure the Workflow Management feature with DefaultLogPersistenceMode = Default
1. Start a workflow instance with multiple steps.
2. Check the journal and output logs in the associated activity instances.
3. Notice the absence of expected logs and data.

## Impact

This issue hinders the ability to effectively monitor and troubleshoot workflows, impacting overall system reliability and user experience.

sfmskywalker:
Fixes #5217

### DIFF
--- a/src/modules/Elsa.Workflows.Runtime/Services/DefaultActivityExecutionMapper.cs
+++ b/src/modules/Elsa.Workflows.Runtime/Services/DefaultActivityExecutionMapper.cs
@@ -11,15 +11,9 @@ using Microsoft.Extensions.Options;
 namespace Elsa.Workflows.Runtime.Services;
 
 /// <inheritdoc />
-public class DefaultActivityExecutionMapper : IActivityExecutionMapper
+public class DefaultActivityExecutionMapper(IOptions<ManagementOptions> options) : IActivityExecutionMapper
 {
-    private LogPersistenceMode _serverLogPersistenceMode;
     private const string LogPersistenceModeKey = "logPersistenceMode";
-
-    public DefaultActivityExecutionMapper(IOptions<ManagementOptions> options)
-    {
-        _serverLogPersistenceMode = options.Value.LogPersistenceMode;
-    }
 
     /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


src/modules/Elsa.Workflows.Runtime/Services/DefaultActivityExecutionMapper.cs
```diff
@@ -95,39 +83,35 @@ public ActivityExecutionRecord Map(ActivityExecutionContext source)
...
+        var persistencePropertyDefault = properties!.GetValueOrDefault("default", defaultFactory);
+
+        if(persistencePropertyDefault == LogPersistenceMode.Default) 
+            return defaultFactory();
+        return persistencePropertyDefault; 
+    }
+
+    private static Dictionary<string,object?> StorePropertyUsingPersistenceMode(IDictionary<string,object?> inputs,
+        IDictionary<string,object> persistenceModeConfiguration,
+        LogPersistenceMode defaultLogPersistenceMode = LogPersistenceMode.Exclude)
```
raymonddenhaan, line 100: Should the default not be Include?

sfmskywalker: Doesn't matter, the parameter is always supplied. I will remove the default value.

raymonddenhaan (approved):

PR #5221, opened 2024-04-12
Bulk BulkDispatchWorkflows fixes

MariusVuscanNx:
- Fixed "Item" use in correlation Id function. Previously the Item property was not accesible.
- Fixed "An item with the same key has already been added. Key: Error". This happened when both workflows that we were trying to dispatch failed during dispatching and they both tried to add the "Error" key in the journal.
- Fixed 'Suspended' bulk dispatch activity bug: If dispatching fails the workflow still waits for the children to complete. The behavior was changed to wait only after the ones for which dispatching succeeded.


sfmskywalker (approved):
Awesome!

PR #5243, opened 2024-04-18
Optimize Workflow Execution and Messaging

### LINKED ISSUE #5241: [PERF] Optimize Message Handling in `IWorkflowInbox` to Reduce Workflow Dispatch Latency
When initiating multiple dispatched workflows concurrently, there is a significant delay before all workflows are processed. This delay appears to stem from excessive messaging within the `IWorkflowInbox` service, which is tasked with broadcasting events to resume the `DispatchWorkflow` activity.

**Issue Details:**
- During the execution of dispatched workflows, an excessive number of messages are generated.
- These messages are often unnecessary as they do not correlate with any actionable bookmarks in the database.

**Proposed Solution:**
To alleviate the processing delay and optimize system performance, we should modify the `IWorkflowInbox` service to conditionally send messages. Specifically, the service should check for the presence of relevant bookmarks in the database before issuing messages. This change aims to prevent the queue from being filled with redundant messages that do not contribute to workflow processing.

**Expected Benefits:**
- Reduced latency in the processing of dispatched workflows
- Decreased load on messaging infrastructure
- More efficient resource utilization

This optimization should lead to a more responsive and efficient workflow dispatch mechanism, enhancing overall system performance.

sfmskywalker:
This PR introduces several optimizations and improvements to enhance the performance and reliability of Elsa Workflows:

- **Optimized Workflow Dispatching:** Modifies the workflow inbox service to first identify all relevant workflow instances before dispatching execution requests. This change ensures that "dispatch workflow" messages are only sent when applicable, significantly reducing unnecessary message traffic.
  
- **Enhanced Message Handling:** Ensures workflow inputs are stored with the workflow instance prior to dispatching messages to the queue. This approach prevents potential message rejection due to input mismatches at the service bus level.

- **Robust JSON Serialization:** Improves the custom JSON serializer to handle `TargetTypeExceptions` more effectively. This enhancement prevents errors during deserialization when the target type is not suitable.

- **Streamlined Application Startup:** Reduces redundancy in the workflow startup process by indexing workflow triggers only once after all workflows are populated, thereby improving the startup time.

These changes aim to streamline workflow processing, reduce overhead, and improve overall system stability.

Fixes #5241


src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
```diff
@@ -145,33 +159,32 @@ private async Task AddOrUpdateCoreAsync(MaterializedWorkflow materializedWorkflo
...
         workflowDefinition.MaterializerContext = materializerContextJson;
         workflowDefinition.MaterializerName = materializedWorkflow.MaterializerName;
-        
-        if (existingDefinitionVersion is null
-            && workflowDefinitionsToSave.Any(w => w.Id == workflowDefinition.Id))
+
+        if (existingDefinitionVersion is null && workflowDefinitionsToSave.Any(w => w.Id == workflowDefinition.Id))
         {
-            _logger.LogError("Trying to create a new workflow with existing id {workflowId}", workflowDefinition.Id);
+            _logger.LogInformation("Workflow with ID {WorkflowId} already exists", workflowDefinition.Id);
```
raymonddenhaan, line 165: If the workflow (definition) already exists wouldn't the `existingDefinitionVersion` not be null?

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
```diff
@@ -49,23 +49,37 @@ public class DefaultWorkflowDefinitionStorePopulator : IWorkflowDefinitionStoreP
...
             var results = await provider.GetWorkflowsAsync(cancellationToken).AsTask().ToList();
 
-            foreach (var result in results) await AddAsync(result, cancellationToken);
+            foreach (var result in results) await AddAsync(result, indexTriggers, cancellationToken);
         }
     }
 
     /// <inheritdoc />
-    public async Task AddAsync(MaterializedWorkflow materializedWorkflow, CancellationToken cancellationToken = default)
+    public Task AddAsync(MaterializedWorkflow materializedWorkflow, CancellationToken cancellationToken = default)
```
raymonddenhaan, line 70: Why not give indexTriggers in the next method a default of `true`? Than this method would not be needed

src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
```diff
@@ -74,31 +77,97 @@ await sendEndpoint.Send(new DispatchWorkflowInstance(request.InstanceId)
...
+                CorrelationId = request.CorrelationId
+            };
+
+            // The workflow instance is created and persisted in the database.
+            var workflowInstance = await workflowInstanceManager.CreateWorkflowInstanceAsync(createWorkflowInstanceRequest, cancellationToken);
+
+            // The workflow instance is then dispatched for execution.
+            var sendEndpoint = await GetSendEndpointAsync(options);
+            var message = DispatchWorkflowDefinition.DispatchExistingWorkflowInstance(workflowInstance.Id, trigger.ActivityId);
+            await sendEndpoint.Send(message, cancellationToken);
```
raymonddenhaan, line 129: This part is the same as line 50-56

src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
```diff
@@ -47,7 +48,11 @@ public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonS
             {
                 return JsonSerializer.Deserialize(ref reader, targetType, newOptions)!;
             }
-            catch (NotSupportedException e)
+            catch (NotSupportedException)
```
raymonddenhaan, line 51:
```suggestion
            catch (Exception e) when (e is NotSupportedException or TargetException)
```

raymonddenhaan (approved):

src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
```diff
@@ -74,31 +77,97 @@ await sendEndpoint.Send(new DispatchWorkflowInstance(request.InstanceId)
...
+                    continue;
+                }
+
+                if (input != null) workflowInstance.WorkflowState.Input.Merge(input);
+                if (properties != null) workflowInstance.WorkflowState.Properties.Merge(properties);
+
+                await workflowInstanceManager.SaveAsync(workflowInstance, cancellationToken);
+            }
+
+            var dispatchInstanceRequest = new DispatchWorkflowInstanceRequest(bookmark.BookmarkId) { BookmarkId = bookmark.BookmarkId, CorrelationId = bookmark.CorrelationId };
```
raymonddenhaan, line 166:
```suggestion
           var workflow = await workflowDefinitionService.FindWorkflowAsync(trigger.WorkflowDefinitionVersionId, cancellationToken);

            if (workflow == null)
            {
                logger.LogWarning("Workflow definition with ID '{WorkflowDefinitionId}' not found", trigger.WorkflowDefinitionVersionId);
                continue;
            }

            var createWorkflowInstanceRequest = new CreateWorkflowInstanceRequest
            {
                Workflow = workflow,
                WorkflowInstanceId = request.WorkflowInstanceId,
                Input = request.Input,
                Properties = request.Properties,
                CorrelationId = request.CorrelationId
            };

            // The workflow instance is created and persisted in the database.
            var workflowInstance = await workflowInstanceManager.CreateWorkflowInstanceAsync(createWorkflowInstanceRequest, cancellationToken);

            // The workflow instance is then dispatched for execution.
            var sendEndpoint = await GetSendEndpointAsync(options);
            var message = DispatchWorkflowDefinition.DispatchExistingWorkflowInstance(workflowInstance.Id, trigger.ActivityId);
```

raymonddenhaan: This suggestions fixes sending the workflowInstanceId instead of the bookmarkId as the instanceId. Preventing resuming the workflows.

raymonddenhaan (changes requested):

src/modules/Elsa.Workflows.Runtime/Services/DefaultWorkflowDefinitionStorePopulator.cs
sfmskywalker: What about callers that want the default of `true`, but not the default of `CancellationToken`? They would have to do this:

```csharp
await populator.AddAsync(MaterializedWorkflow materializedWorkflow, cancellationToken: cancellationToken);
```

With this convenient overload, they can now do this:

```csharp
await populator.AddAsync(MaterializedWorkflow materializedWorkflow, cancellationToken);
```

src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
sfmskywalker: Nice one 👍🏻

src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
sfmskywalker: That's a really good catch!

Issue #5262, opened 2024-04-23
[PERF] Journal Data is logging to redundant information in WorkflowExecutionLogRecords


jdevillard:
### Performance Issue Overview
**Is your performance issue related to a specific functionality? Please describe.**
This performance issue is related to the Storage Capacity of the LogRecord for each activity. The database can growth quickly with some data that are not used or redundant with other data.

### Proposed Enhancement
**Remove redundant Data of the `SerializedPayload` of the  `WorkflowExecutionLogRecords` table**

The SerializedPayload column store the Payload Data which is composed of all the Journal Data of the Activity. 
This Journal Data can be composed of : 
- Outcome
- Exception
- any specific Journalisation data referenced in the activity.

### Alternative Solutions
**Describe alternatives you've considered**
Have you identified any potential fixes or tweaks yourself? Please share your findings, including why they might not have been a perfect or complete solution. This helps us understand the issue better and consider all possible solutions.

### Affected Use Cases
**Identify affected use cases**
The use of the `Elsa.FlowSendHttpRequest` add all the parsed content of the Http call in the journal. This can be very consequent.

But Any activity that create some output will add data to the journal.

### Impact of Enhancement
**Explain the potential impact**
avoid to much storage, enhance performance due to less data in network call to the persistence store.

### Benchmarks and Metrics
**Provide any relevant benchmarks or metrics**

![image](https://github.com/elsa-workflows/elsa-core/assets/3254429/61a852c4-6938-4004-afd0-cb7e9450a05c)

For example, in my Http Sample,  the cost of a record could go from 683 bytes to 505 923 bytes . And this is visible for all activities that create some output.

### Additional Context
**Further though**

It could be useful to also allow selection to what can be stored using the journal. (output are already chosen with the  Persistence Log Record Module and stored in the SerializedActivityState

jdevillard:
closed as done with #5266

PR #5266, opened 2024-04-23
Remove SerializedOutputValue from JournalData

jdevillard:
relative to #5262

This PR remove the SerializedOutput Value from the Journal Data , This avoid duplication of information in the `WorkflowExecutionLogRecords` table.


sfmskywalker (approved):
Thank you 😄

PR #5273, opened 2024-04-24
Cache JavaScript Expressions

### LINKED ISSUE #5272: [PERF] Cache JavaScript AST for Workflow Definitions
**Problem:**
Currently, each execution of a workflow definition repeatedly parses the same JavaScript expressions, which leads to inefficiencies, especially when a workflow is executed multiple times.

**Proposed Solution:**
To enhance performance, it is proposed to implement caching of the parsed Abstract Syntax Tree (AST) of JavaScript expressions used within a workflow definition. This cache would be utilized across multiple executions of the same workflow, avoiding redundant parsing operations.

**Expected Benefits:**
- **Reduced Parsing Overhead:** By caching the AST, we minimize the computational cost associated with parsing the same expressions multiple times.
- **Improved Execution Speed:** Workflow execution times should decrease due to reduced processing time.
- **Enhanced Scalability:** Reduces the load on the workflow engine, allowing it to scale better under high demand.

This feature would require modifications to the JavaScript expression handling mechanism to enable caching and retrieval of ASTs. Additionally, considerations for cache invalidation and updates when the workflow definition changes would be necessary to ensure accuracy and consistency.

sfmskywalker:
This PR implements parsed JavaScript expressions for improved performance when executing workflows multiple times.

Fixes #5272

### DIFF
--- a/src/modules/Elsa.JavaScript/Services/JintJavaScriptEvaluator.cs
+++ b/src/modules/Elsa.JavaScript/Services/JintJavaScriptEvaluator.cs
@@ -201,7 +201,10 @@ public class JintJavaScriptEvaluator : IJavaScriptEvaluator
     [RequiresUnreferencedCode("Calls System.Text.Json.JsonSerializer.Serialize<TValue>(TValue, JsonSerializerOptions)")]
     private static string Serialize(object value)
     {
-        var options = new JsonSerializerOptions { Encoder = JavaScriptEncoder.Create(UnicodeRanges.All) };
+        var options = new JsonSerializerOptions
+        {
+            Encoder = JavaScriptEncoder.Create(UnicodeRanges.All)
+        };
         options.Converters.Add(new JsonStringEnumConverter());
 
         return JsonSerializer.Serialize(value, options);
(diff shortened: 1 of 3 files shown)


MariusVuscanNx (approved):

src/modules/Elsa.JavaScript/Services/JintJavaScriptEvaluator.cs
```diff
@@ -176,9 +180,21 @@ private static void CreateActivityOutputAccessors(Engine engine, ExpressionExecu
...
 
-    private static object? ExecuteExpressionAndGetResult(Engine engine, string expression)
+    private object? ExecuteExpressionAndGetResult(Engine engine, string expression)
     {
-        var result = engine.Evaluate(expression);
+        var cacheKey = "jint:script:" + expression.GetHashCode(StringComparison.Ordinal);
+
+        var parsedScript = _memoryCache.GetOrCreate(cacheKey, entry =>
+        {
+            if (_jintOptions.ScriptCacheTimeout.HasValue)
```
raymonddenhaan, line 189: If no cache timeout is given the item will stay in the cache indefinitly?

sfmskywalker: Yes. And this will only happen if the user explicitly provides a null value for the timeout setting. In other words, out of the box, there will always be a timeout (which defaults to 1 day).

Issue #5313, opened 2024-05-01
[ENH] Include dropdown when using a nullable enum in an activity input


nickrubenstein:
### Enhancement Overview
Using the `InputUIHints.DropDown` on an Activity Input with an `enum` displays the dropdown with all the options in the studio activity editor as expected. 

Changing the Input `enum` to be nullable makes the dropdown no longer show any options and does not work since it is a dropdown without any options to select.

### Proposed Enhancement
Adding a few lines in `StaticDropDownOptionsProvider` class to check if `wrappedPropertyType` is nullable and has an underlying type that could be an `enum`. Then, add a blank item option in the `enumValues` list. I have taken a crack at it myself since it was so easy to run and debug elsa-core! It is such a small change and I do not know what this change could impact, but I can go through the PR process if requested.

### Use Cases
I have an activity that queries data. There is a `status` property that is an `enum` that I would like to be able to filter by. In the meantime, I made a generic `NullableEnumUIHandler` to get around this and have a few type castings. It would definitely be cleaner code to simply add the `InputUIHints.DropDown` with the code below. Any pointers on a better way of having an `enum` input that can be changed back to being null after selecting an option would be great.

### Additional Context

This generates an input with a dropdown of the enum values:
```
[Input(
    UIHint = InputUIHints.DropDown
)]
public Input<TestEnum> TestInput { get; set; } = new(default(TestEnum));
```

This ends up having a blank dropdown where nothing is selectable:

```
[Input(
    UIHint = InputUIHints.DropDown
)]
public Input<TestEnum?> TestNullableInput { get; set; } = new(default(TestEnum?));
```

PR #5320, opened 2024-05-02
Include dropdown when using a nullable enum in an activity input

nickrubenstein:
This PR is for https://github.com/elsa-workflows/elsa-core/issues/5313

Adds a check for a nullable enum type in `StaticDropDownOptionsProvider` so a dropdown works with nullable enums.


nickrubenstein:
@dotnet-policy-service agree

sfmskywalker (approved):
Awesome 👍🏻

PR #5337, opened 2024-05-03
Bugfixes for heartbeats using MongoDB

stephanmelzer:
Using Elsa with MongoDB we encountered two errors during the heartbeat functionality.

The first error was: `System.ArgumentException: Instance property 'Id' is not defined for type 'Elsa.KeyValues.Entities.SerializedKeyValuePair' (Parameter 'propertyName')` in `Elsa.Hosting.Management.HostedServices.InstanceHeartbeatService`.
Here the `Elsa.MongoDb.Modules.Runtime.MongoKeyValueStore` tries to save a SerializedKeyValuePair instance using an ID filter, but SerializedKeyValuePair does not have this property.
Only the other entities do have an ID property through the Entity base class.
I specified a key selector.

The second error was: `System.FormatException: Element '_id' does not match any field or property of class Elsa.KeyValues.Entities.SerializedKeyValuePair.` in `Elsa.Hosting.Management.HostedServices.InstanceHeartbeatMonitorService`.
This time MongoDB could not load the SerializedKeyValuePair instances, because no mapping for the missing ID property could be found.
I added a BISON mapping to ignore not mappable fields.


stephanmelzer:
@dotnet-policy-service agree company="MERA Tiernahrung GmbH"

sfmskywalker (approved):
Awesome 👍🏻

sfmskywalker (approved):
Great, thanks!

PR #5352, opened 2024-05-08
Add possibility to use dynamically loaded assemblies

truthz03:
Because I need a way to load assemblies dynamically (plugins) in my project, I found some problems while serializing/deserializing payloads.
In order to solve this problem, I would like to implement a mechanism to resolve the necessary types manually if needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5352)


truthz03:
@dotnet-policy-service agree

cvijayak:
@truthz03 

Please confirm whether the commit "https://github.com/elsa-workflows/elsa-core/pull/5352/commits/367f5749e8d5676ae10de5abb2003b4781ef6fba" will resolve the below bug 

 https://github.com/elsa-workflows/elsa-core/issues/5356

truthz03:
@cvijayak 
No, my commit will not fix this bug. My commit gives the possibility to deserialize stored things into unknown objects.
This means, if I add some activities and there payloads as plugin (within a different AppDomain), it will be possible to deserialize the json content to a type only known by the plugin AppDomain.

The bug #5356 shows problems with the serialization. It throws an error because the type is `System.Type`, which is a known type. I also don't know why System.Type is not serializeable.

truthz03:
Is there something to do for my to get the "SonarCloud Code Analysis" ready?
Or have I still to wait?

sfmskywalker:
@truthz03 Thanks for your PR! At this point there's nothing for you to do, I just need time to review and test. Thank you for your patience.

truthz03:
@sfmskywalker Hi, can you tell me how long it will take to get my PR merged?

sfmskywalker:
@truthz03 Hey there, sorry it's been taking a while. I will get this merged this week. Thanks again for your contribution, your PR makes sense!

sfmskywalker:
@truthz03 Unfortunately, the main branch received an update that uses the `PolymorphicObjectConverterFactory` as a type in a `JsonConverterAttribute` - which only works with parameterless constructors.

In other words, we need to revisit either that change, or your change, such that you can still resolve services from DI.

truthz03:
@sfmskywalker Ok. Let me know if you find a solution.
Maybe its possible to have 2 constructors. One parameterless for the attribute where it is not possible to inject anything and one with parameters.

PR #5422, opened 2024-05-22
Catch the exception and handling the exception in the workflow

fanslead:
Implemented a catch activity to catch the exception in the flowchart workflow.
When the flowchart has a catch activity, it will catch the exception and can excute the custom exception handling workflow.
image
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5422)


fanslead:
@dotnet-policy-service agree

fanslead:
@sfmskywalker any suggestion for this pr?

fanslead:
@sfmskywalker any suggestion for this pr?

sfmskywalker:
Hi @fanslead , thank you for your PR. I will need some time to think about this approach and to explore your PR in detail. Thank you for your patience.

fanslead:
> Hi @fanslead , thank you for your PR. I will need some time to think about this approach and to explore your PR in detail. Thank you for your patience.

@sfmskywalker  I think this will be a very practical, commonly used and flexible feature. For example, in the event of an abnormality, execute custom event or http notifications or compensation measures without any coding.

fanslead:
@sfmskywalker Hi, Do you have any thoughts about this PR?

sfmskywalker:
Hi @fanslead , I'm afraid I haven't had the chance yet to sit down and look through the PR and think about the proposed feature, but generally I think it makes sense. I can't say when I might be able to circle back to this, but please rest assured that your PR will be reviewed and that you'll get proper feedback. Thank you for your patience.

fanslead:
@sfmskywalker  any update for this pr?

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -28,6 +28,7 @@ public Flowchart([CallerFilePath] string? source = default, [CallerLineNumber] i
         OnSignalReceived<ScheduleActivityOutcomes>(OnScheduleOutcomesAsync);
         OnSignalReceived<ScheduleChildActivity>(OnScheduleChildActivityAsync);
         OnSignalReceived<CancelSignal>(OnActivityCanceledAsync);
+        OnSignalCaptured<ExceptionSignal>(OnActivityExceptionAsync);
```
sfmskywalker, line 31: I believe I had removed the "capturing" phase - do we really need it for this?

src/modules/Elsa.Workflows.Core/Activities/Catch.cs
```diff
@@ -0,0 +1,37 @@
...
+[Activity("Elsa", "Primitives", "Catch the Exception.")]
+public class Catch : Activity
+{    
+    /// <inheritdoc />
+    public Catch([CallerFilePath] string? source = default, [CallerLineNumber] int? line = default) : base(source, line)
+    {
+    }
+
+    /// <summary>
+    /// Catch all the Exception.
```
sfmskywalker, line 20:
```suggestion
    /// A value indicating whether to catch all exceptions from all activities.
```

src/modules/Elsa.Workflows.Core/Activities/Catch.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    [Input(Description = "A value that is the tag a global catcher")]
```
sfmskywalker, line 22:
```suggestion
    [Input(Description = "A value indicating whether to catch all exceptions from all activities.")]
```

src/modules/Elsa.Workflows.Core/Activities/Catch.cs
```diff
@@ -0,0 +1,37 @@
...
+    public Input<bool> CatchAll { get; set; } = default!;
+
+    /// <summary>
+    /// Catch the Activities.
+    /// </summary>
+    [Input(Description = "The Activities will be catch",
+        UIHint = InputUIHints.MultiText)]
```
sfmskywalker, line 29:
```suggestion
    [Input(
        Description = "The IDs of the activities to catch exceptions from.",  
        UIHint = InputUIHints.MultiText)]
```

src/modules/Elsa.Workflows.Core/Activities/Catch.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /// <summary>
+    /// Catch all the Exception.
+    /// </summary>
+    [Input(Description = "A value that is the tag a global catcher")]
+    public Input<bool> CatchAll { get; set; } = default!;
+
+    /// <summary>
+    /// Catch the Activities.
```
sfmskywalker, line 26:
```suggestion
    /// The IDs of the activities to catch exceptions from.
```

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -325,4 +326,55 @@ private async ValueTask OnActivityCanceledAsync(CancelSignal signal, SignalConte
     {
         await CompleteIfNoPendingWorkAsync(context.ReceiverActivityExecutionContext);
     }
+
+    private async ValueTask OnActivityExceptionAsync(ExceptionSignal signal, SignalContext context)
+    {
+        var avtivities = Activities.Where(a => a is Catch).Select(a => a as Catch).ToList();
```
sfmskywalker, line 332:
```suggestion
        var activities = Activities.Where(a => a is Catch).Cast<Catch>().ToList();
```

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -325,4 +326,55 @@ private async ValueTask OnActivityCanceledAsync(CancelSignal signal, SignalConte
...
+
+        if (avtivities.Count == 0)
```
sfmskywalker, line 334:
```suggestion
        if (activities.Count == 0)
```

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -325,4 +326,55 @@ private async ValueTask OnActivityCanceledAsync(CancelSignal signal, SignalConte
...
+        var workflowExecutionContext = flowchartContext.WorkflowExecutionContext;
+
+        var exceptionActivity = signal.Activity;
+
+        var options = new ActivityInvocationOptions
+        {
+            Owner = flowchartContext
+        };
+
+        foreach (var catchAvtivity in avtivities)
```
sfmskywalker, line 347:
```suggestion
        foreach (var catchActivity in activities)
```

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -325,4 +326,55 @@ private async ValueTask OnActivityCanceledAsync(CancelSignal signal, SignalConte
...
+                    await flowchartContext.ScheduleActivityAsync(catchAvtivity, scheduleWorkOptions);
+            }
+            else
+            {
+                var catchActivities = catchAvtivity.CatchActivities?.Get(catchContext);
+                if (catchActivities != null && catchActivities.Contains(exceptionActivity.Id))
+                    await flowchartContext.ScheduleActivityAsync(catchAvtivity, scheduleWorkOptions);
+            }
+        }
+    }
```
sfmskywalker, line 379: I'm not entirely sure about this approach. Although I understand that you have to evaluate the properties of the `Catch` activity in order to see if the faulting activity ID is present in its list and whether the CatchAll is set to true, I would think that, after it is determined that the CatchAll activity should execute, you should simply schedule the catch activity for execution. I.e, I think that if either `catchAll` is true _or_ the faulting activity is present in the "catch activities" list, you should schedule the catch activity. 

E.g.

```suggestion
        {
            var catchContext = workflowExecutionContext.ActivityExecutionContexts.FirstOrDefault(x => x.Activity.Id == catchAvtivity!.Id);

            if (catchContext == null)
            {
                catchContext = workflowExecutionContext.CreateActivityExecutionContext(catchAvtivity!, options);
                workflowExecutionContext.AddActivityExecutionContext(catchContext);

                await catchContext.EvaluateInputPropertiesAsync();
            }

            var scheduleWorkOptions = new ScheduleWorkOptions
            {
                CompletionCallback = OnChildCompletedAsync,
                ExistingActivityExecutionContext = catchContext
            };

            var catchAll = catchActivity!.CatchAll.Get(catchContext);
            var catchActivities = catchActivity.CatchActivities?.Get(catchContext);
            var catchFaultingActivity = catchActivities?.Contains(exceptionActivity.Id) == true;
            var shouldCatch = catchAll || catchFaultingActivity;
            
             if(shouldCatch) 
                    await flowchartContext.ScheduleActivityAsync(catchActivity, scheduleWorkOptions);
        }
    }
```

Please note that I haven't tested it, so I could be totally off here.

src/modules/Elsa.Workflows.Core/Middleware/Activities/ExceptionHandlingMiddleware.cs
```diff
@@ -60,6 +62,8 @@ public async ValueTask InvokeAsync(ActivityExecutionContext context)
 
             var strategy = await _incidentStrategyResolver.ResolveStrategyAsync(context);
             strategy.HandleIncident(context);
+
+            await context.SendSignalAsync(new ExceptionSignal(activity, e));
```
sfmskywalker, line 66: I'm questioning whether we should report an incident if an exception was handled. What are your thoughts on that?

src/modules/Elsa.Workflows.Core/Signals/ExceptionSignal.cs
```diff
@@ -0,0 +1,10 @@
+using Elsa.Workflows.Contracts;
+
+namespace Elsa.Workflows.Signals;
+
+/// <summary>
+/// /// Sent by ExceptionHandlingMiddleware to notify their composite container that it has exception.
```
sfmskywalker, line 6:
```suggestion
/// Sent by ExceptionHandlingMiddleware to notify their composite container that an exception occurred.
```

sfmskywalker (commented):
All in all, I can see that the `Catch` activity could be interesting for handling exceptions centrally.

However, I think that we could do something nicer that is similar to BPMN's [boundary event](https://docs.camunda.io/docs/components/modeler/bpmn/signal-events/#signal-intermediate-catch-events) mechanisms.

Although it definitely doesn't have to be the same, I think that we might do the following:

1. Introduce a new activity called `Try`
2. The `Try` activity exposes two outcomes:
    1. `Done`
    2. `Catch`
3. The `Try` activity has one embedded port called `Body` and is of type `IActivity` - effectively making this a composite activity.
4. Similar what you did with `Flowchart`, the `Try` activity would listen for exception signals.
5. When an exception signal is received, it completes with the `Catch` outcome.
6. Otherwise, it completes with the `Done` outcome.

This way, we have a kind of a "boundary event" in the form of the "Catch" outcome on the `Try` activity, achieving the same centrally managed exception handling but without having to update the `Flowchart` activity.

What do you think?

fanslead:
> All in all, I can see that the `Catch` activity could be interesting for handling exceptions centrally.
> 
> However, I think that we could do something nicer that is similar to BPMN's [boundary event](https://docs.camunda.io/docs/components/modeler/bpmn/signal-events/#signal-intermediate-catch-events) mechanisms.
> 
> Although it definitely doesn't have to be the same, I think that we might do the following:
> 
> 1. Introduce a new activity called `Try`
> 2. The `Try` activity exposes two outcomes:
>    
>    1. `Done`
>    2. `Catch`
> 3. The `Try` activity has one embedded port called `Body` and is of type `IActivity` - effectively making this a composite activity.
> 4. Similar what you did with `Flowchart`, the `Try` activity would listen for exception signals.
> 5. When an exception signal is received, it completes with the `Catch` outcome.
> 6. Otherwise, it completes with the `Done` outcome.
> 
> This way, we have a kind of a "boundary event" in the form of the "Catch" outcome on the `Try` activity, achieving the same centrally managed exception handling but without having to update the `Flowchart` activity.
> 
> What do you think?

This is a good idea. But I think composite activity will to some extent affect the readability of the flow. In my eyes, Catch activities should be detached from the main workflow and receive the signal from the main workflow.  That is more like to BPMN's boundary event mechanisms. Of course, we can implement both options and leave it to users to choose and use.Just like the Switch Activity and the Switch(flow) Activity.

fanslead:
> All in all, I can see that the `Catch` activity could be interesting for handling exceptions centrally.
> 
> However, I think that we could do something nicer that is similar to BPMN's [boundary event](https://docs.camunda.io/docs/components/modeler/bpmn/signal-events/#signal-intermediate-catch-events) mechanisms.
> 
> Although it definitely doesn't have to be the same, I think that we might do the following:
> 
> 1. Introduce a new activity called `Try`
> 2. The `Try` activity exposes two outcomes:
>    
>    1. `Done`
>    2. `Catch`
> 3. The `Try` activity has one embedded port called `Body` and is of type `IActivity` - effectively making this a composite activity.
> 4. Similar what you did with `Flowchart`, the `Try` activity would listen for exception signals.
> 5. When an exception signal is received, it completes with the `Catch` outcome.
> 6. Otherwise, it completes with the `Done` outcome.
> 
> This way, we have a kind of a "boundary event" in the form of the "Catch" outcome on the `Try` activity, achieving the same centrally managed exception handling but without having to update the `Flowchart` activity.
> 
> What do you think?

This is a good idea. But I think composite activity will to some extent affect the readability of the flow. In my eyes, Catch activities should be detached from the main workflow and receive the signal from the main workflow.  That is more like to BPMN's boundary event mechanisms. Of course, we can implement both options and leave it to users to choose and use.Just like the Switch Activity and the Switch(flow) Activity.

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
fanslead: When catchAll is true, we need to consider whether abnormal activities are executed after catch activities. Otherwise, an infinite loop may occur.

src/modules/Elsa.Workflows.Core/Middleware/Activities/ExceptionHandlingMiddleware.cs
fanslead: This is a good question. I think if the user chooses Continue With Incidents Strategy, does he still need to add Catch Activity.If Catch Activity is not added, there will be no impact. If it is added, it means he needs this report.

rolandgriesser: I'd like to upvote the reporting of an incident. Right now, I handle it using a custom `IncidentStrategy`. Same goes for `Faulted` workflows, right now I handle them in the `WorkflowFinished` using the `SubStatus.Faulted`.

```
[Description("An incident strategy that allows the workflow to continue running and reporting the incidents using a notification.")]
    public class NotifyOnIncidentsStrategy(IMediator mediator, IWorkflowStateExtractor workflowStateExtractor) : IIncidentStrategy
    {
        public void HandleIncident(ActivityExecutionContext context)
        {
            var workflowExecutionContext = context.WorkflowExecutionContext;
            var workflow = workflowExecutionContext.Workflow;
            var workflowState = workflowStateExtractor.Extract(workflowExecutionContext);
            var incident = workflowState.Incidents?.LastOrDefault();

            _ = mediator.SendAsync(new WorkflowIncident(workflow, workflowState, workflowExecutionContext, incident));
        }
    }
```

PR #5425, opened 2024-05-22
Revert sending input as global state instead of parameter

### LINKED ISSUE #5417: [BUG] Concurrency Issue in BulkDispatchWorkflows Activity Prevents Completion
## Summary
The `BulkDispatchWorkflows` activity fails to complete due to a concurrency issue introduced during an update. This issue arises when updating the workflow instance to store workflow input before sending a message to the service bus.

## Problem
When we implemented support for storing workflow input with the workflow instance to prevent large message sizes, we inadvertently created a situation where multiple threads could attempt to update the workflow instance state simultaneously. This lack of thread synchronization has led to the `BulkDispatchWorkflows` activity failing to complete.

## Solution
To resolve this issue, we need to acquire a distributed lock on the workflow instance before updating it with the input. This will ensure that only one thread can update the workflow instance at a time, preventing concurrency conflicts.

sfmskywalker:
This PR fixes a concurrency issue when sending concurrent workflow execution requests to the same workflow instance with different inputs.

Originally, this change was made to add support for sending large inputs from a parent workflow to a dispatched child workflow via a service bus.

However, that doesn't work when sending multiple messages concurrently, which we are seeing when using BulkDispatchWorkflow where multiple child workflows complete close to one another, overriding the "global" input state associated with the workflow input.

This PR attempts to only store the input with the workflow instance upon instance creation, while sending input alongside the dispatch message when resuming existing ones. The assumption here being that those messages include smaller input payloads.

The correct fix will be to revise the need for associating workflow input with the workflow instance, and instead provide mechanism to stash input externally from the workflow instance and instead passing in an identifier to that input.

Fixes #5417

### DIFF
--- a/src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
+++ b/src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
@@ -1,4 +1,3 @@
-using Elsa.Extensions;
 using Elsa.MassTransit.Contracts;
 using Elsa.MassTransit.Messages;
 using Elsa.Workflows.Contracts;
@@ -66,7 +65,8 @@ public class MassTransitWorkflowDispatcher(
             ActivityNodeId = request.ActivityNodeId,
             ActivityInstanceId = request.ActivityInstanceId,
             ActivityHash = request.ActivityHash,
-            CorrelationId = request.CorrelationId
+            CorrelationId = request.CorrelationId,
+            Input = request.Input
         }, cancellationToken);
         return DispatchWorkflowResponse.Success();
     }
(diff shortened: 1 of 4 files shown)


src/modules/Elsa.MassTransit/Services/MassTransitWorkflowDispatcher.cs
```diff
@@ -163,28 +163,12 @@ private async Task DispatchBookmarksAsync(IEnumerable<StoredBookmark> bookmarks,
         {
             var workflowInstanceId = bookmark.WorkflowInstanceId;
 
-            if (input != null || properties != null)
-            {
-                // Need to acquire a lock on the workflow instance to prevent concurrent updates.
-                await using var distributedLock = await distributedLockProvider.AcquireLockAsync(workflowInstanceId, TimeSpan.FromMinutes(2), cancellationToken);
```
raymonddenhaan, line 169: Removing this makes the distributedLockProvider unused

raymonddenhaan (approved):

raymonddenhaan (changes requested):
Could you add tests for the scenario that should be fixed with this?

sfmskywalker:
@raymonddenhaan  I considered it, but my last attempt to try and reproduce a race condition using a component test didn't pan out. Perhaps its possible, but I'd need a lot more time to try and achieve that. I am open to suggestions in case you have any.

sfmskywalker:
Actually, I could try something similar to the `WorkflowDefinitionActivityTests` - it's a bit crude and doesn't guarantee the bug appears in case of a regression - but worked 99% of the times.

raymonddenhaan (approved):

raymonddenhaan:
Let's see if we can figure out how we can test these race conditions and if we can do it with 100% certainty without costing us a lot of time and resources in the build pipeline. This PR does not have to wait for that discussion.

Issue #5254, opened 2024-04-19
[PERF] Enhance Scalability and Simplify Workflow Execution with Proto.Actor


sfmskywalker:
### Objective
To improve the scalability and manageability of workflow execution, we propose refactoring the `IWorkflowRuntime` interface and fully adopting Proto.Actor as our primary execution runtime.

### Current Challenges
- The existing `IWorkflowRuntime` interface is responsible for both workflow lookup and execution, leading to code duplication and complexity across its implementations (`Default Runtime` and `ProtoActor Runtime`).
- Managing distributed workflow instance cancellation is cumbersome, requiring complex orchestration and distributed messaging.

### Proposed Changes
1. **Decouple Workflow Matching and Execution**: 
   - Remove workflow matching responsibilities from `IWorkflowRuntime`.
   - Introduce a `WorkflowMatcher` service to handle finding workflows based on stimuli (renamed from "bookmark + trigger hashes").

2. **Interface and Method Refinements**:
   - Add a `CreateWorkflowClient` method to `IWorkflowRuntime` to return an `IWorkflowClient`, serving as a proxy to the workflow instance.
   - Transition to using `Local Workflow Runtime` and `ProtoActor Workflow Runtime` to reflect different operational contexts.

3. **Terminology and Structural Adjustments**:
   - Rename "bookmark + trigger payloads" to "stimuli" to clarify their role in triggering workflow actions in response to external input (stimuli).
   - Workflows can execute in "local context" (e.g. current HTTP context) and in "remote context" (e.g. Proto.Actor Workflow Runtime).
   - A workflow instance can transition from local context to remote context and vice versa through explicit operations that are yet to be defined.

4. **Enhancements through Proto.Actor**:
   - **Lock-Free Execution**: Utilize Proto.Actor’s capabilities for lock-free workflow execution, significantly increasing throughput.
   - **Event Sourcing**: Leverage Proto.Actor's event sourcing to ensure reliability and recoverability, reducing the need for the current `WorkflowInboxMessages` DB table.
   - **Improved Pub/Sub**: Leverage Proto.Actor for faster event broadcasting compared to traditional event buses.
   - **Simplified Cancellation**: Enable more straightforward cancellation of workflows via Proto.Actor, even during a burst of execution.
   - **Performance Improvements**: Maintain workflow state in memory within actors, allowing for quicker resumption and response to new stimuli.

### Benefits
[...]

Issue #5227, opened 2024-04-16
MySQL DELETE ‘WorkflowInboxMessages’ ERROR[BUG]


493141993:
![20240416114112](https://github.com/elsa-workflows/elsa-core/assets/46142629/6764173f-2cdb-4900-a1e9-9be1d44e2d9f)

493141993:
MySql Version:8.0.11

cr0fters:
Worth mentioning, that if you downgrade Pomelo to 7.0.0 (you'll also need to be running on .Net 7) the issue goes away. Maybe worth waiting until they patch it on the Pomelo side rather than change the implementation here?

sfmskywalker:
This issue will be resolved automatically when we merge https://github.com/elsa-workflows/elsa-core/pull/5444. We'll close this issue once that happens.

Issue #5203, opened 2024-04-08
Single workflow event lifetime


Eitanski:
I want that once an event is raised, and has started unblocking, the same event will not be raised, which means that the next workflow instance will block on this same event activity.

Currently this is not the case, a single event remains raised for several workflow executions.

Also, when I publish the event I do not want to resolve the current workflow instance, I saw that it is an option.

sfmskywalker:
This is fixed as part of PR https://github.com/elsa-workflows/elsa-core/pull/5444 - We can close this issue when that PR gets merged into `main`.

PR #5444, opened 2024-05-24
Refactor Workflow Runtimes

### LINKED ISSUE #5409: [BUG] Infinite Looping in Event Activity in v3
## Description
We have a workflow with events 1, 2, 3, and 4.
The execution sequence 1 -> 2 -> 3 works as expected. However, once event 3 is completed, event 4 is triggered, which loops back to event 3. This is where the problem starts. After event 4 is completed, the workflow enters an infinite loop, continuously cycling through events 2, 3, and 4 without waiting for user intervention.
The workflow repeats in the following manner: 2 -> 3 -> 4 -> 2 -> 3 -> 4 -> 2 -> 3 -> 4, and so on. This looping happens without pausing to allow the user to manually complete the events.
Ideally, the execution should wait for the user to manually complete each event, even if the event has been completed before and is triggered again. The workflow should only proceed once the user has manually completed the event.
This is a minimum reproducible example. In a real-world scenario, there would be an if block before triggering event 4.

## Steps to Reproduce
To help us identify the issue more quickly, please follow these guidelines:

1. **Detailed Steps**: I am attaching video and json of that workflow so that you will have some idea about it.

3. **Attachments**:
   - **Workflow JSON**: 
[workflow-2.json](https://github.com/elsa-workflows/elsa-core/files/15347058/workflow-2.json)

4. **Video/Screenshots**: 

https://github.com/elsa-workflows/elsa-core/assets/79979623/25be56cf-a3cd-4ee4-aa19-8478cc00185a

Providing a comprehensive and clear description will greatly assist in diagnosing the problem quickly and accurately. If you're unable to share some of the requested information publicly, please note that, and we can arrange for secure information transfer.

## Expected Behavior
Workflow should not continue further automatically when it returns to the Event2. It should wait to complete the Event2. This flow was working perfectly in v2 of elsa.

## Actual Behavior
You can see it in the uploaded video it is autocompleting Event2 /Event3 and going into infinite loop.

sfmskywalker:
This PR refactors the workflow runtime APIs in a significant way.

Key Areas of focus:

1. **Workflow Invocation via Workflow Clients**
   - Targeting a specific instance in a given `IWorkflowRuntime` implementation, of which there are three:
     - **Local:** No locking, suitable only for single-node hosting.
     - **Distributed Locking:** Locking, suitable for clustered hosting but with limited throughput due to distributed locking.
     - **Actor Model:** Each workflow instance is its own actor instance, suitable for clustered hosting without distributed locking.
       - Can be optimized further to become database-less for maximum throughput, but this is out of scope for this PR. Additional enhancements will be in a separate repo owned by ELSA-X.
       - Distributed cancellation is only supported if a given runtime supports it (only the Actor Model will support this, which might be an extra incentive for users to move towards this model).

2. **Paradigm Shift Towards Sending Stimuli to the System**
   - New workflows will start and existing ones will resume when stimuli match triggers and bookmarks associated with workflow definitions and instances.

3. **HTTP Endpoint Workflows**
   - Execute "in context" using the `IWorkflowRunner` directly, ensuring fast execution without marshaling workflow instances to and from workflow runtimes. This means no locking for the distributed locking runtime and no back-and-forth for the Actor model runtime.

4. **WorkflowInbox Store No Longer Used**
   - Results in less DB data being written to and read from.

Fixes #5409 #5227 #5254 #5203

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5444)


raymonddenhaan:
SonarCloud has found some issues that are worth looking into.

PR #5446, opened 2024-05-24
Add Http Response Behavior to answer synchronously using the HttpResponse Activity

jdevillard:
This change aims to include the change of the PR #5268 and add a feature flag in the configuration to configure (by opt-in) if we want to respond synchronously or not using the HttpResponse activity (instead of after the Workflow Flush (at the end of the http Middleware)

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5446)


sfmskywalker (approved):

sfmskywalker:
@ruudkobes you're right, we should close them. Thanks for letting us know 👍🏻

Issue #5532, opened 2024-06-04
[BUG] Flowchart variables are not serialized by FlowchartJsonConverter


bobhauser:
## Description
When adding workflow to a `WorkflowDefinitionStore` using `DefaultWorkflowDefinitionStorePopulator`, flowchart variables are not serialized. This of course results in the workflow not executing properly

## Steps to Reproduce
This issue can be reproduced by running this code snippet. Notice that there is a variable called `TextVar` specified, which is used by the `WriteLine` activity.
```
Workflow testWorkflow = new Workflow()
{
	Id = "abc123",
	Identity = new WorkflowIdentity(
		DefinitionId: "123abc",
		Version: 1,
		Id: "abc123"
	),
	WorkflowMetadata = new WorkflowMetadata(
		Name: "TestWorkflow",
		ToolVersion: new Version(3, 0) // This, oddly, is required to use the activity name when getting results via javascript (getResultFromRunScript())
	),
	Root = new Flowchart()
	{
		Id = "flowchart",
		Variables = new List<Variable>() { 
			new Variable("TextVar", "This is the text to write") 
		},
		Activities = new List<IActivity>() {
			new WriteLine(new Input<string>(new Expression("JavaScript", "getVariable('TextVar')")))
			{
				Id = "writeLine",
				Name = "WriteLine",
				CanStartWorkflow = true,
			},
		},
		Connections = new List<Connection>()
	}
};

// save the workflow
IWorkflowDefinitionStorePopulator workflowDefinitionStorePopulator = scope.ServiceProvider.GetRequiredService<IWorkflowDefinitionStorePopulator>();
await workflowDefinitionStorePopulator.AddAsync(new MaterializedWorkflow(testWorkflow, "RapidResponse", JsonWorkflowMaterializer.MaterializerName));

// run the workflow
IWorkflowDispatcher dispatcher = scope.ServiceProvider.GetRequiredService<IWorkflowDispatcher>();
DispatchWorkflowResponse dispatchWorkflowResponse = await dispatcher.DispatchAsync(new Elsa.Workflows.Runtime.Requests.DispatchWorkflowDefinitionRequest()
{
	DefinitionId = "123abc",
});

```

This is the serialized flowchart - notice that the `TestVar` variable is not present in the resultant JSON
```
{
	"type": "Elsa.Flowchart",
	"version": 1,
	"id": "flowchart",
	"nodeId": "abc123:flowchart",
	"metadata": {},
	"customProperties": {
[...]

PR #5533, opened 2024-06-04
Flowchart variables are not serialized by FlowchartJsonConverter

bobhauser:
Add variables to the serialization/deserialization of Flowcharts.

See #5532 for more details. The bug report includes an example of the serialized Flowchart not containing the specified variable.

**Serialized Flowchart after this change (notice variables are included):**
```
{
	"type": "Elsa.Flowchart",
	"version": 1,
	"id": "flowchart",
	"nodeId": "abc123:flowchart",
	"metadata": {},
	"customProperties": {
		"source": "FlowchartTest.cs:126"
	},
	"activities": [
		{
			"text": {
				"typeName": "String",
				"expression": {
					"type": "JavaScript",
					"value": "getVariable('TextVar')"
				}
			},
			"id": "writeLine",
			"nodeId": "abc123:flowchart:writeLine",
			"name": "WriteLine",
			"type": "Elsa.WriteLine",
			"version": 1,
			"customProperties": {
				"source": "FlowchartTest.cs:131",
				"canStartWorkflow": true
			},
			"metadata": {}
		}
	],
	"connections": [],
	"variables": [
		{
			"id": "28f554ba2b294ce5b52456f89e755fa2",
			"name": "TextVar",
			"typeName": "Object",
			"value": "This is the text to write"
		}
	]
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5533)


sfmskywalker:
Your PR + issue makes sense. I have an additional thought; given that the `Variables` property is inherited from the `Composite` base class, I wonder if there is another way to include the Variables field as part of the deserialization process so that it would also work for e.g. `Sequence` activities?

bobhauser:
Turns out that Variables were serialized properly for `Sequence` and `Parallel` containers since they are able to just use regular activity serialization. The issue with `Flowchart` is that it has it's own Converter that missed a number of properties (`Variables`, `Name`, `Type`, `Version).

I also noticed that `Activity.CustomProperties` and `Activity.Metadata` deserialized their values to `JsonElement` objects. I found that specifying to use `PolymorphicObjectConverterFactory` for those properties resolved this (after a small change to it's `CanConvert`).

Unit tests were added to ensure properties were equivalent after serialization/deserialization.

sfmskywalker (approved):

sfmskywalker:
Ah, that makes sense, great catch! Thanks also for the tests 👍🏻

PR #5561, opened 2024-06-08
Add application roles and configure them in MassTransit

sfmskywalker:
A new enum ApplicationRole has been added for distinguishing among different roles (Hybrid, Api, Worker) an application can take. In the configuration of MassTransit, it is now possible to disable the consumers based on application role, which can help optimize the usage of resources and increase application efficiency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5561)

### DIFF
--- a/src/bundles/Elsa.Server.Web/Enums/ApplicationRole.cs
+++ b/src/bundles/Elsa.Server.Web/Enums/ApplicationRole.cs
@@ -1,0 +1,8 @@
+namespace Elsa.Server.Web;
+
+public enum ApplicationRole
+{
+    Hybrid,
+    Api,
+    Worker
+}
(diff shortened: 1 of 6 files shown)


src/bundles/Elsa.Server.Web/Enums/ApplicationRole.cs
```diff
@@ -0,0 +1,8 @@
+namespace Elsa.Server.Web;
+
+public enum ApplicationRole
+{
+    Hybrid,
+    Api,
+    Worker
```
raymonddenhaan, line 7: Is there no configuration needed for Workers? E.g. disabling the Http module?

raymonddenhaan (approved):

src/bundles/Elsa.Server.Web/Enums/ApplicationRole.cs
sfmskywalker: I considered that, but since the load balancer needs to be configured to not route HTTP requests to this role anyway, there's no need. This keeps it simple, but we can do it if necessary for some reason.

PR #5574, opened 2024-06-12
Fix issue where creation of workflow no longer returned a 201

raymonddenhaan:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5574)


MariusVuscanNx (approved):

sfmskywalker:
I changed this a short while ago to fix another bug, I believe in Elsa Studio. the problem with the 201 and URL pointing at the created object is that that URL responds with a different shape. I.e. depending on whether you create or update a workflow, you would get two different shapes as a response

PR #5617, opened 2024-06-16
Fixing some analyzer warnings

lahma:
Fixing some analyzer warnings reported by `<AnalysisLevel>latest-Recommended</AnalysisLevel>`. Less changes and maybe incrementally fix the issues.

* use concrete types for better performance
* static methods are better when possible
* invariant culture for parsing so no need to think about what's user's locale

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5617)

### DIFF
--- a/src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
+++ b/src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
@@ -6,8 +6,6 @@ namespace Elsa.Api.Client.Shared.Models;
 /// <typeparam name="T">The type of the items.</typeparam>
 public class PagedListResponse<T> : LinkedEntity
 {
-#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
-    public ICollection<T> Items { get; set; }
-#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+    public ICollection<T> Items { get; set; } = default!;
     public long TotalCount { get; set; }
 }
--- a/src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
+++ b/src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
@@ -3,11 +3,11 @@ namespace Elsa.Api.Client.Shared.Models;
 /// <summary>
 /// Represents a generic paged list response that offers a unified format for returning paged list of things from API endpoints.
 /// </summary>
-/// <param name="Items">A page of items.</param>
-/// <param name="TotalCount">The total number of items.</param>
 /// <typeparam name="T">The type of the items.</typeparam>
 public class PagedListResponse<T> : LinkedEntity
 {
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
     public ICollection<T> Items { get; set; }
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
     public long TotalCount { get; set; }
 }
--- a/src/common/Elsa.Mediator/Middleware/Command/CommandPipelineBuilder.cs
+++ b/src/common/Elsa.Mediator/Middleware/Command/CommandPipelineBuilder.cs
@@ -6,7 +6,7 @@ namespace Elsa.Mediator.Middleware.Command;
 public class CommandPipelineBuilder : ICommandPipelineBuilder
 {
     private const string ServicesKey = "mediator.Services";
-    private readonly IList<Func<CommandMiddlewareDelegate, CommandMiddlewareDelegate>> _components = new List<Func<CommandMiddlewareDelegate, CommandMiddlewareDelegate>>();
+    private readonly List<Func<CommandMiddlewareDelegate, CommandMiddlewareDelegate>> _components = new();
 
     /// <summary>
     /// Initializes a new instance of the <see cref="CommandPipelineBuilder"/> class.
@@ -38,8 +38,10 @@ public class CommandPipelineBuilder : ICommandPipelineBuilder
     {
         CommandMiddlewareDelegate pipeline = _ => new ValueTask();
 
-        foreach (var component in _components.Reverse())
-            pipeline = component(pipeline);
+        for (int i = _components.Count - 1; i >= 0; i--)
+        {
+            pipeline = _components[i](pipeline);
+        }
 
         return pipeline;
     }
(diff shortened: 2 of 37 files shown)


src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
```diff
@@ -3,11 +3,11 @@ namespace Elsa.Api.Client.Shared.Models;
 /// <summary>
 /// Represents a generic paged list response that offers a unified format for returning paged list of things from API endpoints.
 /// </summary>
-/// <param name="Items">A page of items.</param>
-/// <param name="TotalCount">The total number of items.</param>
 /// <typeparam name="T">The type of the items.</typeparam>
 public class PagedListResponse<T> : LinkedEntity
 {
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
sfmskywalker, line 9: Out of curiosity, what about:

```csharp
public ICollection<T> Items { get; set; } = default!;
```

To get rid of the warning?

src/common/Elsa.Mediator/Middleware/Command/CommandPipelineBuilder.cs
```diff
@@ -38,8 +38,10 @@ public CommandMiddlewareDelegate Build()
     {
         CommandMiddlewareDelegate pipeline = _ => new ValueTask();
 
-        foreach (var component in _components.Reverse())
-            pipeline = component(pipeline);
+        for (int i = _components.Count - 1; i >= 0; i--)
+        {
+            pipeline = _components[i](pipeline);
+        }
```
sfmskywalker, line 44: This changes seems both more complex and less performant (theoretically speaking), given that now we're using am indexer into the components list during each iteration. What's the upshot to this change that I don't see?

sfmskywalker (commented):
This is great, thanks @lahma! ❤️ 
I had two questions that piqued my curiosity.

src/clients/Elsa.Api.Client/Shared/Models/PagedListResponse.cs
lahma: I changed from ugly #pragma to your suggested approach.

src/common/Elsa.Mediator/Middleware/Command/CommandPipelineBuilder.cs
lahma: Well this is purely because `.Reverse()` now maps to 'List<T>.Reverse()' which actually reverses the list data and is a `void` method 🙂

lahma:
As the old saying goes: trust, but verify. Here's example benchmark I created:

```csharp
[MemoryDiagnoser]
public class ReverseBenchmark
{
    private List<string> _list;
    private IList<string> _interfaceList;

    [GlobalSetup]
    public void GlobalSetup()
    {
        _list = Enumerable.Range(0, N).Select(x => x.ToString()).ToList();
        _interfaceList = Enumerable.Range(0, N).Select(x => x.ToString()).ToList();
    }

    [Params(5, 10, 50, 100)]
    public int N { get; set; }

    [Benchmark]
    public int Reverse_Interface()
    {
        var counter = 0;
        foreach (var item in _interfaceList.Reverse())
        {
            counter++;
        }

        return counter;
    }

    [Benchmark]
    public int Reverse_For()
    {
        var counter = 0;
        for (var i = _list.Count - 1; i >= 0; i--)
        {
            counter++;
        }

        return counter;
    }
}
```

And the results:

```

BenchmarkDotNet v0.13.12, Windows 11 (10.0.26120.961)
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK 8.0.302
  [Host]     : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX2


```
| Method            | N   | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------ |---- |-----------:|----------:|----------:|-------:|----------:|
| **Reverse_Interface** | **5**   |  **36.209 ns** | **0.4178 ns** | **0.3704 ns** | **0.0067** |     **112 B** |
| Reverse_For       | 5   |   1.067 ns | 0.0006 ns | 0.0005 ns |      - |         - |
| **Reverse_Interface** | **10**  |  **54.144 ns** | **0.5943 ns** | **0.5268 ns** | **0.0091** |     **152 B** |
| Reverse_For       | 10  |   2.357 ns | 0.0534 ns | 0.0615 ns |      - |         - |
| **Reverse_Interface** | **50**  | **157.237 ns** | **1.2980 ns** | **1.2142 ns** | **0.0281** |     **472 B** |
| Reverse_For       | 50  |  13.741 ns | 0.1100 ns | 0.1029 ns |      - |         - |
| **Reverse_Interface** | **100** | **284.629 ns** | **5.0112 ns** | **4.6875 ns** | **0.0520** |     **872 B** |
| Reverse_For       | 100 |  31.936 ns | 0.2053 ns | 0.1921 ns |      - |         - |

sfmskywalker:
Ah, but there’s another saying: “Apples to apples”. Although the unit tests show a clear performance benefit to using `for` over `foreach` on an enumerable from Reverse, it doesn’t take into account the indexing into a List, which wouldn’t be necessary when using the foreach approach. Surely this must add to the nanoseconds?

lahma:
I'm not sure I follow. Indexing is always faster compared to taking `foreach` against a List, `foreach` takes an enumerator out of List to fulfill the contract. (a struct enumerator which wont box when done against concrete `List` instead of `IList`)  `foreach` against a plain array  (`string[]`) is special cased an turns into integer indexing without bounds checks - but this does not apply to `List<T>.`  

If you have a code example in mind I'd gladly try some performance comparisons 🙂

lahma:
So basically what I would like to bring to the table is that any LINQ method can be complex, here's the logic for `Reverse` in question: https://github.com/dotnet/runtime/blob/06e00763dfff79f2b3ca532d3b0c4003986f4107/src/libraries/System.Linq/src/System/Linq/Reverse.cs , `for` loop is verbose, but skips all the state handling etc.

sfmskywalker:
Sorry for the confusion earlier. Let me clarify with an example.

Initially, I assumed that the following code would perform worse due to indexing into the list:

```csharp
for (var i = 0; i < myList.Count; i++)
{
    var someValue = myList[i]; // <-- indexing into a list or array.
}
```

I believed that iterating over an enumerable would be faster because I thought it used an iterator to maintain the current value:

```csharp
foreach (var value in myList)
{
    var someValue = value; // <-- value presumably takes the "current item" from the iterator on myList.
}
```

However, I now understand that this may not be the same as iterating over an iterator from `Reverse`, and the `Reverse.cs` code you referenced indeed performs various operations, including indexing into a buffer (an array).

As the old saying goes: To assume is to make an **ass** out of **u** and **me** 😄

Thanks for teaching me how things _actually_ work 👍🏻

lahma:
Jint is on the other extreme where LINQ is "banned", so much interpretation and hot paths so anything costing extra is bad. LINQ in wonderful for regular line of business code which are not called in tight loops etc. In this case I went for `for` loop because it solved the reverse problem and of course while being verbose, can be much faster.

sfmskywalker:
Makes total sense 🙏

Issue #5391, opened 2024-05-16
[BUG] Elsa.UseMongoDb() options.DatabaseName does not set Database


Nekrosis1:
## Description
When registering MongoDB for ELSA, setting the options.DatabaseName does nothing.
Instead, ELSA relies on the "defaultauthdb", the database defined in the connection string after the /.
https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/connection/connect/

## Steps to Reproduce

1. **Detailed Steps**:
- Use Elsa with a mongoDB.
- removing the defaultauthdb piece results in ELSA unable to connect to the db.
- Setting a Database when registering the service does not change anything.

2. **Code Snippets**: 
```
builder.Services.AddElsa(elsa =>
{
   elsa.UseMongoDb(MongoDB.ConnectionString, options =>
   {
     options.DatabaseName = "myDatabase";
   });
 }
 ```

## Expected Behavior
The Database set in the service registration gets used by ELSA

## Actual Behavior
The Database set in the connection string as default auth DB gets used by ELSA.
If none is set (piece is optional), Elsa fails on startup.

## Environment
- **Elsa Package Version**: 3.1.3
- **Operating System**: Win10.
- **Storage**: MongoDB

## Log Output
In program.cs, on 
`app.UseWorkflowsApi();`
System.ArgumentNullException: 'Value cannot be null. (Parameter 'databaseName')'

## Troubleshooting Attempts
Setting the desired Database as the default auth DB is a workaround, as long as you do not use an auth DB that differs from the Elsa DB.

PR #5634, opened 2024-06-19
Use database name from options when specified

4lexKislitsyn:
MongoDb client allows to specify authentication database in connection string but use different databases to store data. `DatabaseName` was not used in `MongoDbFeature` to configure storage in database other than the one specified in the connection string.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5634)


4lexKislitsyn:
It will resolve #5391 issue

raymonddenhaan (approved):

PR #5636, opened 2024-06-19
Implement automated cleanup for Azure Service Bus subscriptions

raymonddenhaan:
### Reverted naming conventions of Azure Service Bus temporary queues and subscriptions.

This fixes being able to remove temporary queues and subscriptions from the service bus namespace based on missing heartbeats.

### Added a new service that is able to clean up the entire service bus namespace

This service will use all topics which name start with `elsa` to clean up all
- Topics without any subscriptions
- Subscriptions for which the forwarding queue can not be found in the same namespace.

#fixes 5635

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5636)

### DIFF
--- a/src/modules/Elsa.MassTransit.AzureServiceBus/HostedServices/CleanSubscriptionsWithoutQueues.cs
+++ b/src/modules/Elsa.MassTransit.AzureServiceBus/HostedServices/CleanSubscriptionsWithoutQueues.cs
@@ -1,4 +1,4 @@
-using Elsa.MassTransit.AzureServiceBus.Notifications;
+using Elsa.MassTransit.AzureServiceBus.Commands;
 using Elsa.MassTransit.AzureServiceBus.Options;
 using Elsa.Mediator.Contracts;
 using Medallion.Threading;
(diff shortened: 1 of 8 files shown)


src/modules/Elsa.MassTransit.AzureServiceBus/HostedServices/CleanSubscriptionsWithoutQueues.cs
```diff
@@ -0,0 +1,57 @@
...
+        using var scope = serviceProvider.CreateScope();
+        var lockProvider = scope.ServiceProvider.GetRequiredService<IDistributedLockProvider>();
+        var notificationSender = scope.ServiceProvider.GetRequiredService<INotificationSender>();
+        
+        const string lockKey = "SubscriptionCleanupService";
+        await using var monitorLock = await lockProvider.TryAcquireLockAsync(lockKey, TimeSpan.Zero);
+        if (monitorLock == null)
+            return;
+        
+        await notificationSender.SendAsync(new CleanupSubscriptions());
```
sfmskywalker, line 55: This looks like a candidate for sending a command rather than a notification.

src/modules/Elsa.MassTransit.AzureServiceBus/Notifications/CleanupSubscriptions.cs
```diff
@@ -0,0 +1,8 @@
+using Elsa.Mediator.Contracts;
+
+namespace Elsa.MassTransit.AzureServiceBus.Notifications;
+
+/// <summary>
+/// Notification to clean up the Azure Service Bus subscriptions.
+/// </summary>
+public record CleanupSubscriptions : INotification;
```
sfmskywalker, line 8: This sounds more like a command than an event. Unless there are multiple handlers interested in this, it might make more sense to turn this into a command.

sfmskywalker (approved):

Issue #5499, opened 2024-06-03
[BUG] Cron Expression causes workflow to fire twice in a row


MaxBrooks114:
## Description
I have a workflow that is triggered by a cron expression to run once a day. For some reason it causes the workflow to be triggered twice immediately. The workflow then never completes and stays in the suspended state even though all the activities have been completed successfully. 
## Steps to Reproduce

1. **Detailed Steps**: 
     var cronTrigger = new Cron
     {
         CronExpression = new("0 30 16 * * ?"),
         CanStartWorkflow = true
     };

4. **Reproduction Rate**: 
Happens every time. 

## Expected Behavior
Workflow triggers once and completes.

## Actual Behavior
workflow triggers twice and enters suspended state.

## Screenshots
![image](https://github.com/elsa-workflows/elsa-core/assets/45081361/2ef40f19-f781-4315-ba4e-6fd54d6f5330)

## Environment
Elsa 3.1.3, .net 8 on windows

## Additional Context
it is a flowchart workflow.

MaxBrooks114:
The suspended state is unrelated. However the duplicate triggering still happens, unknown why.

sfmskywalker:
@MaxBrooks114 @yutao815  Can you try with the latest 3.2.0-preview packages to see if the issue still exists?

sfmskywalker:
I just tried to reproduce this issue, but it works without issue. I will close this as unreproducible, but if you can reproduce the issue with the latest preview packages in a sample project, then please feel free to attach a zip file and I'll check it out. Thank you.

MaxBrooks114:
@sfmskywalker I can confirm this bug still happens for me on preview version 3.2.0-sonar-cloud.1502. I will try to reproduce with a sample project when i have time

sfmskywalker:
3.2.0-sonar-cloud.1502 Is quite an old build. You might want to try 3.2.0-preview.xxx.

MaxBrooks114:
@sfmskywalker same issue with 3.2.0-preview.1717 too.

sfmskywalker:
Ok, thank you for checking.  I look forward to the repro.

MaxBrooks114:
[ElsaCronBugReplication.zip](https://github.com/user-attachments/files/15778068/ElsaCronBugReplication.zip)


[...]

PR #5687, opened 2024-07-02
Register scheduler as singleton to avoid multiple starts of same task

4lexKislitsyn:
Scoped scheduler can't detect start duplicates because it's called from different scopes and doesn't have shared registry for tasks. (#5499)

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5687)


sfmskywalker (approved):

sfmskywalker:
Thanks @4lexKislitsyn !

PR #5713, opened 2024-07-05
Workflow refresh feature

MariusVuscanNx:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5713)


sfmskywalker (approved):

src/modules/Elsa.Workflows.Runtime/Services/WorkflowDefinitionRefresher.cs
```diff
@@ -0,0 +1,55 @@
...
+        var processedWorkflowDefinitions = new List<WorkflowDefinition>();
+        var batchSize = request.BatchSize;
+        var order = new WorkflowDefinitionOrder<string>(x => x.Id, OrderDirection.Ascending);
+
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            var pageArgs = PageArgs.FromPage(currentPage, batchSize);
+            var definitions = await store.FindManyAsync(filter, order, pageArgs, cancellationToken);
+
+            if (definitions.Items.Count == 0)
```
raymonddenhaan, line 36: We can also add a check for count < batchSize. That will prevent an extra call to the database.

test/component/Elsa.Workflows.ComponentTests/Scenarios/WorkflowDefinitionRefresh/DynamicEndpointTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        StaticValueHolder.Value = "second-value";
+        await _workflowDefinitionsRefresher.RefreshWorkflowDefinitionsAsync(new Runtime.Requests.RefreshWorkflowDefinitionsRequest
+        {
+            BatchSize = 10,
+            DefinitionIds = ["f69f061159adc3ae"]
+        }, CancellationToken.None);
+
+        var secondResponse = await client.GetStringAsync("second-value");
+
+        Assert.Equal("", firstResponse);
```
raymonddenhaan, line 33: Why not check against the response code?

src/modules/Elsa.MassTransit/Consumers/WorkflowDefinitionEventsConsumer.cs
```diff
@@ -1,27 +1,28 @@
 using Elsa.MassTransit.Messages;
+using Elsa.MassTransit.Services;
+using Elsa.Mediator.Contracts;
 using Elsa.Workflows.Management.Contracts;
 using JetBrains.Annotations;
 using MassTransit;
 
 namespace Elsa.MassTransit.Consumers;
 
-/// <summary>
```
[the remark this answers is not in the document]
sfmskywalker: Generally speaking, the less noise on the screen, the better. Removing these optional tags help with that.

src/modules/Elsa.Workflows.Runtime/Services/WorkflowDefinitionRefresher.cs
sfmskywalker: Yes, nice. @MariusVuscanNx can we incorporate that in your next PR?

test/component/Elsa.Workflows.ComponentTests/Scenarios/WorkflowDefinitionRefresh/DynamicEndpointTests.cs
raymonddenhaan: The test is checking if the new endpoint is set up within the triggers. If this is not the case I believe a 404 should be returned, not a 200.

sfmskywalker: Ah, yes you are right. @MariusVuscanNx what do you think?

MariusVuscanNx: The method perform an http GET request and tries to retrieve the content (string). If case it doesn't work a 404 will be returned and an exception will be thrown. 

In short it will work only if we have a successful status code.

sfmskywalker: That makes sense. But then, I wonder, if it wouldn't be more reflective of the test to do e.g. `client.SendAsync()`, receive the `HttpResponseMessage` and then explicitly check for its `StatusCode` to be 200.

MariusVuscanNx: I updated the code to check the status code.

PR #5781, opened 2024-07-16
Fix reloading logic for workflow definitions

sfmskywalker:
Fixed cache invalidation in response to reloading workflow definitions. This includes creating a "ReloadedWorkflowDefinition" model, updating the caching logic, and modifying notification handlers to work with the enhanced workflow reloading logic. This ensures workflow definitions are updated and managed correctly when published, retracted, or deleted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5781)


src/modules/Elsa.Workflows.Runtime/Handlers/RefreshActivityRegistry.cs
```diff
@@ -0,0 +1,33 @@
...
+using Elsa.Workflows.Management.Activities.WorkflowDefinitionActivity;
+using Elsa.Workflows.Management.Contracts;
+using Elsa.Workflows.Management.Entities;
+using Elsa.Workflows.Runtime.Notifications;
+using JetBrains.Annotations;
+
+namespace Elsa.Workflows.Runtime.Handlers;
+
+/// <summary>
+/// Refreshes the <see cref="IActivityRegistry"/> for the <see cref="WorkflowDefinitionActivityProvider"/> provider whenever an <see cref="WorkflowDefinition"/> is published, retracted or deleted.
```
raymonddenhaan, line 12: This will only happen on a reload, not on any publish, retract of deletion.

PR #5828, opened 2024-07-24
Drop NET 7 support and cleanup NuGet package references

lahma:
* NET 7 is no longer supported
* `Directory.Packages.props` should mandate used versions
* Add NuGet audit mode setting to show any problems during restore
* `System.Text.Json` should be `8.0.4 `which is patched against security vulnerabilities

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5828)

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Begin SonarCloud analysis
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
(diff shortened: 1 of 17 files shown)


sfmskywalker (approved):
Many thanks @lahma 🔥

lahma:
I had to make small tweaks still. Hopefully getting a green build soon 😅

lahma:
OK now should report problems during restore and build is green. If I alter package references in projects (remove explicit) I start to get warnings about vulnerable packages.

sfmskywalker:
The PR build step is all green - which is good enough - the SonarCloud Code Analysis step tends to take a long time to report back for some reason, but this is not of your doing. If you're happy with the PR then I am, too.

PR #5860, opened 2024-08-06
Add engine exception handling middleware

raymonddenhaan:
Integrate new EngineExceptionHandlingMiddleware into the workflow execution pipeline to catch and log exceptions outside of running workflow activity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5860)

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
+++ b/src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
@@ -529,7 +529,8 @@ public partial class WorkflowExecutionContext : IExecutionContext
             WorkflowSubStatus.Suspended => WorkflowStatus.Running,
             _ => throw new ArgumentOutOfRangeException(nameof(subStatus), subStatus, null)
         };
-
+    
+    // TODO: Check if we should not use the target subStatus here instead.
     private bool ValidateStatusTransition()
     {
         var currentMainStatus = GetMainStatus(SubStatus);
--- a/src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
+++ b/src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
@@ -16,18 +16,14 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace Elsa.Workflows;
 
-/// <summary>
 /// A delegate entry that is used by activities to be notified when the activities they scheduled are completed.
-/// </summary>
 /// <param name="Owner">The activity scheduling the <see cref="Child"/> activity.</param>
 /// <param name="Child">The child <see cref="IActivity"/> being scheduled.</param>
 /// <param name="CompletionCallback">The <see cref="ActivityCompletionCallback"/> delegate to invoke when the scheduled <see cref="Child"/> activity completes.</param>
 /// <param name="Tag">An optional tag.</param>
 public record ActivityCompletionCallbackEntry(ActivityExecutionContext Owner, ActivityNode Child, ActivityCompletionCallback? CompletionCallback, object? Tag = default);
 
-/// <summary>
 /// Provides context to the currently executing workflow.
-/// </summary>
 [PublicAPI]
 public partial class WorkflowExecutionContext : IExecutionContext
 {
@@ -40,9 +36,7 @@ public partial class WorkflowExecutionContext : IExecutionContext
     private readonly IHasher _hasher;
     private readonly Action<WorkflowExecutionContext>? _statusUpdatedCallback;
 
-    /// <summary>
     /// Initializes a new instance of <see cref="WorkflowExecutionContext"/>.
-    /// </summary>
     private WorkflowExecutionContext(
         IServiceProvider serviceProvider,
         WorkflowGraph workflowGraph,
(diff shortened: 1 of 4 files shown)


src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
```diff
@@ -670,7 +530,7 @@ subStatus switch
             _ => throw new ArgumentOutOfRangeException(nameof(subStatus), subStatus, null)
         };
 
-    private bool ValidateStatusTransition(WorkflowSubStatus targetSubStatus)
+    private bool ValidateStatusTransition()
     {
         var currentMainStatus = GetMainStatus(SubStatus);
```
sfmskywalker, line 535: I think this is the wrong fix and instead we should change this line, to:

```csharp
if (!ValidateStatusTransition(subStatus))
```

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
```diff
@@ -543,16 +419,14 @@ public T UpdateProperty<T>(string key, Func<T?, T> updater)
...
-    /// </summary>
     public bool HasProperty(string name) => Properties.ContainsKey(name);
 
-    internal bool CanTransitionTo(WorkflowSubStatus targetSubStatus) => ValidateStatusTransition(targetSubStatus);
+    internal bool CanTransitionTo(WorkflowSubStatus targetSubStatus) => ValidateStatusTransition();
 
     internal void TransitionTo(WorkflowSubStatus subStatus)
     {
-        if (!ValidateStatusTransition(SubStatus))
+        if (!ValidateStatusTransition())
```
sfmskywalker, line 429:
```suggestion
        if (!ValidateStatusTransition(subStatus))
```

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
```diff
@@ -670,7 +530,7 @@ subStatus switch
             _ => throw new ArgumentOutOfRangeException(nameof(subStatus), subStatus, null)
         };
 
-    private bool ValidateStatusTransition(WorkflowSubStatus targetSubStatus)
+    private bool ValidateStatusTransition()
```
sfmskywalker, line 533:
```suggestion
    private bool ValidateStatusTransition(WorkflowSubStatus subStatus)
```

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
sfmskywalker, line 535:
```suggestion
        var currentMainStatus = GetMainStatus(subStatus);
```

src/modules/Elsa.Workflows.Core/Middleware/Workflows/EngineExceptionHandlingMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+using Elsa.Workflows.Models;
+using Elsa.Workflows.Pipelines.WorkflowExecution;
+using Elsa.Workflows.State;
+using Microsoft.Extensions.Logging;
+
+namespace Elsa.Workflows.Middleware.Workflows;
+
+/// <summary>
+/// Adds extension methods to <see cref="ExceptionHandlingMiddleware"/>.
+/// </summary>
```
sfmskywalker, line 12: Not a big deal (at all), but it gave me a giggle to see the effort that was made in the `WorkflowExecutionContext` where all the `///<summary>` lines were removed, but are added here 😄

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
raymonddenhaan: I did not make a fix here, just removed code that was not used. The validate method is not so much checking if the new state can be moved to as it is checking if the current state is finished and therefor a state that can be moved from. Sending the target sub status and using it instead of the current `SubStatus` will prevent us from moving to the Finished state.

sfmskywalker: I understand. I'm just wondering if the original code is a mistake and whether it should have been using the parameter instead of the property. I'll add a TODO to follow up.

src/modules/Elsa.Workflows.Core/Contexts/WorkflowExecutionContext.cs
sfmskywalker:
```suggestion
// TODO: Check if this method should be using `targetSubStatus` instead of `SubStatus`.
private bool ValidateStatusTransition(WorkflowSubStatus targetSubStatus)
    private bool ValidateStatusTransition()
    {
        var currentMainStatus = GetMainStatus(SubStatus);
```

Suchiman:
Why was the `<summary>` removed? without that, VS does not display documentation anymore

sfmskywalker:
To reduce the amount of noise in a file. However, we weren’t aware that this would break VS.

Suchiman:
You can change it from
```csharp
/// <summary>
/// description
/// </summary>
```
to
```csharp
/// <summary>description</summary>
```
if you desire a more compact format, without breaking VS.

PR #5882, opened 2024-08-10
Fix background activity completion

sfmskywalker:
Fixed background activity completion by fixing a bug in the bookmark updater.
The bookmark updater was deleting too many bookmarks, preventing background activities from being resumed & completed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5882)

### DIFF
--- a/samples/aspnet/Elsa.Samples.AspNet.OrchardCoreIntegration/appsettings.json
+++ b/samples/aspnet/Elsa.Samples.AspNet.OrchardCoreIntegration/appsettings.json
@@ -2,7 +2,8 @@
   "Logging": {
     "LogLevel": {
       "Default": "Information",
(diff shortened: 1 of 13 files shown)


JulienAmbos:
@sfmskywalker Thank you for merging this fix. But is this already released? I think it did not make it into 3.2.0?
This might fix my issue with mass transit scheduled background activities not being completed, when inside a `Parallel` as the last activity (except `Finish()`) in a workflow.

sfmskywalker:
Hi @JulienAmbos , this merge commit went into the `main` branch, which means it will be part of 3.3.0 (and is currently part of the 3.3.0-preview builds available from the feedz.io feed. If possible, you can try it out against your project, but keep in mind that the 3.3 previews can be volatile until it’s released later this year.

PR #5899, opened 2024-08-14
[FIX] cannot end a flowchart when a End is Then Activity in If

### LINKED ISSUE #5823: [BUG]Cannot break a Flowchart in While
## Description

When a While’s Body is a flowchart , and there is a Break as a Then activity in a If, I found that I cannot break the While.
```
While w = new() {
    Body = new Flowchart() {
        Activities = {
         new WriteLine(“Do1”),
            new If(e => true) {
            Then = new Break(),
            },
        new WriteLine(“Do2”),
        }
       // Connected like Sequence
    }
};
```

**Also**, without the while , I cannot terminate the flowchart when the Then activity is End.
```
Flowchart chart = new() {
    Activities = {
        new WriteLine(“Do1”),
        new If(e => true) {
            Then = new End(),
        },
        new WriteLine(“Do2”)
    }
    // Connected like Sequence
}
```

## Expected Behavior
- For first example, it should output **“Do1”** as expected. 
- For second example, it should output **“Do1”** as expected. 

## Actual Behavior
- For first example, it outputted **”Do1” “Do2”** in actual. 
- For second example, it outputted **“Do1” “Do2”** in actual. 

## Environment
- **Elsa Package Version**: 3.1.3.
- **Operating System**: Windows 10
- **Browser and Version**: No browser.

## Troubleshooting Attempts
It seems `Flowchart` doesn’t receive `BreakSignal` and judge whether `IsBreak` in `OnChildCompletedAsync` method of `Flowchart`.
I have tried to substitute `Flowchart` with `Sequence` and it goes well.

iio888:
fix #5823 
Welcome to post suggestions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5899)


iio888:
@dotnet-policy-service agree

sfmskywalker (commented):
It makes sense. Only thing I'm unsure of is whether we should reuse the `BreakSignal`, or if it would be better to create a dedicated `EndSignal`. The former is intended to break out of a "scope" such as a for-each activity's Body, while the latter would be intended to complete the closest flowchart activity containing the End activity.

Thoughts?

iio888:
Good idea! The `EndSignal` is better than `BreakSignal` exactly.
I’ll modify my code. Thanks for your advice

iio888:
Now the `End` Activity will ends all the parent flowcharts. Is it will be better that just ends the last flowchart and stop propagation? @sfmskywalker

src/modules/Elsa.Workflows.Core/Activities/Break.cs
```diff
@@ -24,5 +24,6 @@ protected override async ValueTask ExecuteAsync(ActivityExecutionContext context
     {
         // Send a signal to the parent scope to break out of the loop.
         await context.SendSignalAsync(new BreakSignal());
+        await context.SendSignalAsync(new EndSignal());
```
sfmskywalker, line 27: What might be the reason for the Break activity to send an End signal?

iio888: It is set to be used to send an End signal to Flowchart when is contained in a while.
It doesn’t look like a perfect solution indeed.

iio888:
@sfmskywalker Hi. Any thoughts about this or?

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.cs
```diff
@@ -27,6 +27,9 @@ public Flowchart([CallerFilePath] string? source = default, [CallerLineNumber] i
         OnSignalReceived<ScheduleActivityOutcomes>(OnScheduleOutcomesAsync);
         OnSignalReceived<ScheduleChildActivity>(OnScheduleChildActivityAsync);
         OnSignalReceived<CancelSignal>(OnActivityCanceledAsync);
+
+        OnSignalReceived<BreakSignal>(OnBreakSignalReceived);
```
sfmskywalker, line 31: Why does the Flowchart activity need to handle the Break signal?

bobhauser:
I am concerned about this PR and how it would adversely affect existing workflows. The way that I expect End activities to work is for them to end the current flow. If there is a single flow in the process, then it would end the process. But if there are concurrent flows, only the flow that reached an End should stop - the other concurrent flows should continue until they reach their respective End activities (or complete all activities if no explicit End).

This would be a very breaking change for many workflows in our organization. And it would go against how End activities are treated in BPMN.

Perhaps there should be a new TerminalEnd (or ProcessEnd) activity or something if this new behaviour is required.

sfmskywalker:
@bobhauser Thanks for chiming in. You're right, actually. The description of the End activity states as much: it's a milestone event, and an explicit way to mark the end of a sequence.

@iio888 what do you think of the suggestion to introduce a new activity called Terminate? Its effect should be equivalent to that of the BPMN's Terminate End, which (I think) is essentially the same behavior as the one you implemented for the End activity in this PR.

iio888:
Maybe there is something mistakes in my code. What I want to fix is equivalent to **None End** actually. As in #5823, I found some potential problems when I am using `End` and `If` and I have no idea to change the original functionality of `End`. 
Thanks for your suggestions.

PR #5902, opened 2024-08-14
Add JS functions for encoding/decoding byte arrays from and to strings

sfmskywalker:
This PR introduces JavaScript functions to handle encoding and decoding of byte arrays to and from strings. The encoding converts a byte array into a string, and the decoding function does the reverse, converting a string back into a byte array. These utility functions will be useful for scenarios where byte arrays need to be represented as strings for storage or transmission and then restored to their original byte form.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5902)


src/modules/Elsa.JavaScript/TypeDefinitions/Providers/CommonFunctionsDefinitionProvider.cs
```diff
@@ -91,6 +91,26 @@ protected override IEnumerable<FunctionDefinition> GetFunctionDefinitions(TypeDe
...
+        
+        yield return CreateFunctionDefinition(builder => builder
+            .Name("bytesToBase64")
+            .Parameter("value", "Byte[]")
+            .ReturnType("string"));
+        
+        yield return CreateFunctionDefinition(builder => builder
+            .Name("bytesFromBase64")
+            .Parameter("value", "string")
+            .ReturnType("Byte[]"));
```
raymonddenhaan, line 114: Should we add methods for converting between Base64 and string directly?

raymonddenhaan (approved):

src/modules/Elsa.JavaScript/TypeDefinitions/Providers/CommonFunctionsDefinitionProvider.cs
sfmskywalker: Why not indeed, might be convenient 👍🏻

PR #5909, opened 2024-08-16
Added support to retrieve the workflow details through Liquid script

cvijayak:
Added support for retrieving the following workflow details through Liquid script:
1. WorkflowDefinitionId
2. WorkflowDefinitionVersionId
3. WorkflowDefinitionVersion
4. WorkflowInstanceId

Currently, JavaScript supports all of the above, but it seems they were missing in Liquid script.


sfmskywalker (approved):
Thanks @cvijayak !

PR #5944, opened 2024-09-05
Enh/5943 DropIns - Prevention of File Locks and Unloading

### LINKED ISSUE #5943: [ENH] DropIns - Prevention of File Locks and Unloading
# Enhancement Request

## Context
At the present time there are two ways to insert drop-ins onto Elsa.

1. Via NuGet Packages 
   - Unzips the package and loads each DLL onto memory
   - Doesn't lock the file since it's closed after copying the DLLs onto memory
   - When the original `.npkg` is deleted the associated drop-ins are not unregistered from the registry

2. Via DLLs
    - Loads directly the Assembly from the DLL
    - Locks the file since the Assembly is pointing to the DLL file

## Planned Implementation
1. Change Drop-in Directory Monitor to allow tracking of the installed Drop-ins
    - Every time a drop-in is added, it should be added to a dictionary having the key as the path and the value the drop-ins associated to that file
2. Prevent the File locking of DLLs
    - Implement something similar to the NuGet's behavior (copy the DLL to a memory stream and load it from there). This will prevent the file locking.
      - If needed, this can possibly be improved on the future to copy the DLLs to an internal folder to not consume unbearable amounts of ram
3. Change Drop-ins to allow a safe unregister
    - Add a Unconfigure method to the IDropIn interface.
      - This method's implementation should unregister the activity from the registry as well as possible added services to the service provider.
4. Change Drop-ins Directory Monitor to keep track of deletion events
    - Add a callback function for the watcher.Deleted
      - This function will retrieve all drop-ins associated with the path that was deleted and call their Unconfigure method

RenatoCapelo:
#### PR Classification
New feature and code improvement.

#### PR Summary
Introduces an `Unconfigure` method for drop-ins and improves assembly loading to avoid file locking issues.
- `DropIn.cs` and `IDropIn.cs`: Added `Unconfigure` method for drop-ins.
- `NuGetPackageAssemblyLoadContext.cs`: Disposed `packageReader` to close `.nupkg` file.
- `AssemblyLoader.cs`: Modified to load assemblies from a memory stream.
- `DropInDirectoryMonitorHostedService.cs`: Added debounced unloader, async loading, and handling of file deletion events.
- `DropInDescriptor.cs`: Changed to class with `DynamicallyAccessedMembers` attribute to suppress IL2072 warning.

Fixes #5943

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5944)

### DIFF
--- a/samples/drop-ins/SampleDropIn/DropIn.cs
+++ b/samples/drop-ins/SampleDropIn/DropIn.cs
@@ -1,9 +1,11 @@
 using Elsa.DropIns.Core;
 using Elsa.Extensions;
 using Elsa.Features.Services;
+using Elsa.Workflows;
 using Elsa.Workflows.Contracts;
 using JetBrains.Annotations;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using SampleDropIn.Activities;
 
 namespace SampleDropIn;
@@ -20,5 +22,19 @@ public class DropIn : IDropIn
     {
         var activityRegistry = serviceProvider.GetRequiredService<IActivityRegistry>();
         await activityRegistry.RegisterAsync<SampleActivity>(cancellationToken: cancellationToken);
+    }
+
+    public void Unconfigure(IServiceProvider serviceProvider)
+    {
+        var logger = serviceProvider.GetRequiredService<ILogger<DropIn>>();
+        try
+        {
+            var activityRegistry = serviceProvider.GetRequiredService<IActivityRegistry>();
+            activityRegistry.Remove(typeof(ActivityRegistry), activityRegistry.Find<SampleActivity>()!);
+            logger.LogInformation("Drop-in unconfigured.");
+        }catch(Exception ex)
+        {
+            logger.LogError(ex, "Error unconfiguring drop-in.");
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


andrenevesgomes (approved):
Seems good to me. Great Job @RenatoCapelo! 👍

samples/drop-ins/SampleDropIn/DropIn.cs
```diff
@@ -21,4 +23,18 @@ public async ValueTask ConfigureAsync(IServiceProvider serviceProvider, Cancella
         var activityRegistry = serviceProvider.GetRequiredService<IActivityRegistry>();
         await activityRegistry.RegisterAsync<SampleActivity>(cancellationToken: cancellationToken);
     }
+
+    public void Unconfigure(IServiceProvider serviceProvider)
+    {
+        var logger = serviceProvider.GetRequiredService<ILogger<DropIn>>();
+        try
```
sfmskywalker, line 30: I'm wondering if it would be better for the calling code to do the try/catch stuff? That way, we don't have to repeat this exception handling in other drop-ins.

sfmskywalker (approved):
This looks great! I added one small request for your consideration.

samples/drop-ins/SampleDropIn/DropIn.cs
RenatoCapelo: I agree with you, I'm going to try and implement this on the calling code. 
I think I made it like this on the first place because the exception was not being caught by the calling code because of it being dynamically loaded, but I'm most likely wrong.
I'll come back with updates in a few minutes.

RenatoCapelo: Resolved on latest commit.

sfmskywalker (approved):
Awesome, thank you!

PR #5949, opened 2024-09-08
Upgrade Quartz and some warning fixes

lahma:
New Quartz version with a bit trimmed dependencies as start of removing some analyzer warnings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5949)


Directory.Build.props
```diff
@@ -28,9 +28,13 @@
         <IncludeSymbols>true</IncludeSymbols>
         <SymbolPackageFormat>snupkg</SymbolPackageFormat>
         <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
+
+        <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
lahma, line 32: This should ideally at some point be true (to give a TODO list)

Directory.Build.props
```diff
@@ -28,9 +28,13 @@
...
     </PropertyGroup>
     <PropertyGroup>
         <NoWarn>$(NoWarn);CS0162;CS1591</NoWarn>
+        <!-- IL trimming warnings -->
+        <NoWarn>$(NoWarn);IL2026;IL2046;IL2057;IL2067;IL2070;IL2072;IL2075;IL2087;IL2091</NoWarn>
```
lahma, line 37: This works as a TODO list for getting things trim-friendly

Directory.Packages.props
```diff
@@ -82,10 +82,8 @@
     <PackageVersion Include="Proto.Persistence.SqlServer" Version="1.6.0" />
     <PackageVersion Include="Proto.Remote" Version="1.6.0" />
     <PackageVersion Include="pythonnet" Version="3.0.3" />
-    <PackageVersion Include="Quartz" Version="3.11.0" />
-    <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.11.0" />
-    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.11.0" />
-    <PackageVersion Include="Quartz.Serialization.Json" Version="3.11.0" />
+    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.13.0" />
+    <PackageVersion Include="Quartz.Serialization.Json" Version="3.13.0" />
```
lahma, line 86: There's new SystemTextJson package for serialization which should preferred over the Newtonsoft based, I'm just unsure how much of a breaking change that would be

src/modules/Elsa.Quartz/Elsa.Quartz.csproj
```diff
@@ -9,8 +9,6 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Quartz" />
-        <PackageReference Include="Quartz.Extensions.DependencyInjection" />
         <PackageReference Include="Quartz.Extensions.Hosting" />
```
lahma, line 12: Relying on transitive behavior, Quartz 4 will eventually have these all built in o referencing just Quartz should be enough

lahma:
Tests might be also failing in `main` 🤷🏻‍♂️

Directory.Build.props
sfmskywalker: Yeah exactly. My approach is to try and fix warnings incrementally. Once that's reduced to a manageable size, we can flip this switch and fix the remaining ones and or ignore the ones we don't care about (if any).

Directory.Build.props
sfmskywalker: Nice

Directory.Packages.props
sfmskywalker: I'll do some testing to see if things break when switching. Thanks for pointing out.

sfmskywalker (approved):

sfmskywalker:
> Tests might be also failing in `main` 🤷🏻‍♂️

Yes, it's an issue in the main branch. Fixing it.

PR #5952, opened 2024-09-10
Upgrade to NUKE 8.1

lahma:
This also returns the build.* files back to root if it's OK, the workflow file is auto generated so it shouldn't be manually modified. It's also quite hard to find the idiomatic build scripts if they are hidden somewhere in the depths 🙂

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5952)


sfmskywalker (approved):
Sweet, thanks @lahma ! You're right about moving back the script, whatever is easier.

Issue #5984, opened 2024-09-24
[ENH] Enhance logging by adding logger scope as part of activity and workflow pipelines


bobhauser:
### Enhancement Overview
When viewing application logs, it is often useful to see contextual properties to help understand the logs. For instance, when viewing a log message it might be useful to know the workflow instance Id of the workflow being executed, or the activity type of the current activity. Adding this sort of context is usually accomplished through the use of the `ILogger.BeginScope` method. The desired format of the logger scope context is often very specific to the application. For instance, we use structured logging (serilog/datadog), and we might like to see the logging context nested like

```
workflow 
    / name
    / instance_id
    / activity
         / id
         / type
````
but of course everyone would not like it like that... so the logger scope generation should be injectable.

### Proposed Enhancement
I propose that new middleware be added to Elsa, both for the workflow pipeline and activity pipeline. The new middleware would make use of an injectable `ILoggerScopeGenerator` (name TBD) which would return the logger scope. The default implementation may just add `WorkflowInstanceId` / 'ActivityId` properties, where as a custom implementation could be injected if requried.

Note: `WorkflowInstanceId` logger scope is currently added directly by `WorkflowRunner.RunAsync` - this should be removed in preference to the proposed middleware.

### Alternative Solutions
Although I believe new pipeline middleware would be the cleanest approach, it would also be possible to have the code which triggers the pipeline execution set the log scope around the pipeline execution. As I see it, this would be in `WorkflowRunner.RunAsync` for workflow scope and `ActivityInvoker.InvokeAsync` for activity scope.

### Additional Context
We currently have this implemented using custom middleware and would be happy to contribute this to Elsa. Or, if the alternative solution is preferred, we'd be happy to do this work too.

bobhauser:
On further reflection, I think that the alternative solution mentioned above would be the best approach as it would apply even in situations where the default workflow and/or activity pipelines have been overwritten.

Also, a scope is already defined in `WorkflowRunner.RunAsync` - moving the scope creation into a pipeline middleware would mean that the scope would no longer be applied around notifications send using `notificationSender`.

PR #5985, opened 2024-09-25
Enhance logging by adding logger state to activity and workflow pipelines

bobhauser:
Added ability to set ILogger state when running workflows and invoking activities. This will make reviewing generated logs easier and provide the ability to correlate logs by "WorkflowInstanceId" (already in the product) and "ActivityInstanceId" (added in this PR). It is also possible to inject more specific ILoggerStateGenerators for when more specific context is required to match format of other application logs.

See #5984 for more details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5985)


src/modules/Elsa.Workflows.Core/Features/WorkflowsFeature.cs
```diff
@@ -128,9 +129,11 @@ private void AddElsaCore(IServiceCollection services)
             .AddScoped(typeof(Func<IWorkflowBuilder>), sp => () => sp.GetRequiredService<WorkflowBuilder>())
             .AddScoped<IWorkflowBuilderFactory, WorkflowBuilderFactory>()
             .AddScoped<IVariablePersistenceManager, VariablePersistenceManager>()
-            .AddScoped<IIncidentStrategyResolver, DefaultIncidentStrategyResolver>()
-
-            // Incident Strategies.
+            .AddScoped<IIncidentStrategyResolver, DefaultIncidentStrategyResolver>()
+            .AddSingleton<ILoggerStateGenerator<WorkflowExecutionContext>, WorkflowLoggerStateGenerator>()
```
bobhauser, line 133: @sfmskywalker Do you ever use TryAddSingleton? It seems like it could be useful here since otherwise the built-in implementation would override a custom implementation if already declared earlier. Or is it generally expected to add the override implementations after the .AddElsa()?

sfmskywalker: In Elsa, the typical pattern is to expose a delegate property from the feature class, which allows other modules or the application itself to register custom service implementations.

For example, consider the following structure in `FeatureA`:

```csharp
public class FeatureA : FeatureBase
{
    public Func<IServiceProvider, IMyService> MyServiceFactory { get; set; } = sp => new DefaultMyService();

    // This method is invoked by the feature framework after all features have had a chance to configure themselves and other features.
    public override void Apply()
    {
        Services.AddSingleton(MyServiceFactory);
    }
}
```

In the specific scenario you mentioned, this pattern isn't being followed, but it could certainly be introduced.

Alternatively, we could argue that it's likely an edge case for someone to want to provide their own implementation of `ILoggerStateGenerator<T>`. In this case, instead of using `AddSingleton`, we could switch to `TryAddSingleton`. However, this would require careful consideration of the order in which features are registered. The developer would need to ensure that their `TryAddSingleton` call occurs before the one from `WorkflowsFeature`. If that's the case, the developer could just as easily invoke the feature early and use `AddSingleton`. In such a scenario, multiple service descriptors might exist for the same service, but the last one to be registered would take precedence.

My take: the most flexible way is to use the delegate pattern that the Features framework supports.

sfmskywalker:
I really like the PR. It made me wonder if we could take it a step further by making it even more extensible, allowing multiple logger state generators to contribute to the logging state of a given type. For example:

```csharp
public class GeneralActivityLoggerStateGenerator : ILoggerStateGenerator<ActivityExecutionContext>
{
    public Dictionary<string, object> GenerateLoggerState(ActivityExecutionContext activityExecutionContext)
    {
        return new Dictionary<string, object>
        {
            ["ActivityInstanceId"] = activityExecutionContext.Id
        };
    }
}

// Custom implementation for a specific activity in my app:
public class MySpecialActivityLoggerStateGenerator : ILoggerStateGenerator<ActivityExecutionContext>
{
    public Dictionary<string, object> GenerateLoggerState(ActivityExecutionContext activityExecutionContext)
    {
        if (activityExecutionContext.Activity is not MySpecialActivity mySpecialActivity)
            return new Dictionary<string, object>();

        return new Dictionary<string, object>
        {
            ["MySpecialProperty"] = mySpecialActivity.MySpecialProperty
        };
    }
}
```

If you agree that such a scenario might be beneficial for others, it would inherently resolve the AddSingleton vs TryAddSingleton vs delegate-based registration question. In this case, we would always use `AddSingleton` to support multiple registrations. The `ActivityInvoker` and `WorkflowRunner` could then resolve all registered services of type `ILoggerStateGenerator<ActivityExecutionContext>` and `ILoggerStateGenerator<WorkflowExecutionContext>`, iterating through them to aggregate the logger states.

sfmskywalker (approved):

sfmskywalker:
@bobhauser Once the conflicts are resolved, we can merge this. Thanks for your contributions!

bobhauser:
@sfmskywalker I resolved the conflict. This branch seems to now be suffering the same build issue as main... pretty sure it is not related to my changes.

sfmskywalker:
Thanks. Yes definitely not of your doing. Will fix.

Issue #5583, opened 2024-06-12
[ENH] Enable External Interaction with Publish Event and Event Activities


sfmskywalker:
Currently, the PublishEvent and Event activities only operate internally within the workflow engine. Enabling external applications to leverage these primitives would significantly enhance the system's flexibility and integration capabilities.

### Proposed Enhancements

1. **PublishEvent Activity Enhancement**
   - **Current Behavior**: The PublishEvent activity handles events internally.
   - **Proposed Change**: Allow the PublishEvent activity to publish events that external applications can subscribe to. For example, an `EventPublished` event can be picked up by the Webhooks module, enabling external applications to handle these events and perform corresponding actions.

2. **Event Activity Enhancement**
   - **Current Behavior**: The Event activity is triggered by internal mechanisms.
   - **Proposed Change**: Enable external applications to trigger Event activities by sending HTTP requests or AMQP messages to the workflow engine. This will allow external systems to initiate events within the workflow, facilitating better integration and interaction.

### Implementation Guidance
- **Reference RunTask Activity**: The RunTask activity already has the capability to integrate with external applications. We can look at the RunTask activity to understand how this integration is implemented today and apply similar patterns to the PublishEvent and Event activities.

### Benefits
- **Enhanced Flexibility**: By allowing PublishEvent and Event activities to interact with external applications, the system's flexibility is greatly increased.
- **Improved Integration**: These enhancements will facilitate better integration with external systems, making it easier for external applications to interact with the workflow engine.
- **Expanded Use Cases**: Enabling external triggers and responses for PublishEvent and Event activities will open up new use cases and applications, enhancing the overall functionality of the workflow engine.

Implementing these enhancements will significantly improve the capabilities of the workflow engine, making it more robust and versatile in handling complex integrations with external systems.

cfgiugliano:
As discussed on Discord, it could be possible to add a boomarkless version of the RunTask:

```cs
[Activity("Elsa", "Primitives", "Requests a given domain event to be published. ", Kind = ActivityKind.Action)]
public class DomainEvent: Activity
{
    /// <summary>
    /// The name of the domain event being published.
    /// </summary>
    [Input(Description = "The name of the domain event being published.")]
    public Input<string> DomainEventName { get; set; } = default!;

    /// <summary>
    /// The payload of the domain event being requested.
    /// </summary>
    [Input(Description = "Any additional parameters to send to the domain event.")]
    public Input<object?> Payload { get; set; } = default!;

[...]

PR #5994, opened 2024-10-02
feat: add EventNotification

cfgiugliano:
This adds the DomainEvent activity as proposed in https://github.com/elsa-workflows/elsa-core/issues/5583

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5994)


cfgiugliano:
@dotnet-policy-service agree

sfmskywalker:
Would it make sense to use INotificationSender instead of IWebhookDispatcher?

sfmskywalker:
I was more thinking that we want to be able to publish domain events, which are "lower-level" than webhook events. Ergo, it doesn't make sense to me to use a service called "webhook dispatcher" when it has nothing to do with webhooks. For that reason, I'm thinking that using INoticationSender is more suitable. The webhook dispatcher uses this too under the hood. Alternatively, we could consider adding a new service called e.g. `IDomainEventDispatcher` (which would probably still use INotificationSender, just like IWebhookDispatcher).

cfgiugliano:
But the activity execution is already publishing a the domain event as notification using INotificationSender (this could be chanced to consume a IDomainEventDispatcher, yes) but the current handler is just a default implementation of the notification handler like for RunTask. If required, it is still possible to implement a custom handler and handle the domain event with custom logic

cfgiugliano:
I've added a IDomainEventDispatcher with both a background and synchronous implementation, and exposed a factory in WorfklowsRuntimeFeature just like for the ITaskDispatcher, then updated the handler under webhook.

Issue #6004, opened 2024-10-07
Customize collection names when using MongoDB


rosca-sabina:
### Discussed in https://github.com/elsa-workflows/elsa-core/discussions/5977

<div type='discussions-op-text'>

<sup>Originally posted by **rosca-sabina** September 20, 2024</sup>
Hello,

When using MongoDB for workflow management and runtime storage, is there a way to set a custom collection name for the the collections created by Elsa (such as workflow_instances, workflow_definitions etc.)?

I have a scenario where multiple microservices that will use Elsa use the same MongoDB database. So I need a way to keep the Elsa collections used by each service separate, even if they're in the same database.

Thanks!</div>

sfmskywalker:
Thanks for the PR that adds this capability!

PR #6005, opened 2024-10-07
Add MongoDB collection naming strategy

rosca-sabina:
Hello! This PR addresses issue #6004.

I created a `ICollectionNamingStrategy` interface, to allow customizing the names of the MongoDB collections created by Elsa Workflows. By creating their own implementation for this interface, users would be able to add a prefix etc. to the names of the collections. I used DI for resolving the `ICollectionNamingStrategy` implementation to allow users to inject `IConfiguration` or `IOptions<T>` in their implementations, in case for instance they want to get their prefix from a config file.

The scenario that inspired this PR was multiple services that use Elsa Workflows using the same MongoDB database. So they needed a way to differentiate between each service's Elsa collections. Also, I noticed the EF Core and Elasticsearch implementations already offer something similar to this (through `IElsaDbContextSchema` and respectively `IIndexNamingStrategy`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6005)

### DIFF
--- a/src/modules/Elsa.MongoDb/Common/PersistenceFeatureBase.cs
+++ b/src/modules/Elsa.MongoDb/Common/PersistenceFeatureBase.cs
@@ -1,5 +1,6 @@
 using Elsa.Features.Abstractions;
 using Elsa.Features.Services;
+using Elsa.MongoDb.Contracts;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


sfmskywalker (approved):
Looks great!

PR #6013, opened 2024-10-09
Register MongoDb client as singleton

rosca-sabina:
This PR changes the DI registration of `IMongoDatabase`. In the existing implementation, `IMongoDatabase` is registered as scoped, and the factory method creates a new instance of `MongoClient` each time `IMongoDatabase` is resolved.

According to the MongoDB documentation ([here](https://www.mongodb.com/docs/drivers/csharp/current/faq/#how-does-connection-pooling-work-in-the-.net-c--driver-) and [here](https://www.mongodb.com/community/forums/t/best-practice-for-singleton-vs-multiple-mongoclients-when-you-need-to-authenticate-with-different-users/125935/2)), the `MongoClient` is meant to be used as a singleton, because it maintains a pool of connections to the MongoDB server.

In this PR, I registered an instance of `IMongoClient` as singleton, and then resolved that instance from the `IServiceProvider` in the factory method for `IMongoDatabase`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6013)

### DIFF
--- a/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
+++ b/src/modules/Elsa.MongoDb/Features/MongoDbFeature.cs
@@ -41,7 +41,9 @@ public class MongoDbFeature : FeatureBase
     {
         Services.Configure(Options);
 
-        Services.AddScoped(sp => CreateDatabase(sp, ConnectionString));
+        var mongoUrl = new MongoUrl(ConnectionString);
+        Services.AddSingleton(sp => CreateMongoClient(sp, mongoUrl));
+        Services.AddScoped(sp => CreateDatabase(sp, mongoUrl));
 
         RegisterSerializers();
         RegisterClassMaps();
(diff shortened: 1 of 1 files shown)


rosca-sabina:
Hi @sfmskywalker! Is there a release 3.2.1 planned? If so, do you think it could be possible to include this fix in it?

It would be really useful for everyone using MongoDB with Elsa Workflows. With the way `IMongoDatabase` is currently registered, I started getting MongoDb timeouts and high CPU and memory usage after a while with a service I deployed, which was due to `WorkflowInboxCleanupHostedService` opening a new `MongoClient` every time it ran (the timeouts and high resource usage stopped after when I set `SweepInterval` to 24 hours, so that's how I figured out it was because of this service).

sfmskywalker:
Working on it now 👍🏻

sfmskywalker (approved):
Great find about the MongoClient!

PR #6069, opened 2024-10-30
Enable CPM transitive pinning and resolve version conflicts

lahma:
Transitive pinning enables `Dictory.Packages.props` to mandate the versions so no need to have per-project STJ version policies, packages file version will flow into restore targets as minimum. Upgraded some package in net6.0 target as having for example SemanticKernel as dependency does not allow 7.0.0 series dependencies and thus restore fails.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6069)


sfmskywalker (approved):
That cleans up nicely. Many thanks @lahma !

PR #6080, opened 2024-10-31
Add bookmark queue purger functionality

sfmskywalker:
Introduce a new class `DefaultBookmarkQueuePurger` to purge old bookmark queue items. This includes an interface `IBookmarkQueuePurger` and a recurring task `PurgeBookmarkQueueRecurringTask`, with necessary updates to `IBookmarkQueueStore` implementations and application configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6080)

### DIFF
--- a/src/apps/Elsa.Server.Web/Program.cs
+++ b/src/apps/Elsa.Server.Web/Program.cs
@@ -492,11 +492,12 @@ services
 // Obfuscate HTTP request headers.
 services.AddActivityStateFilter<HttpRequestAuthenticationHeaderFilter>();
 
-// Configure recurring tasks.
+// Optionally configure recurring tasks using alternative schedules.
 services.Configure<RecurringTaskOptions>(options =>
 {
     options.Schedule.ConfigureTask<TriggerBookmarkQueueRecurringTask>(TimeSpan.FromSeconds(30));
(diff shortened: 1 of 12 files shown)


englebertnicolas:
Hello, it seems that the DefaultBookmarkQueuePurger does not purge tenant-related rows in case of multitenancy. The purger only deletes BookmarkQueueItems rows that have TenantId Is Null (Elsa 3.3.0). 

I also tried to inject a custom IBookmarkQueuePurger in order to set BookmarkQueueFilter.TenantAgnostic = true, but it does not work.

```
public async Task PurgeAsync(CancellationToken cancellationToken = default) {
    ...
        var filter = new BookmarkQueueFilter {
            CreatedAtLessThan = thresholdDate,
            TenantAgnostic = true //LINE ADDED -> but does not work, this filter parameter does not seems to be taken into account...
        };
        var order = new BookmarkQueueItemOrder<DateTimeOffset>(x => x.CreatedAt, OrderDirection.Ascending);
        var page = await store.PageAsync(pageArgs, filter, order, cancellationToken);
    ...
}
```

sfmskywalker:
@englebertnicolas I created an issue for it. Thanks for the heads up.

PR #6083, opened 2024-10-31
[Feat: OTEL + Tenants] : Add TenantId Property to Span 

jdevillard:
This PR add the `TenantId` property to the span for :
- Workflow Span
- Activity Span

The aim is to facilitate the querying in the end analysis tools

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6083)


sfmskywalker:
@jdevillard We can approve & merge this PR as soon as its marked "Ready for review".

jdevillard:
I wasn't sure about the tenantId vs TenantId in the Log. On OTEL side, it's explain that attribute naming SHOULD be lowercase, so If we want to respect, we should change a lot of attribute,  let's go using the convention already present in this file (`camelcase`) and let discuss later on this point

sfmskywalker:
Agreed, the casing is a separate discussion and shouldn't hold back this PR 👍🏻

PR #6093, opened 2024-11-02
Switch MemoryStream to Microsoft.IO.RecyclableMemoryStream.

### LINKED ISSUE #6087: [CHORE] Use Microsoft.IO.RecyclableMemoryStream
### Task Overview
**What maintenance task do you propose? Please describe.**
In the code new `MemoryStream()`s are created 7 times. This could be optimized with pooling by using the `Microsoft.IO.RecyclableMemoryStream` nuget.

### Purpose of the Task
**Explain why this task is necessary**
Improved performance.

### Potential Issues
**Identify any potential issues that might arise from this task**
Can't think of any.

### Additional Context
**Add any other context**
Repo: [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream)

adamradocz:
Closes #6087

Initial commit switching from `MemorySteam` to `Microsoft.IO.RecyclableMemoryStream`.
What do you think about the implementation in general? If OK, I have added a few comments in the code that should be addressed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6093)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -56,6 +56,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="4.11.0" />
     <PackageVersion Include="Microsoft.EntityFrameworkCore.Sqlite.Design" Version="1.1.6" />
     <PackageVersion Include="Microsoft.Identity.Client" Version="4.66.1" />
+    <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <PackageVersion Include="Microsoft.SemanticKernel" Version="1.25.0" />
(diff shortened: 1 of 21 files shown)


bobhauser:
It feels to me that this might also be accomplished using an IMemoryStreamManager interface (perhaps just specifying the public methods in RecyclableMemoryStreamManager) and injecting the implementation. The default implementation could very well create an instance using the defaults you specify in your helper class. I think making the stream manager injectable would be useful for a number of reasons
- The parent process host Elsa might already have a RecyclableMemoryStreamManager which they might want to reuse, and an IMemoryStreamManager implementation might be created to reference it.
- The default settings you specify might not be appropriate in all instances
- Perhaps the usage of memory streams in a particular application is very occasional, and it might be better to just use bare MemoryStreams to avoid losing the allocated (and cached) memory forever
- It might be desired to attach to the RecyclableMemoryStreamManager events, perhaps to log when a stream is requested to expand beyond the maximum length

sfmskywalker:
I think @bobhauser is proposing to introduce a new interface called `IMemoryStreamManager` rather than using static helper classes. I haven't immersed myself into this PR yet, but do you think that might be a good idea @adamradocz ?

BasJanssenQuadira:
Is this a feature that is still wanted? I would be open to creating a proposal for an interface and a facade/adapter class to make the `RecyclableMemoryStreamManager` injectable.

PR #6097, opened 2024-11-06
Fix hangfire resume workflow

sNakiex:
There was a bug in the hangfire implementation where it would always generate a random workflowinstanceid because it was not passed into the createclient. 
Because of this hangfire was unable to resume instances.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6097)


sfmskywalker (approved):
Great catch!

PR #6105, opened 2024-11-07
Undo Migrations reset

### LINKED ISSUE #6103: [BUG] Elsa 3.3 is incompatible with existing 3.1 and 3.2 databases
## Description
If you're upgrading an Elsa v3.1 or v3.2 project to v3.3, you'll run into an error, saying that "WorkflowDefintions table already exists".

## Steps to Reproduce
1. Create a project with Elsa 3.2 using EFC + SqlServer for example.
2. Run it once to have the Schema created
3. Upgrade Elsa nugets to 3.3
4. Run it again
5. WorkflowDefintions table already exists error will show up

## Expected Behavior
Just works and silently upgrades the db schema to 3.3

## Actual Behavior
Crashes

## Root cause
In #6059 you removed all existing migrations and created a new V3_3 migration. The way how migrations in EFC works, is that it enumerates all migrations in the assembly, then consults the `__EFMigrationsHistory` table and executes all migrations not listed in that table. Since V3_3 is not listed, it will be executed, but since you've removed all migrations, V3_3 does contain the entire schema and not just the delta from V3_2, meaning it will attempt to create tables that already exist and fail. The proper procedure to reset all migrations in a safe way is listed here https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/managing?tabs=dotnet-core-cli#resetting-all-migrations

Suchiman:
fixes #6103

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6105)


Suchiman:
I've also noticed that the 3.2 Migrations appear to be all empty, those could be safely deleted to reduce the number of migrations

Suchiman:
This is a partial revert of only #6059

sfmskywalker (approved):

PR #6145, opened 2024-11-25
Fix the memory surge caused by script running and not being GC controlled for a long time

fanslead:
When there are a large number of C # scripts, frequent compilation and running of scripts can cause memory to skyrocket and not be released by GC for a long time.It will cause the server to run out of memory and crash.
By pointing scriptState to a null reference to break the reference relationship, it can be released by GC as early as possible, thereby reducing memory resource consumption.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6145)

### DIFF
--- a/src/modules/Elsa.CSharp/Services/CSharpEvaluator.cs
+++ b/src/modules/Elsa.CSharp/Services/CSharpEvaluator.cs
@@ -51,7 +51,11 @@ public class CSharpEvaluator(INotificationSender notificationSender, IOptions<CS
         script = notification.Script.ContinueWith(expression, scriptOptions);
         script = GetPreCompiledScript(script);
         var scriptState = await script.RunAsync(globals, cancellationToken: cancellationToken);
-        return scriptState.ReturnValue;
+
+        var returnValue = scriptState.ReturnValue;
+        scriptState = null;
+
+        return returnValue;
     }
 
     private Script<object> GetPreCompiledScript(Script<object> script)


Suchiman:
To my knowledge, this changes effectively nothing. `scriptState` is a local variable, not a field, it stops existing the moment the method returns which happens immediately after returning `ReturnValue`.

fanslead:
> To my knowledge, this changes effectively nothing. `scriptState` is a local variable, not a field, it stops existing the moment the method returns which happens immediately after returning `ReturnValue`.

@Suchiman I have tested it before, before adding this operation,It will cause the server to run out of memory and crash. GC is in effect, but it seems that the situation of freeing memory is not good.
image
By pointing scriptState to a null reference to break the reference relationship.During each GC, a large amount of memory can be released
image

sfmskywalker:
@fanslead Are you sure the difference in GC activity is caused by setting the local variable scriptState to null? Because that really makes no sense given how GC works - at least, how I *think* I know it works. The local variable will cease to exist as soon as the method returns. What would be more logical is if the caller of that method holds on to the returned value.
Perhaps you are able to explain how nullifying a local variable has this effect? My fear is that it *seems* to have an effect, but that there's another factor at play that causes the different results that you're seeing.

sfmskywalker:
@Suchiman I just noticed that the caching of the Script object might be ineffective, because the script is **always** created, and then cached:

```csharp
private Script<object> GetPreCompiledScript(Script<object> script)
    {
        var cacheKey = "csharp:script:" + Hash(script);

        return memoryCache.GetOrCreate(cacheKey, entry =>
        {
            if (_csharpOptions.ScriptCacheTimeout.HasValue)
                entry.SetSlidingExpiration(_csharpOptions.ScriptCacheTimeout.Value);

            return script;
        })!;
    }
```

Notice that the script object is simply passed in as an argument, and then added to te cache before being returned.
Is this on purpose, and if so, how does this work exactly? Unless there's some magic going on, it would seem to me that the script object is created every time a script is being evaluated, rather than the script being fetched from cache only.

fanslead:
> @fanslead Are you sure the difference in GC activity is caused by setting the local variable scriptState to null? Because that really makes no sense given how GC works - at least, how I _think_ I know it works. The local variable will cease to exist as soon as the method returns. What would be more logical is if the caller of that method holds on to the returned value. Perhaps you are able to explain how nullifying a local variable has this effect? My fear is that it _seems_ to have an effect, but that there's another factor at play that causes the different results that you're seeing.

@sfmskywalker When by pointing scriptState to a null reference and Immediately execute GC forced recycling. Memory can be stably released. But immediate GC will lead to increased CPU consumption. 
```
var scriptState = await script.RunAsync(globals, cancellationToken: cancellationToken);

var returnValue = scriptState.ReturnValue;

scriptState = null;
GC.Collect();
GC.WaitForPendingFinalizers();
GC.Collect();
```
image

But,When I do not change any. The server memory quickly fills up and crashes 
image

fanslead:
> @fanslead A more logical explanation for the increase in memory consumption is the fact that the scripts are cached. The larger the number of scripts, the larger the amount of scripts being stored in cache. To see if this is the case, maybe try and see what happens when you update the following method:
> 
> ```cs
> private Script<object> GetPreCompiledScript(Script<object> script)
> {
>     var cacheKey = "csharp:script:" + Hash(script);
> 
>     return memoryCache.GetOrCreate(cacheKey, entry =>
>     {
>         if (_csharpOptions.ScriptCacheTimeout.HasValue)
>             entry.SetSlidingExpiration(_csharpOptions.ScriptCacheTimeout.Value);
> 
>         return script;
>     })!;
> }
> ```
> 
> Remove the caching logic:
> 
> ```cs
> private Script<object> GetPreCompiledScript(Script<object> script)
> {
>     return script;
> }
> ```

I try caching the script. Of course, it is effective, but if there are a large number of different scripts and the memory is not released for a long time, it will also fill up the server's memory and cause it to crash. Meanwhile, when running with cache, there will also be a slight increase in memory, about 5-10MB.  If cache is not used, the memory growth during each compilation and execution of the script is approximately 200Mb-300Mb

sfmskywalker:
Ok. I'd like to do some experiments as well. In order to reproduce the memory consumption, what's the number of C# scripts we're looking at, give or take?

Suchiman:
@fanslead 
> If cache is not used, the memory growth during each compilation and execution of the script is approximately 200Mb-300Mb

Caching of the scripts is a very recent addition (new in 3.3-RC3), was the leak observed with or without the caching? Not caching the scripts might actually lead to memory exhaustion, because i'm not sure, if the dynamically generated assemblies from the scripting engine are ever freed. IIRC they didn't use to, because roslyn couldn't generate collectible assemblies so they stuck around forever, but i don't know if that changed.

fanslead:
> @fanslead
> 
> > If cache is not used, the memory growth during each compilation and execution of the script is approximately 200Mb-300Mb
> 
> Caching of the scripts is a very recent addition (new in 3.3-RC3, released 14h ago), was the leak observed with or without the caching? Not caching the scripts might actually lead to memory exhaustion, because i'm not sure, if the dynamically generated assemblies from the scripting engine are ever freed. IIRC they didn't use to, because roslyn couldn't generate collectible assemblies so they stuck around forever, but i don't know if that changed.

@Suchiman I think caching scripts can certainly suppress memory growth to a certain extent. However, cached scripts will not be released by GC, and when the number of different scripts reaches a certain level, the memory will still be full.The best solution should be to ensure that the script is promptly reclaimed by GC.

fanslead:
@Suchiman @sfmskywalker  I made a mistake. The server environment and VS Debug environment are different. When I force GC in VS, the memory can be stably released, but in the server environment, this effect doesn't look good.
I also tried caching scripts. I used a 2C8G server and with only 3 C # scripts, the memory usage after caching reached 40%.

sfmskywalker:
@fanslead Do you agree that the proposed change in this PR doesn't solve the memory consumption issue, given that the local variable being set to null , will automatically be set to null as soon as the method exits?

fanslead:
> @fanslead Do you agree that the proposed change in this PR doesn't solve the memory consumption issue, given that the local variable being set to null , will automatically be set to null as soon as the method exits?

Yes, After observation, the memory situation deployed on the server and VS Debug shows inconsistency.Currently unable to perfectly solve the problem of C # compilation memory usage not being released.

sfmskywalker:
Then I'll proceed to close this PR. Please feel free to open a separate issue with your observations if you want. Thanks.

fanslead:
> @fanslead #6192 also changes a bit how this work, i'm not confident about it but maybe that helps?

I have also tried using CreateDelegate, but I don't feel it has effect.

noapiosss:
after executing 4-5 complicated workflows, my server memory has gone (8gb), is there some advice how can i handle it?

fanslead:
> after executing 4-5 complex workflows, my server memory has gone (8gb), is there some advice how can i handle it? is there some way to force memory release? (im using cs scripts as inputs)

change to use js script.

noapiosss:
> > after executing 4-5 complex workflows, my server memory has gone (8gb), is there some advice how can i handle it? is there some way to force memory release? (im using cs scripts as inputs)
> 
> change to use js script.

so there is no chance to release memory using cs scipts?

PR #6147, opened 2024-11-25
Fix/32953

raymonddenhaan:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6147)

### DIFF
--- a/src/modules/Elsa.MassTransit.AzureServiceBus/Handlers/RemoveOrphanedSubscriptions.cs
+++ b/src/modules/Elsa.MassTransit.AzureServiceBus/Handlers/RemoveOrphanedSubscriptions.cs
@@ -45,7 +45,7 @@ public class RemoveOrphanedSubscriptions(MessageTopologyProvider topologyProvide
             }
             catch (ServiceBusException ex) when (ex.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
             {
-                logger.LogWarning(ex, "Service bus entity {entityPath} was not found", ex.EntityPath);
+                logger.LogWarning("Unable to remove orphaned subscription: Service bus entity {entityPath} was not found, most likely due to it being removed automatically", ex.EntityPath);
             }
         }
     }
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -55,6 +55,8 @@
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.7" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="4.10.0" />
     <PackageVersion Include="Microsoft.EntityFrameworkCore.Sqlite.Design" Version="1.1.6" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.2" />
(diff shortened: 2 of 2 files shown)


MariusVuscanNx (approved):

src/modules/Elsa.MassTransit.AzureServiceBus/Handlers/RemoveOrphanedSubscriptions.cs
```diff
@@ -45,7 +45,7 @@ await Task.WhenAll(
             }
             catch (ServiceBusException ex) when (ex.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
             {
-                logger.LogWarning(ex, "Service bus entity {entityPath} was not found", ex.EntityPath);
+                logger.LogWarning("Unable to remove orphaned subscription: Service bus entity {entityPath} was not found, most likely due to it being removed automatically", ex.EntityPath);
```
sfmskywalker, line 48: As far as I know, it is recommended to include the exception with the log statement. If so, then at the very least we should include a comment why we're deviating from that norm.

raymonddenhaan: Looking at those sources, there is no mentioning of logging the exception for warnings as well. However there is mention of avoiding the exception altogether if possible. What solution has your preference?

1. Add a comment why we are not logging the entire exception.
2. Put the exception back in the warning (This will cause the warning to be confused with an exception again)
3. Check if the subscription and topic which we are trying to delete here actually exist. This will add 2 calls for every subscription and instance that has been deemed inactive.

sfmskywalker: On second thought, I understand now after having thought it through: the exception is expected, so there's no reason to catch the details. The scenario is clear: we want to try and delete orphaned subscriptions, but that might already have happened, so catching the exception is in order.

Which does make me wonder if the WARN log level is appropriate - does this condition give cause for concern (warning) or is it simply a healthy condition and would INFO or DEBUG be more appropriate?

raymonddenhaan: Regardless of why the subscription or queue has already been deleted, it means the system is in the desired state. Looking at it like I think INFO is more appropriate. DEBUG might be a bit too low a level for someone who wants some insights into these cleaning processes

sfmskywalker: @raymonddenhaan 

1. Adding a comment would be sufficient, just to avoid overzealous team members to potentially add the exception back ;)
2. Although I disagree with the reason to remove it to avoid confusing it with an ERROR log level, I do think the exception details is completely unnecessary because we're simply logging information that an attempt to delete a subscription failed because it doesn't exist anymore. On that note, we should be **sure** that it no longer exists. In other words, if the failure to delete the subscription is caused by something else, then a warning + exception would be justified. However, as is clear from the exception handling condition, we are catching this very specific case, so no additional exception detail is necessary.
3. Exactly, I think simply trying to delete it and catching the exception as done here is the right approach - otherwise, as you said, we'd have to make a second call in case the resource exists in order to then delete it. Since we need to make an HTTP call anyway (to delete it), might as well keep it at one request.

PR #6148, opened 2024-11-25
Fix Serialization of generic types including collections

Suchiman:
#5682 broke serialization in several ways:
- When a generic type was being serialized, such as ``List`1``, the `` ` `` was being escaped by `JsonSerializer.Serialize` to `"\u0060"`. The PolymorphicObjectConverter would then `.Trim('"')` without unescaping the `` ` ``, causing `"\\u0060"` to be written into the final json.
- When the type is generic with one type parameter, e.g. `List<T>`, and also `IEnumerable<T>`, as `List<T>` is, then it would append `[]` at the end of the type name, resulting in: `"System.Collections.Generic.List\\u00601[[Elsa.Workflows.IntegrationTests.Serialization.JsonSerialization.WorkflowAction, Elsa.Workflows.IntegrationTests]], System.Private.CoreLib[]"` which is very invalid. It may only do that, if the typename was simplified because the `elementType` is in the `WellKnownTypeRegistry`.
- If it is a collection and the `elementType` is in the `WellKnownTypeRegistry`, then you'd be deserializing an `string[]` where you've serialized a `List<string>` because it throw's away that information and just appends `[]` which is questionable but i didn't change that here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6148)

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
+++ b/src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
@@ -271,13 +271,10 @@ public class PolymorphicObjectConverter(IWellKnownTypeRegistry wellKnownTypeRegi
         {
             if (shouldWriteTypeField)
             {
-                var typeOptions = newOptions.Clone();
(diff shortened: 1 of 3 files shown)


sfmskywalker (approved):
Looks good, thanks @Suchiman !

PR #6157, opened 2024-11-28
Fix Primitive Collection serialization

Suchiman:
This was broken by both #5871 and #5682:
1. When a `List<Guid>` was serialized, it was recognized as a primitive collection and thus plainly written to the JSON without any type information: `["d4d8404c-4357-47ff-a343-649a116539f5"]`
2. When this JSON was deserialized, due to lack of type info, it was deserialized as `List<object>`, containing `string`s. This is already not good.
3. When this `List<object>` gets serialized again (e.g. due to multiple workflow suspends causing `WorkflowState` serialization), this time it fails the primitive collection recognition, because `object` is not a primitive type. It now gets serialized as `{"_items": ["d4d8404c-4357-47ff-a343-649a116539f5"], "_type": "Object[]"}`
4. When that JSON gets deserialized, it tries to `ReadType(...)` but `ReadType` fails to parse `Object[]` since it lacks the logic from `TypeJsonConverter` to throw away the `[]` before looking up `Object` in the `WellKnownTypeRegistry`, so it returns `null` as a type. Without a type but being faced with a json object `{ ... }` it now deserializes into an `ExpandoObject`
5. Any further serialization / deserializations will now cause the expando object to get nested deeper and deeper every time.

What is the motivation behind the `WellKnownTypeRegistry`? It seems to hugely complicate the codebase for little benefit and is handled inconsistently, it is being used in many places without the special logic around recognizing collections, which sometimes works on ICollection, sometimes on Lists, attaching `[]` or not, using the word "array" when a `List` is meant, etc.

Also is there a reason for reinventing the polymorphic serialization wheel, when STJ supports it out of the box?

~~Additionally i've adjusted the `Write` method to write the `_type` field first so that the `Read` method does not have to skip over all the content while searching for `_type`.~~

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6157)

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
+++ b/src/modules/Elsa.Workflows.Core/Serialization/Converters/PolymorphicObjectConverter.cs
@@ -199,8 +199,8 @@ public class PolymorphicObjectConverter(IWellKnownTypeRegistry wellKnownTypeRegi
         if (type == typeof(JObject) || type == typeof(JArray) || type == typeof(JsonObject) || type == typeof(JsonArray))
         {
             writer.WriteStartObject();
-            writer.WriteString(TypePropertyName, type.GetSimpleAssemblyQualifiedName());
             writer.WriteString(IslandPropertyName, value.ToString());
+            writer.WriteString(TypePropertyName, type.GetSimpleAssemblyQualifiedName());
             writer.WriteEndObject();
             return;
         }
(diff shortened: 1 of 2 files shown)


Suchiman:
I'll have a look at the test failures later, i need to sleep x_x

Suchiman:
@sfmskywalker good to go

Suchiman:
rebased onto 97b59974d9517498a3c33ffc8234c2e3724ee0ab

sfmskywalker (approved):
Looks good to me, thanks @Suchiman 👍🏻

sfmskywalker:
> What is the motivation behind the WellKnownTypeRegistry

When serializing .NET types, they tend to be rather long-winded when serializing the fully namespace and assembly qualified type name, so the idea was to maybe serialize them using an alias.

I agree that its usefulness compared to the amount of complexity it introduces is questionable. If you see a way to remove it while maintaining compatibility with existing data that use type aliasing, I wouldn't object.

sfmskywalker:
> Also is there a reason for reinventing the polymorphic serialization wheel, when STJ supports it out of the box?

I could be wrong, but when this code was written, it was targeting .NET 6, where System.Text.Json didn't have the same set of capabilities as .NET 8. When we drop .NET 6 support, maybe that would be a nice opportunity to revisit this converter.
Long term, I'm hoping to revise the entire serialization story. It's currently all over the place with different combinations of converters and settings for a variety of scenarios, which makes the system very brittle to change in the area of serialization.

Issue #6163, opened 2024-11-29
[BUG] JavaScript BigInt mapping to BigInteger serialization incorrect


fanslead:
When using BigInt to handle long types, it will be mapped to C #'s BigInteger structure. If BigInteger is directly serialized, it will become this structure。
```
{
	"isPowerOfTwo": false,
	"isZero": false,
	"isOne": false,
	"isEven": true,
	"sign": 1
}
```
Need to add a BigIntegerJsonConverter

PR #6164, opened 2024-11-29
fix JavaScript BigInt mapping to BigInteger serialization incorrect

fanslead:
#6163

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6164)


MrHPotter:
Maybe a Test Serializing and Deserializing with an Options Object Created with this Converter included, would be nice, to prove it works as expected. (and keep it working in the future)

fanslead:
> Maybe a Test Serializing and Deserializing with an Options Object Created with this Converter included, would be nice, to prove it works as expected. (and keep it working in the future)

@MrHPotter You're right, I've already added testing

sfmskywalker (approved):
Awesome 👍🏻

PR #6175, opened 2024-12-03
Fix JS Variable Setting Bug in Workflow Variable Sync Logic  

sfmskywalker:
This PR addresses a bug where variables set using the `set{VariableName}` JavaScript syntax were being overridden during the workflow variable synchronization process. The issue occurred because the `variables` object, which serves as a copy of the workflow variables, was empty and incorrectly copied back, overwriting the manually set values.

### Fix:  
- Added logic to check if copied variables have been modified.  
- Only modified variables are copied back into the workflow variables.  
- Retains values set using `set{VariableName}` syntax, ensuring expected behavior.

This ensures that user-defined variable values persist without interference from the synchronization logic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6175)

### DIFF
--- a/test/component/Elsa.Workflows.ComponentTests/Scenarios/JavaScriptVariables/JavaScriptVariablesWorkflow3.cs
+++ b/test/component/Elsa.Workflows.ComponentTests/Scenarios/JavaScriptVariables/JavaScriptVariablesWorkflow3.cs
@@ -1,0 +1,16 @@
+using Elsa.Extensions;
+using Elsa.JavaScript.Activities;
+
+namespace Elsa.Workflows.ComponentTests.Scenarios.JavaScriptVariables;
+
+public class JavaScriptVariablesWorkflow3 : WorkflowBase
+{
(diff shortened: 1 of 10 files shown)


test/component/Elsa.Workflows.ComponentTests/Scenarios/JavaScriptVariables/JavaScriptVariablesWorkflow3.cs
raymonddenhaan, line 1: Why would we need 3 the same workflows for these tests?

raymonddenhaan (approved):

test/component/Elsa.Workflows.ComponentTests/Scenarios/JavaScriptVariables/JavaScriptVariablesWorkflow3.cs
raymonddenhaan: Nvm, I see the differences now

PR #6183, opened 2024-12-05
Introduce job unscheduling functionality

sfmskywalker:
Add a new method `UnscheduledAsync` to unschedule jobs across various components, including the job queue, background activity scheduler, and Hangfire integration. This enhancement provides a more robust way to remove jobs from scheduling, complementing the existing cancellation functionality.

This fixes an issue where a background activity execution job that resumed a workflow, which in turn would remove any associated bookmarks, which in turn would cancel the background job while that job is still executing and has to e.g. persist changes made to the DB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6183)


src/modules/Elsa.Workflows.Runtime/Handlers/CancelBackgroundActivities.cs
raymonddenhaan, line 1: Since we are no longer cancelling anything here we should rename this class

raymonddenhaan (approved):

PR #6192, opened 2024-12-08
Fix C# Script<T> race conditino

Suchiman:
Compilation of the Script wasn't entirely thread safe, specifically, if the same script was being evaluated twice at the same time on an empty cache:

Thread 1: Puts Script<T> into cache, then starts compiling it
Thread 2: Finds not yet compiled Script<T> in cache, starts compiling it as well
Thread 1: Done compiling, Assembly.Load it for running
Thread 2: Done compiling, Assembly.Load it <--- Exception: Assembly with name XYZ already loaded

The Solution is to use CreateDelegate which compiles the script the moment it is being put into the cache and the cache then only serves precompiled delegates that are immediately runnable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6192)


sfmskywalker (approved):
Great catch!

PR #6199, opened 2024-12-11
Introduce the possibility to exclude variable value retrieval for cer…

MariusVuscanNx:
…tain storage drivers

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Contracts/IVariablePersistenceManager.cs
+++ b/src/modules/Elsa.Workflows.Core/Contracts/IVariablePersistenceManager.cs
@@ -6,9 +6,12 @@ namespace Elsa.Workflows;
 public interface IVariablePersistenceManager
 {
     /// <summary>
-    /// Loads the variables into the specified <see cref="WorkflowExecutionContext"/>. 
+    /// Loads the variables into the specified <see cref="WorkflowExecutionContext"/>.
     /// </summary>
-    Task LoadVariablesAsync(WorkflowExecutionContext context);
+    /// <param name="context"></param>
+    /// <param name="excludeValueRetrievalForLargeDataDrivers"></param>
+    /// <returns></returns>
+    Task LoadVariablesAsync(WorkflowExecutionContext context, bool excludeValueRetrievalForLargeDataDrivers = false);
 
     /// <summary>
     /// Persists all persistable variables from the specified <see cref="WorkflowExecutionContext"/>. 
--- a/src/modules/Elsa.Workflows.Core/Enums/StorageDriverTag.cs
+++ b/src/modules/Elsa.Workflows.Core/Enums/StorageDriverTag.cs
@@ -1,0 +1,6 @@
+﻿namespace Elsa.Workflows;
+
+public enum StorageDriverTag
+{
+    LargeData
+}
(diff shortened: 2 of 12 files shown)


src/modules/Elsa.Workflows.Core/Contracts/IVariablePersistenceManager.cs
```diff
@@ -6,9 +6,12 @@ namespace Elsa.Workflows;
...
 {
     /// <summary>
-    /// Loads the variables into the specified <see cref="WorkflowExecutionContext"/>. 
+    /// Loads the variables into the specified <see cref="WorkflowExecutionContext"/>.
     /// </summary>
-    Task LoadVariablesAsync(WorkflowExecutionContext context);
+    /// <param name="context"></param>
+    /// <param name="excludeValueRetrievalForLargeDataDrivers"></param>
+    /// <returns></returns>
+    Task LoadVariablesAsync(WorkflowExecutionContext context, bool excludeValueRetrievalForLargeDataDrivers = false);
```
sfmskywalker, line 14: It would be better to accept a set of Storage Driver Tags, because the term "exclude value retrieval for large data drivers" sounds very specific to a single tag. Given the introduction of tags, which I like, it would follow that we can use it here to filter by.

src/modules/Elsa.Workflows.Core/Enums/StorageDriverTag.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Elsa.Workflows;
+
+public enum StorageDriverTag
```
sfmskywalker, line 3: Using a closed enum limits the system from being extended in the future. If we're going to introduce tagging, let's keep it open-ended and instead use strings or (long) integers.

src/modules/Elsa.Workflows.Core/Services/VariablePersistenceManager.cs
```diff
@@ -34,6 +34,9 @@ public async Task LoadVariablesAsync(WorkflowExecutionContext workflowExecutionC
                 if (driver == null)
                     continue;
 
+                if (excludeValueRetrievalForLargeDataDrivers && driver.Tags.Contains(StorageDriverTag.LargeData))
+                    continue;
```
sfmskywalker, line 38: When you replace the `excludeValueRetrievalForLargeDataDrivers` parameter with a list of tags, you can replace this logic by simply checking to see if the driver contains all of the specified tags.

src/modules/Elsa.Workflows.Core/VariableStorageDrivers/WorkflowInstanceStorageDriver.cs
```diff
@@ -17,8 +17,11 @@ public class WorkflowInstanceStorageDriver : IStorageDriver
     /// The key used to store the variables in the workflow state.
     /// </summary>
     public const string VariablesDictionaryStateKey = "Variables";
-
+    
+    /// <inheritdoc />
     public double Priority => 5;
+    /// <inheritdoc />
+    public StorageDriverTag[] Tags => [];
```
sfmskywalker, line 24: Public list-type properties are types as `IEnumerable` or `ICollection` everywhere else, so it would be good to repeat that here. In this case, I would opt for an `IEnumerable`.

sfmskywalker (changes requested):
I like the premise, now fix the story ;)

PR #6200, opened 2024-12-11
Fix Quartz scheduler implementation

sfmskywalker:
This PR fixes an issue where jobs would scheduled only once because their key would already exist. Instead, jobs are now stored durably and we check to see if a trigger already exists, rather than a job.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6200)

### DIFF
--- a/src/modules/Elsa.Quartz/Services/JobKeyProvider.cs
+++ b/src/modules/Elsa.Quartz/Services/JobKeyProvider.cs
@@ -1,0 +1,19 @@
+using Elsa.Common.Multitenancy;
+using Elsa.Quartz.Contracts;
+using Quartz;
+
+namespace Elsa.Quartz.Services;
+
+internal class JobKeyProvider(ITenantAccessor tenantAccessor) : IJobKeyProvider
+{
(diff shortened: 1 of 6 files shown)


src/modules/Elsa.Quartz/Services/JobKeyProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+using Elsa.Quartz.Contracts;
+using Quartz;
+
+namespace Elsa.Quartz.Services;
+
+internal class JobKeyProvider(ITenantAccessor tenantAccessor) : IJobKeyProvider
+{
+    public JobKey GetJobKey<TJob>() where TJob : IJob
+    {
+        return new(typeof(TJob).Name, GetGroupName());
```
raymonddenhaan, line 11: Will `typeof(TJob).Name` generate a unique name or will this be the same for each workflowDefinition? If not, should we not use something like the instanceId here instead?

raymonddenhaan (approved):

PR #6210, opened 2024-12-13
Refactor variable handling in storage driver logic

sfmskywalker:
Refactored `StorageDriverContext` to include `Variable` and updated its usage across relevant methods to ensure context accuracy. Enhanced `ObjectConverter` to support conditional deserialization of JSON objects. Adjusted `WorkflowInstanceStorageDriver` to utilize improved variable resolution with new conversion options.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6210)

### DIFF
--- a/src/modules/Elsa.Workflows.Core/VariableStorageDrivers/WorkflowInstanceStorageDriver.cs
+++ b/src/modules/Elsa.Workflows.Core/VariableStorageDrivers/WorkflowInstanceStorageDriver.cs
@@ -1,6 +1,7 @@
 using System.ComponentModel.DataAnnotations;
 using System.Text.Json;
 using System.Text.Json.Nodes;
+using Elsa.Expressions.Helpers;
 using Elsa.Extensions;
 using JetBrains.Annotations;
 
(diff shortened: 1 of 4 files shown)


src/modules/Elsa.Workflows.Core/VariableStorageDrivers/WorkflowInstanceStorageDriver.cs
```diff
@@ -39,7 +40,15 @@ public ValueTask WriteAsync(string id, object value, StorageDriverContext contex
     {
         var dictionary = GetVariablesDictionary(context);
         var node = dictionary.GetValueOrDefault(id);
-        return new(node);
+        var variable = context.Variable;
+        var variableType = variable.GetVariableType();
+        var options = new ObjectConverterOptions
+        {
+            DeserializeJsonObjectToObject = true,
+            SerializerOptions = payloadSerializer.GetOptions()
```
raymonddenhaan, line 48: Could we just inject the options instead of the entire payloadSerializer if that is all that we are using?

sfmskywalker: Those options aren’t injectable, but if you have something in mind I’m all ears.

Issue #6227, opened 2024-12-19
[BUG] Outputs from activities which are run asynchronously not captured in ActivityExecutionRecord


bobhauser:
## Description
There is an option on activities to have them run asynchronously (when Kind == ActivityKind.Task). When these activities are executed asynchronously their outputs are not properly reported in the associated ActivityExecutionRecord - only outputs that are actively mapped to a variable appear, unmapped outputs are always reported as null.

Additionally, only mapped outputs are reported in the activity output register, but incorrectly - all mapped outputs are reported to the activity output register with the output name of "output". This implies that you would not ever be able to map the outputs from an activity which was run asynchronously to a subsequent activity.

## Steps to Reproduce

1. Create a test activity similar to this
```
[Activity("Elsa.Test", Kind = ActivityKind.Task)]
internal class SampleActivity : CodeActivity
{

    public SampleActivity()
    {
        RunAsynchronously = true;
    }

    public Input<int>? Number1 { get; set; } = default;
    public Input<int>? Number2 { get; set; } = default;
    public Output<int>? Sum { get; set; } = default;
    public Output<int>? Product { get; set; } = default;

    protected override void Execute(ActivityExecutionContext context)
    {
        int number1 = context.Get(Number1);
        int number2 = context.Get(Number2);

        context.Set(Sum, number1 + number2);
        context.Set(Product, number1 * number2);
    }
}
```
2. Run a workflow that uses this activity
```
TestWorkflow workflow = new(workflowBuilder =>
	{
		workflowBuilder.DefinitionId = "sampleWorkflow";
		workflowBuilder.Id = "sampleWorkflow";

		var variable1 = new Variable<int>();
		workflowBuilder.Root = new Sequence
		{
			Variables =
			{
				variable1
			},
			Activities =
			{
				new SampleActivity()
				{
					Id = "SampleActivity1",
					RunAsynchronously = runAsynchronously,
					Number1 = new(4),
					Number2 = new(8),
					Sum = new(variable1),
					Product = null, 
				}
			}
		};
	}
);
```
3. Notice that the associated ActivityExecutionRecord only contains the mapped Sum output value, and that the unmapped Product output is indicated to be null.
![image](https://github.com/user-attachments/assets/b6a615a7-455e-4981-844e-ba8dd3359ccb)
4. Also notice that activity output register only shows the value for the mapped value. And that the name of the output is incorrectly set as "output", not "Sum"
[...]

PR #6228, opened 2024-12-19
Updated output handling of asynchronously run workflows to be the same as when run synchronously

bobhauser:
These changes result in outputs of activities run asynchronously to be treated the same as if the activity were run synchronously with regards to ActivityExecutionRecord and the activity output register.

See https://github.com/elsa-workflows/elsa-core/issues/6227 for more details.

### DIFF
--- a/src/common/Elsa.Testing.Shared.Integration/WorkflowExtensions.cs
+++ b/src/common/Elsa.Testing.Shared.Integration/WorkflowExtensions.cs
@@ -1,0 +1,89 @@
+﻿using Elsa.Extensions;
+using Elsa.Features.Services;
+using Elsa.Mediator.Contracts;
+using Elsa.Workflows;
+using Elsa.Workflows.Notifications;
+using Elsa.Workflows.Runtime;
+using Elsa.Workflows.Runtime.Contracts;
+using Elsa.Workflows.Runtime.Requests;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Elsa.Testing.Shared;
+
+public static class WorkflowExtensions
+{
+    public static async Task<WorkflowFinished?> DispatchWorkflowAndRunToCompletion(
+        this IWorkflow workflowDefinition,
+        Action<IServiceCollection>? configureServices = null,
+        Action<IModule>? configureElsa = null,
+        string? instanceid = null,
+        string? definitionVersionId = null,
+        TimeSpan? timeout = default)
+    {
+        SemaphoreSlim semaphore = new (0, 1);
+        WorkflowFinished? workflowFinishedRecord = null;
+
+        var host = Host.CreateDefaultBuilder()
+            .ConfigureServices(services =>
+            {
+                configureServices?.Invoke(services);
+
+                // This notification handler will capture the WorkflowFinished record (to be returned) and release the semaphore
+                services.AddNotificationHandler<WorkflowFinishedAction, WorkflowFinished>(sp => new WorkflowFinishedAction(notification =>
+                {
+                    workflowFinishedRecord = notification;
+                    semaphore.Release();
+                }
+                ));
+
+                services.AddElsa(elsa => {
+                    configureElsa?.Invoke(elsa);
+                });
+            })
+            .Build();
+
+        try
+        {
+            // Start the Host
+            await host.StartAsync(CancellationToken.None);
+
+            await host.Services.PopulateRegistriesAsync();
+
+            // Build the workflow
+            IWorkflowBuilderFactory workflowBuilderFactory = host.Services.GetRequiredService<IWorkflowBuilderFactory>();
+            var workflow = await workflowBuilderFactory.CreateBuilder().BuildWorkflowAsync(workflowDefinition);
+
+            // Register the workflow
+            var workflowRegistry = host.Services.GetRequiredService<IWorkflowRegistry>();
+            await workflowRegistry.RegisterAsync(workflow);
+
+            // Dispatch the workflow
+            var workflowDispatcher = host.Services.GetRequiredService<IWorkflowDispatcher>();
+            var dispatchWorkflowResponse = await workflowDispatcher.DispatchAsync(new DispatchWorkflowDefinitionRequest()
+            {
+                DefinitionVersionId = "sampleWorkflow",
+                InstanceId = "instanceId"
+            });
+            dispatchWorkflowResponse.ThrowIfFailed();
+
+            // Wait for the workflow to complete, and then return the WorkflowFinished notification
+            bool signaled = await semaphore.WaitAsync(timeout ?? TimeSpan.FromSeconds(5));
+            return signaled ? workflowFinishedRecord : null;
+        }
+        finally
+        {
+            // Stop the Host
+            await host.StopAsync(CancellationToken.None);
+        }
+    }
+
+    class WorkflowFinishedAction(Action<WorkflowFinished> action) : INotificationHandler<WorkflowFinished>
+    {
+        public Task HandleAsync(WorkflowFinished notification, CancellationToken cancellationToken)
+        {
+            action(notification);
+            return Task.CompletedTask;
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


test/integration/Elsa.Workflows.IntegrationTests/Scenarios/RunAsynchronousActivityOutput/Tests.cs
```diff
@@ -0,0 +1,164 @@
...
+        WorkflowFinished? workflowFinishedRecord = await workflow.DispatchWorkflowAndRunToCompletion(
+            configureServices: services =>
+            {
+                services.AddScoped<DistributedWorkflowRuntime>();
+            },
+            configureElsa: elsa =>
+            {
+                elsa.UseWorkflowRuntime(workflowRuntime => {
+                    workflowRuntime.ActivityExecutionLogStore = sp => activityExecutionStore;
+                    workflowRuntime.WorkflowRuntime = sp => sp.GetRequiredService<DistributedWorkflowRuntime>();
```
bobhauser, line 141: Not sure if this is a bug or not... but to be able to run multiple asynchronous activities in parallel (the only way it makes sense to me) the new DistributedWorkflowRuntime workflow runtime must be used. I don't believe that this was not the case with 3.2.

src/modules/Elsa.Workflows.Runtime/Middleware/Activities/BackgroundActivityInvokerMiddleware.cs
```diff
@@ -122,7 +122,7 @@ private static void CaptureOutputIfAny(ActivityExecutionContext context)
                 continue;
 
             var output = (Output?)outputDescriptor.ValueGetter(activity);
-            context.Set(output, outputEntry.Value);
+            context.Set(output, outputEntry.Value, outputDescriptor.Name);
```
bobhauser, line 125: This fixes the output name always being set as "output" in the activity output register.

src/modules/Elsa.Workflows.Runtime/Services/BackgroundActivityInvoker.cs
```diff
@@ -90,35 +90,8 @@ private async Task ResumeWorkflowAsync(ActivityExecutionContext activityExecutio
...
-
-            var outputValue = activityExecutionContext.Get(memoryBlockReference);
-
-            if (outputValue != null)
-                outputValues[outputDescriptor.Name] = outputValue;
-        }
-
-        return outputValues;
+    {
+        var activityOutputRegister = activityExecutionContext.WorkflowExecutionContext.GetActivityOutputRegister();
```
bobhauser, line 94: I fear that this "solution" may be too simplistic since the old code clearly went out of it's way to only return values that are backed by `WorkflowInstanceStorageDriver`. I'm not sure that this is a concern now considering that the parent workflow seems to be now resumed in the same thread. Anyway - I'd welcome discussion in this particular change (as well as any other change!).

sfmskywalker: The concern is that when an activity produces a large object, e.g. a file or JSON string of 5 MB, the user might use a different variable storage driver, such as "Blob Storage Driver" that stores the data in Azure Storage.

For this reason, we want to avoid those large values to become part of the output register, given that it will be serialized as well.

Obviously, the current implementation is rather crude and intended to be a temporary hack.
The ultimate solution is to allow for better control of how output is recorded.

For example, for small values, it might be OK to be recorded. The assumption here is that if the storage driver is "Workflow Instance", which stores the value as part of the workflow instance, it's safe to also include it as part of the output register.

For large values, the assumption is that the user will have used the Memory storage driver or a custom one such as "Blob Storage Driver". In this case, we definitely do not want the value to be recorded in the output register.

Now, we actually do want to record _something_ in the output register. In the case of large blobs, a download link could be appropriate.

So perhaps a cleaner approach would be to leverage the associated storage driver to have it produce a value suitable for storing in the output register. This would clean up this code because now we no longer need to test against the storage driver type.

Pending that solution, perhaps for now it might be best to keep the code to only include variables into the output  if those variables are associated with a workflow instance driver.

test/integration/Elsa.Workflows.IntegrationTests/Scenarios/RunAsynchronousActivityOutput/Tests.cs
sfmskywalker: This was indeed intentional to require the `DistributedWorkflowRuntime` or `ProtoActorWorkflowRuntime` to have support for parallel activity execution. To be honest, I'd love it to be able to still have support for parallel activity execution using the `LocalWorkflowRuntime`. Originally, the `DistributedWorkflowRuntime` was created for hosting Elsa on multiple nodes, which requires distributed locking. But it's fair to expect that when running on a single node, you should still be able to run activities in parallel. Perhaps this could be added easily by updating the `LocalWorkflowRuntime` with a semaphore to synchronise access to the workflow instance.

sfmskywalker (changes requested):
Thanks for a wonderful PR, and even with tests!
I think that the only thing left to be able to merge is the `BackgroundActivityInvoker` (see comments)

src/modules/Elsa.Workflows.Runtime/Services/BackgroundActivityInvoker.cs
bobhauser: Hmmm. Having inclusion in the output register dependent on the associated storage driver implies that unmapped (unbound) outputs would continue to not be registered in the output register when activities are run asynchronously. And having synchronously executed activities behave differently than asynchronously executed activities also feels odd (for synchronously executed activities I think all outputs are registered regardless of being bound or not, and regardless of storage driver). And, in fact, for bound outputs I think it is less important to be included in the output register since down stream activities can just map to the variable.

Perhaps the best solution would be for me to introduce a `IBackgroundActivityOutputExtractor` service where the default implementation would do exactly what the existing (prior to my PR) `BackgroundActivityInvoker.ExtractActivityOutput` method does and would allow me to use an implementation that instead returns all outputs (basically what I have in this PR). What do you think of this @sfmskywalker?

bobhauser: Actually, considering the bug is actually more about the logging of outputs in the `ActivityExecutionRecord` and not about the activity output register (although there is certainly overlap)... how about we use the activities specified persistence mode to determine whether to return the output from the background activity (also taking into account any outputs which are bound to WorkflowInstanceStorageDriver variables)? So, if an activity executed asynchronously, we would return outputs that are either configured to LogPersistenceMode.Include or bound to a WorkflowInstanceStorageDriver variable.

test/integration/Elsa.Workflows.IntegrationTests/Scenarios/RunAsynchronousActivityOutput/Tests.cs
bobhauser: My solution for this was to use a memory based implementation of distributed lock, and then to just use DistributedWorkflowRuntime. The main snag I ran into when I tried a difference approach is that DistributedWorkflowClient internally uses LocalWorkflowClient. If you are interested in the memory based implementation of distributed lock let me know, otherwise I'll just keep it local.

src/modules/Elsa.Workflows.Runtime/Services/BackgroundActivityInvoker.cs
sfmskywalker: That makes a lot of sense 👍🏻

sfmskywalker:
@bobhauser I created [this PR](https://github.com/elsa-workflows/elsa-core/pull/6542) that solves the issue in a similar way (using output register) but reuses the activity state mapper functionality so that it also applies log output property filtering. I'll be combining your tests with that PR. Thank you for your nice contributions 👍🏻

src/common/Elsa.Testing.Shared.Integration/WorkflowExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+using Elsa.Workflows.Runtime.Contracts;
+using Elsa.Workflows.Runtime.Requests;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Elsa.Testing.Shared;
+
+public static class WorkflowExtensions
+{
+    public static async Task<WorkflowFinished?> DispatchWorkflowAndRunToCompletion(
```
sfmskywalker, line 16: That's a really cool and clever way of setting up a test host, very flexible. Taking this in 👍🏻 👍🏻

sfmskywalker:
Closing this PR in favor of https://github.com/elsa-workflows/elsa-core/pull/6542. Credits to @bobhauser for invaluable input and to this PR from which I included the tests.

PR #6233, opened 2024-12-22
Drop .NET 6 Support

sfmskywalker:
This PR removes support for .NET 6.0 from the project. Updates include:  
- Adjusted project files to target only supported .NET versions.  
- Updated documentation and CI/CD pipelines to reflect the removal of .NET 6.0.  
- Removed any .NET 6.0-specific code or dependencies.  

This change simplifies maintenance and focuses development on currently supported .NET versions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6233)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -16,35 +16,35 @@
     <PackageVersion Include="Azure.ResourceManager" Version="1.13.0" />
     <PackageVersion Include="Azure.ResourceManager.AppContainers" Version="1.3.0" />
     <PackageVersion Include="Azure.ResourceManager.Resources" Version="1.9.0" />
-    <PackageVersion Include="Azure.ResourceManager.ServiceBus" Version="1.0.1" />
+    <PackageVersion Include="Azure.ResourceManager.ServiceBus" Version="1.1.0" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
     <PackageVersion Include="Bogus" Version="35.6.1" />
     <PackageVersion Include="ConfigureAwait.Fody" Version="3.3.2" PrivateAssets="All" />
     <PackageVersion Include="Confluent.Kafka" Version="2.6.1" />
-    <PackageVersion Include="Confluent.SchemaRegistry.Serdes.Avro" Version="2.4.0" />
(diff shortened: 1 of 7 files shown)


Directory.Packages.props
```diff
@@ -1,234 +1,152 @@
...
-    <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="7.0.18" />
-    <PackageVersion Include="Oracle.EntityFrameworkCore" Version="7.21.13" />
-    <PackageVersion Include="Polly" Version="7.2.4" />
-    <PackageVersion Include="Pomelo.EntityFrameworkCore.MySql" Version="7.0.0" />
-    <PackageVersion Include="Refit" Version="7.2.22" />
-    <PackageVersion Include="Refit.HttpClientFactory" Version="7.0.0" />
-    <PackageVersion Include="System.Formats.Asn1" Version="8.0.1" />
-    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
jayachandra21, line 152: Hello @sfmskywalker Removing this causing problem in .Net8.0 Target framework packages

jayachandra21 (commented):
In the Directory.packages.props file by removing the <ItemGroup Condition="'$(TargetFramework) is created a problem

Directory.Packages.props
sfmskywalker: That's odd - the Elsa.Server.Web project seems to build fine when targeting .NET 8.0.

Do you have a repro?

jayachandra21: I think as Target Framework 6.0 is removed, particular  to those package section can be removed.
For TargetFramework 8.0 and 9.0, can we still go with validation Condition check as below 
<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'"><PackageVersions>.

sfmskywalker: We could, but it’s better if we don’t have to, keeps things simple. I haven’t seen any issues so far with referencing 9.0 packages. Can you show me the issue?

PR #6281, opened 2025-01-09
Simplify and optimize workflow instance creation.

sfmskywalker:
- Replaced asynchronous instance creation methods with streamlined synchronous alternatives where applicable.

- Introduced `CreateAndCommitWorkflowInstanceAsync` for combined instantiation and persistence, along with a separate `CreateWorkflowInstance` method for non-committal instantiation. 

- Refactored related code to improve readability, maintainability, and runtime performance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6281)


src/modules/Elsa.Workflows.Runtime/Services/LocalWorkflowClient.cs
```diff
@@ -130,16 +97,71 @@ public async Task ImportStateAsync(WorkflowState workflowState, CancellationToke
...
+            Status = workflowState.Status,
+            SubStatus = workflowState.SubStatus,
+            Incidents = workflowState.Incidents
+        };
+    }
+    
+    public async Task<WorkflowInstance> CreateInstanceInternalAsync(CreateWorkflowInstanceRequest request, CancellationToken cancellationToken = default)
+    {
+        var workflowDefinitionHandle = request.WorkflowDefinitionHandle;
+        var workflowGraph = await workflowDefinitionService.FindWorkflowGraphAsync(workflowDefinitionHandle, cancellationToken);
```
raymonddenhaan, line 141: Should we not use the `GetWorkflowGraphAsync` method here?

raymonddenhaan (approved):

src/modules/Elsa.Workflows.Runtime/Services/LocalWorkflowClient.cs
sfmskywalker: `GetWorkflowGraphAsync` doesn't have an overload accepting a workflow definition handle, but I think I can update `GetWorkflowGraphAsync` such that it only accepts workflow definition handles and then have the other callers pass in a handle, which encapsulates DefinitionVersionId, which is really what the current `GetWorkflowGraphAsync` is after. This would be more elegant and reduces some duplication 👍🏻

PR #6309, opened 2025-01-16
Regenerate 3_3 Runtime module migrations

sfmskywalker:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6309)

### DIFF
--- a/src/modules/Elsa.Dapper.Migrations/Runtime/V3_3.cs
+++ b/src/modules/Elsa.Dapper.Migrations/Runtime/V3_3.cs
@@ -20,6 +20,7 @@ public class V3_3 : Migration
         Alter.Table("WorkflowExecutionLogRecords").AddColumn("TenantId").AsString().Nullable();
         Alter.Table("ActivityExecutionRecords").AddColumn("TenantId").AsString().Nullable();
         Alter.Table("KeyValuePairs").AddColumn("TenantId").AsString().Nullable();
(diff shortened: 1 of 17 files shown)


src/modules/Elsa.Dapper.Migrations/Runtime/V3_3.cs
```diff
@@ -20,6 +20,7 @@ public override void Up()
         Alter.Table("WorkflowExecutionLogRecords").AddColumn("TenantId").AsString().Nullable();
         Alter.Table("ActivityExecutionRecords").AddColumn("TenantId").AsString().Nullable();
         Alter.Table("KeyValuePairs").AddColumn("TenantId").AsString().Nullable();
+        Rename.Column("Key").OnTable("KeyValuePairs").To("Id");
```
raymonddenhaan, line 23: It seems we are still deleting the WorkflowInboxMessages for Dapper (line17) while we are not doing that for other migrations

raymonddenhaan (approved):

raymonddenhaan (approved):

src/modules/Elsa.Dapper.Migrations/Runtime/V3_3.cs
sfmskywalker: That's a mistake. Good catch :+1:

PR #6322, opened 2025-01-22
[New Activity] Command Executer

IbrahimNadaa:
This Simple lightweight Activity Scans your Assembly and register your custom-made commands 
(**The ones that inherits the class `ElsaWorkFlowCommand`**)

Then execute them in Elsa Context. 

But what about the Command payload?

It can be sent in the form of  `Json` and the command will automatically
map the json to the command object. Yes ! using reflection. 

# Example use case: 

after an endpoint got hit , i need to save the data at the database or i need to send an OTP , of course currently Elsa support using webhooks and submit the body to an Api, But this one saves you an extra layer of work which is the REST Api ....... 
you can use it as follows : 
image

But how can i send paylaod of the command ? 

simply like this :  of course you can use Elsa variables inputs and so on to build up you payload, and it can take a very complex objects 
image

when Elsa executes this activity 
it runs the handler , in case of exception the workflow become faulted with the exception details for revive.
image

# Getting started with this activity:

at the `Program.cs`

```
    builder.Services.AddElsa(elsa =>
    {
            ....... 
      else.UseCommandExecuter();
    }
```
and then you have to register your handles to Elsa Mediator like this 

`builder.Services.AddCommandHandlersFrom<YoutCommandsAssembly>();`

# Example of a command and a handler : 

```
public class CreateOrderCommand : ElsaWorkFlowCommand
{
    public string CustomerName { get; set; }
    public DateTime OrderDate { get; set; }
}

public class CreateOrderCommandHandler : ICommandHandler<CreateOrderCommand>
{
    public async Task<Unit> HandleAsync(CreateOrderCommand command, CancellationToken cancellationToken)
    {
        Console.WriteLine("Order submitted!");
        return new Unit();
    }
}

```

then enjoy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6322)


IbrahimMNada:
@sfmskywalker 

kindly check this when ever you can

PR #6328, opened 2025-01-24
Fix variable default value serialization and array type parsing

sfmskywalker:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6328)

### DIFF
--- a/src/modules/Elsa.Workflows.Management/Mappers/VariableDefinitionMapper.cs
+++ b/src/modules/Elsa.Workflows.Management/Mappers/VariableDefinitionMapper.cs
@@ -58,8 +58,9 @@ public class VariableDefinitionMapper(IWellKnownTypeRegistry wellKnownTypeRegist
     {
         var variableType = source.GetType();
         var valueType = variableType.IsConstructedGenericType ? variableType.GetGenericArguments().FirstOrDefault() ?? typeof(object) : typeof(object);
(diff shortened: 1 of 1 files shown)


src/modules/Elsa.Workflows.Management/Mappers/VariableDefinitionMapper.cs
```diff
@@ -61,19 +58,19 @@ public VariableDefinition Map(Variable source)
     {
         var variableType = source.GetType();
         var valueType = variableType.IsConstructedGenericType ? variableType.GetGenericArguments().FirstOrDefault() ?? typeof(object) : typeof(object);
-        var isArray = valueType.IsCollectionType();
-        var elementValueType = isArray ? valueType.GenericTypeArguments[0] : valueType; 
+        var isArray = valueType.IsArray;
+        var isCollection = valueType.IsCollectionType();
+        var elementValueType = isArray ? valueType.GetElementType() : isCollection ? valueType.GenericTypeArguments[0] : valueType;
```
raymonddenhaan, line 63: Will this still work as intended for a dictionary with only using the first genericTypeArguments?

sfmskywalker: I am pretty sure yes, but I'll do a test to be 100% sure.

PR #6359, opened 2025-01-31
Fix Oracle EF Core Migrations

sfmskywalker:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6359)


kylemwhite:
It looks like this change assumes that Oracle has a boolean datatype. That is only true for the latest version of Oracle. Thus, this change breaks Oracle for anything less than 23c (23ai). Is the intention to discontinue support for older versions of Oracle?

image

PR #6370, opened 2025-02-04
Added Alterations API to Client API library and updated server API co…

yinzara:
This PR adds the Alterations API to the API client library.  This will enable the use of these APIs within the Elsa Studio application.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6370)


src/modules/Elsa.Alterations/Endpoints/Alterations/DryRun/Endpoint.cs
yinzara, line 1: I also opted to fix the existing comments that were incorrect.

sfmskywalker (approved):
Great PR, thanks!

PR #6375, opened 2025-02-06
Exposed incidents in the Elsa.Api.Client

MariusVuscanNx:


sfmskywalker:
Please create a PR targeting the main branch as well with these updates in order to not loose them when migrating to 3.3.

MariusVuscanNx:
> Please create a PR targeting the main branch as well with these updates in order to not loose them when migrating to 3.3.

I will cherry pick the changes in main and raspberry

Issue #4433, opened 2023-09-13
[ENH] Consider joining "active" branches


Nokecy:
```
public class FlowJoinConditionWaitAllWorkflow : IWorkflow
    {
        public ValueTask BuildAsync(IWorkflowBuilder builder, CancellationToken cancellationToken = default)
        {
            var switchFlow = new FlowSwitch
            {
                CanStartWorkflow = true,

                Cases = new List<FlowSwitchCase>()
                {
                     new FlowSwitchCase() { Label="Case1" ,Condition = new JavaScriptExpression("1 == 1") },

                     new FlowSwitchCase() { Label="Case2" ,Condition = new JavaScriptExpression("1 != 1")  },
                },

                Mode = new(SwitchMode.MatchAny)
            };

            var userTask1 = new AdvancedUserTask()
            {
                Name = "userTask1",

                Cases = new(context => new List<string>() { "下一步userTask2" }),

                AssignUser = new(c => "admin123")
            };

            var userTask2 = new AdvancedUserTask
            {
                Name = "userTask2",

                Cases = new(context => new List<string>() { "下一步userTask3" }),

                AssignUser = new(c => "admin456")
            };

            var flowJoin = new FlowJoin
            {
                Mode = new(FlowJoinMode.WaitAll)
            };

            var finish = new Finish
            {
            };

            builder.Root = new Flowchart
            {
                Activities =
                {
                    switchFlow,

                    userTask1,

                    userTask2,

                    flowJoin,

                    finish
                },

                Connections =
                {
                    new Connection(new Endpoint(switchFlow,"Case1"),new Endpoint(userTask1,"In")),

                    new Connection(new Endpoint(userTask1,"下一步userTask2"),new Endpoint(flowJoin)),

                    new Connection(new Endpoint(switchFlow,"Case2"),new Endpoint(userTask2,"In")),

                    new Connection(new Endpoint(userTask2,"下一步userTask3"),new Endpoint(flowJoin)),

                    new Connection(new Endpoint(flowJoin),new Endpoint(finish)),
                }
            };

            return new ValueTask();
        }
    }
```

like this  ,Because Case2 will never execute! @sfmskywalker

sfmskywalker:
Hi @Nokecy , is this happening with the latest build?

sfmskywalker:
[...]

PR #6406, opened 2025-02-13
Update flowchart activity scheduling, implicit join (and FlowJoin/WaitAll) now only waits for followed connections

bobhauser:
This change updates the implicit join (and FlowJoin/WaitAll) to wait only for followed connections. It addresses issue [#4433](https://github.com/elsa-workflows/elsa-core/issues/4433). Initially, I intended to maintain an option to use the legacy behavior for flowchart scheduling. However, after further consideration, I decided against this approach as I could not identify any scenarios where the legacy behavior would be preferable.

## Flowchart Execution Model Enhancements
This update enhances the scheduling mechanism for flowchart activities, focusing on implicit joins, join activities, and backward connections. The new execution flow ensures that activities are scheduled only when necessary, preventing redundant executions and aligning with BPMN-style process control.

### Execution Tracking: Visiting vs. Following Connections
Connections in the flowchart can be either visited or followed:

- Visited: The connection was evaluated, regardless of whether it was followed.
- Followed: The connection was actively taken as part of the execution.
- Skipped: The connection was visited but not used in the execution.

The `FlowScope `class tracks visits using methods such as `RegisterActivityVisit()` and `RegisterConnectionVisit()`, ensuring the execution state is properly recorded. This prevents unnecessary re-execution of activities and ensures correct behavior for joins.

### Implicit WaitAll and FlowJoin(WaitAll) Behavior
The implicit `WaitAll `behavior, as well as the explicit `FlowJoin` with `WaitAll `mode, ensures that activities execute only after all their inbound connections have been visited. This aligns with BPMN-style process execution, where an activity must wait for all preceding activities to complete, regardless of whether the connections were followed or skipped. The execution rules are:

- An activity waits for all inbound connections to be visited (whether followed or skipped).
- If at least one inbound connection was followed, the activity executes normally.
- If all inbound connections were skipped, the activity is also skipped.
- This prevents activities from executing prematurely while allowing workflows to correctly propagate skipped execution paths.

### FlowJoin(WaitAny) Behavior
The `FlowJoin` activity with `WaitAny` mode completes once the first inbound connection is followed. Subsequent connections, whether followed or skipped, do not trigger the execution of outbound activities again.

### Backward Connections
A backward connection is an edge in the flowchart that moves against the typical forward execution flow. These connections are useful for loops, retries, or conditional branches but must be carefully handled to avoid invalid execution paths. The update ensures that backward connections are validated before execution:

- A connection is considered backward if it does not belong to the forward execution traversal.
- A backward connection is only valid if every possible execution path from the connection source to the root activity passes through its target. If an invalid backward connection is detected while scheduling outgoing activities, an exception will be thrown. (Without this check, the workflow could become stalled due to how child scheduling is managed.)
- When backward connections are utilized, care must be taken to prevent infinite loops (e.g., by using a FlowSwitch or FlowDecision to break the connection loop).

Tracking and validation are managed using `FlowGraph.IsBackwardConnection`, ensuring invalid backward links do not disrupt workflow execution.

### Dangling Activities
A dangling activity has no inbound forward connections, meaning it cannot be reached through normal execution flow. These activities are effectively orphaned unless explicitly triggered. The method `FlowGraph.IsDanglingActivity()` detects such cases. If a dangling activity completes unexpectedly, an exception will be thrown in `Flowchart.OnChildCompletedAsync()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6406)


sfmskywalker (approved):
Amazing PR. Many thanks! 🙏🏻

Issue #6161, opened 2024-11-28
ComputeBookmarkHash  in HttpWorkflowsMiddleware 


mehmet-hoke:
In HttpWorkflowsMiddleware class, ComputeBookmarkHash method

private string ComputeBookmarkHash(IServiceProvider serviceProvider, string path, string method)
{
var bookmarkPayload = new HttpEndpointBookmarkPayload(path, method);
var bookmarkHasher = serviceProvider.GetRequiredService();
var activityTypeName = ActivityTypeNameHelper.GenerateTypeName();
return bookmarkHasher.Hash(activityTypeName, bookmarkPayload);
}

ActivityTypeNameHelper.GenerateTypeName();

in this line it uses HttpEndpoint as a constant. But it records to the Bookmarks table with its customized name (in my case Elsa.HRVHTTPEndpoint) and creates the hash value in this way.
Therefore, it cannot find the record with the Hash value created with HttpEndpoint and the worklow does not continue.

Different type names work when creating(HRVHTTPEndpoint) a bookmark and consuming(HTTPEndpoint)

dmq-kmd:
I have the exact same issue. The mismatch between HttpEndpoint and the customized name (Elsa.HRVHTTPEndpoint) causes the workflow to fail when trying to retrieve the bookmark. @sfmskywalker Are there any plans to address this in the near future? 🙂

sfmskywalker:
Although I cannot commit to a timeline, this is definitely something I wish to add support for ("this" referring to the ability to derive custom activities from existing ones, without having to duplicate bookmarking resumption logic).

sfmskywalker:
I assigned it to the 3.4 milestone to keep my eyes on it, but as with any items added to any milestone, it is subject to be removed from scope if it would otherwise take too long for the milestone release to be delivered.

dmq-kmd:
I've made the necessary changes to the code that caused the problem:

This line:
var activityTypeName = ActivityTypeNameHelper.GenerateTypeName<HttpEndpoint>();

always assumes that activityTypeName comes from HttpEndpoint.

However, I don't have permission to publish my branch and create a PR. Is there a way you could grant me access? 🙂

sfmskywalker:
Reviewing the PR takes time, despite it being a small one. The PR targets a specific case, whereas I am looking for an approach that allows for trigger inheritance universally and not only this case. This may or may not be feasible, which will become clear when investigating the options in more detail. All of this takes time.

[...]

PR #6407, opened 2025-02-13
#6161 ComputeBookmarkHash in HttpWorkflowsMiddleware

dmq-kmd:
When creating custom triggers, the system uses activityTypeName to generate a hash. However, when the trigger is invoked, it searches through all the hashes in the database but does not use the activityTypeName of the trigger itself (for good reason, as it does not have access to it). Instead, it defaults to using the activityTypeName of HttpEndpoint.

As a result, it's not possible to create custom triggers with an activityTypeName different from HttpEndpoint, which leads to issues.

It is also mentioned here: #6161

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6407)

### DIFF
--- a/src/modules/Elsa.Http/Middleware/HttpWorkflowsMiddleware.cs
+++ b/src/modules/Elsa.Http/Middleware/HttpWorkflowsMiddleware.cs
@@ -1,4 +1,4 @@
-using System.Diagnostics.CodeAnalysis;
+﻿using System.Diagnostics.CodeAnalysis;
 using System.Net;
 using System.Net.Mime;
 using System.Text.Json;
@@ -368,10 +368,29 @@ public class HttpWorkflowsMiddleware(RequestDelegate next, ITenantAccessor tenan
         return await httpEndpointAuthorizationHandler.AuthorizeAsync(new AuthorizeHttpEndpointContext(httpContext, workflow, bookmarkPayload.Policy));
     }
 
-    private string ComputeBookmarkHash(IServiceProvider serviceProvider, string path, string method)
+    private static string ComputeBookmarkHash(IServiceProvider serviceProvider, string path, string method)
     {
         var bookmarkPayload = new HttpEndpointBookmarkPayload(path, method);
         var bookmarkHasher = serviceProvider.GetRequiredService<IStimulusHasher>();
-        return bookmarkHasher.Hash(activityTypeName: null, bookmarkPayload);
+        var httpEndpointActivityTypeName = ActivityTypeNameHelper.GenerateTypeName<HttpEndpoint>();
+        var newHash = bookmarkHasher.Hash(activityTypeName: null, bookmarkPayload);
+        var legacyHash = bookmarkHasher.Hash(activityTypeName: httpEndpointActivityTypeName, bookmarkPayload); // Backward compatible
+
+        if (WorkflowExists(serviceProvider, newHash))
+            return newHash;
+
+        return legacyHash; // Fallback for old workflows
+
+    }
+
+    private static bool WorkflowExists(IServiceProvider serviceProvider, string hash)
+    {
+        var triggerStore = serviceProvider.GetRequiredService<ITriggerStore>();
+        var bookmarkStore = serviceProvider.GetRequiredService<IBookmarkStore>();
+
+        var triggerExists = triggerStore.FindManyAsync(new TriggerFilter { Hash = hash }, CancellationToken.None).Result.Any();
+        var bookmarkExists = bookmarkStore.FindManyAsync(new BookmarkFilter { Hash = hash }, CancellationToken.None).Result.Any();
+
+        return triggerExists || bookmarkExists;
     }
 }
--- a/src/modules/Elsa.Http/Middleware/HttpWorkflowsMiddleware.cs
+++ b/src/modules/Elsa.Http/Middleware/HttpWorkflowsMiddleware.cs
@@ -1,26 +1,26 @@
-using Elsa.Extensions;
-using Elsa.Http.Bookmarks;
-using Elsa.Http.Options;
-using Elsa.Workflows.Runtime.Filters;
-using JetBrains.Annotations;
-using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


dmq-kmd:
@dotnet-policy-service agree

sfmskywalker:
@dmq-kmd I think there's now a better way for us to create custom HTTP Endpoint activities as described [here](https://docs.elsaworkflows.io/reusable-triggers-3.5-preview#httpendpointbase). Please give it a go and let mw know if you run into any limitations. If you agree this is better, then let's close this PR.

dmq-kmd:
@sfmskywalker That looks great and is definitely a cleaner approach. However, I'm running into an issue. When I create a custom endpoint like this:

```
public class CustomHttpEndpoint : HttpEndpointBase
{
    protected override HttpEndpointOptions GetOptions()
    {
        return new()
        {
            Path = "test",
            Methods = [HttpMethods.Get]
        };
    }

    protected override async ValueTask OnHttpRequestReceivedAsync(ActivityExecutionContext context, HttpContext httpContext)
    {
        httpContext.Response.StatusCode = 200;
        await httpContext.Response.WriteAsync("Hello World", context.CancellationToken);
    }
}
```

It seems like the endpoint isn’t being exposed as expected — at least I’m not able to call it. I’ve recorded a short video demonstrating the issue:
WorkflowHttpEndpointCantCall

sfmskywalker:
Thanks for the repro, I'll take a look soon.

dmq-kmd:
@sfmskywalker just checking in to see if you had a chance to look at the repro I shared. Let me know if there’s anything else I can provide to help debug the issue 🙌

sfmskywalker:
Not yet, but it's on my list for this week. I'll be sure to let you know if I have any questions 👍🏻

sfmskywalker:
@dmq-kmd I just tried to reproduce the issue, but for me it works fine. I'm wondering if your custom HTTP Endpoint activity is setup correctly, i.e. does it override `GetOptions` and then registers the Get verb and expected path?

Here's the custom HTTP endpoint activity that I tried with:

```csharp
public class CustomHttpEndpoint : HttpEndpointBase
{
    protected override HttpEndpointOptions GetOptions()
    {
        return new()
        {
            Path = "my-path",
            Methods = [HttpMethods.Get],
        };
    }

    protected override async ValueTask OnHttpRequestReceivedAsync(ActivityExecutionContext context, HttpContext httpContext)
    {
        httpContext.Response.StatusCode = 200;
        await httpContext.Response.WriteAsync("Hello World", context.CancellationToken);
    }
}
```

Here's my recording:

custom-http-endpoint

dmq-kmd:
Hey @sfmskywalker — thanks again for your help so far 🙌

I’ve matched your CustomHttpEndpoint implementation and confirmed that my workflow is published and uses the correct trigger. However, I’m still getting a 404 when calling the endpoint.

The one remaining difference I suspect is how the Elsa HTTP endpoints are registered in your setup.

I've added               

`   .AddActivity<CustomHttpEndpoint>()
`

to my ElsaFacility, but maybe I need to do something different? 😄

sfmskywalker:
@dmq-kmd That's really odd. If you want, please send me a small project that demonstrates what you are seeing. I'll take a closer look to see what's going on. Conversely, what happens if you copy in the SampleHttpEndpoint from elsa-core (main branch) into your project, does it work then? 

> I've added
>  `.AddActivity<CustomHttpEndpoint>()`

That looks all good 👍🏻

dmq-kmd:
@sfmskywalker I don't see anything named SampleHttpEndpoint? if you mean CustomHttpEndpoint, it still gets me 404 😒

sfmskywalker:
You will find it here: https://github.com/elsa-workflows/elsa-core/blob/main/src/apps/Elsa.Server.Web/Activities/CustomHttpEndpoint.cs

It's a sample I used for testing, and will be moved to elsa-samples at some point.

dmq-kmd:
Thank you @sfmskywalker. I've create a small repo: https://github.com/dmq-kmd/ElsaCustomHttpEndpoint where I get 404 when I add the CustomHttpEndpoint to a workflow, run the workflow and call a get request to `https://localhost:5001/api/workflows/my-path` in postman

sfmskywalker:
Hi @dmq-kmd , thank you for the rep.

I just tried it, but I could not reproduce the issue - everything works as expected. Maybe a couple of things to double-check:

- When I start the project using the launch profile, the application starts at https://localhost:14690/
  - This means that the workflow base path will be https://localhost:14690/api/workflows
- Make sure that your custom endpoint is marked as a trigger ("Trigger workflow" checked).
- Make sure that your workflow is **published**.

With that, I can successfully request https://localhost:14690/api/workflows/my-path from my browser.

dmq-kmd:
Thank you so much, @sfmskywalker! Turns out it was just me being a bit of an idiot 😅 Apparently, there's a difference between actually publishing the workflow and just running it — triggering works fine when published, but not when simply run (at least for me).

Sorry for wasting your time on this — I’ll go ahead and close it now. Thanks again!

sfmskywalker:
No worries, happy to hear it’s working! 😃

dmq-kmd:
Another quick question @sfmskywalker:
Does HttpEndpointBase support multitenancy in Elsa?

I noticed that everything works as expected when I return Unresolved() in my TenantResolver, but as soon as I return a specific tenant and publish the same trigger under that tenant, it returns a 404 when I call the endpoint.

Just wondering if this is expected behavior or if I might be missing a configuration detail 🤔

dmq-kmd:
Hi @sfmskywalker -  I think I’ve figured out what the issue was with multitenancy and custom HTTP endpoints.

When the trigger is indexed in the database, the TenantId field is not being set, it’s always stored as null. So when I create a workflow using my custom trigger while logged in under a specific tenant, the trigger record in the database doesn’t include that tenant context. As a result, the HTTP request results in a 404 because it can’t match the trigger correctly.

To confirm, I manually updated the database entry with the correct TenantId, and the endpoint started working as expected.

From what I can tell, the issue might be in the TriggerIndexer, specifically in the CreateWorkflowTriggersAsync method. It looks like the tenant ID may not be getting mapped there.

PR #5343, opened 2024-05-06
Conditional inputs elsa 3

af-git-64:
This pull request adds the necessary input attributes classes for https://github.com/elsa-workflows/elsa-studio/issues/186.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5343)

PR #5429, opened 2024-05-22
Add support for collections and dictionaries of input

bobhauser:
Implemented the ability specify collections and dictionaries of Inputs on activities. 

For example:
`public List<Input<string>>? Recipients { get; set; } = default;`
`public Dictionary<string, Input<string>>? Arguments { get; set; } = default;`

See #5421 for more details.

This change is not breaking in any way, and all existing tests pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/5429)

PR #6410, opened 2025-02-14
Conditional inputs

LukasBenner:
This pull request introduces new functionality to support conditional inputs in the Elsa workflows. The changes include adding new attributes and models to handle conditional inputs and state dropdowns, and updating existing descriptors to incorporate these new features.

This PR is a refined version of PR #5343 for the issue .
The corresponding PR in elsa-studio is 

### New Attributes and Models:

* [`src/modules/Elsa.Workflows.Core/Attributes/Conditional/ConditionalInputAttribute.cs`](diffhunk://#diff-ab2ba2cb98f633189fbce0abec6e6f361b08ac3aee0baede6fa2bc51db181719R1-R11): Introduced the `ConditionalInput` attribute to specify inputs that should only be shown for certain states.
* [`src/modules/Elsa.Workflows.Core/Attributes/Conditional/StateDropdownInput.cs`](diffhunk://#diff-2e177dd716439f623edeb7d668d0ff18a49442ba160dab5b3adf5efae69f6dc5R1-R14): Added the `StateDropdownInput` attribute to specify dropdown inputs with state options.
* [`src/modules/Elsa.Workflows.Core/Attributes/Conditional/ConditionalInputOptions.cs`](diffhunk://#diff-e42559eaed3bccdc1e999c85d9dca0a7245917404b05b74c2d3f823764e9f8c4R1-R16): Defined constants for different conditional input options.

### Updated Descriptors:

* [`src/clients/Elsa.Api.Client/Resources/ActivityDescriptors/Models/ConditionalDescriptor.cs`](diffhunk://#diff-17e2f178310abf38378d798579447bd9a544159bb51dd9eaeb701667c10e8c5cR1-R9): Created a new `ConditionalDescriptor` model to describe conditional inputs.
* [`src/modules/Elsa.Workflows.Core/Models/ConditionalDescriptor.cs`](diffhunk://#diff-955d9ec6fc333a29457eed95b299a023b3cf79ba9a56d7a1f6b6d4d59447c408R1-R11): Added a similar `ConditionalDescriptor` model within the workflows core module.

### Integration with Existing Code:

* [`src/modules/Elsa.Workflows.Core/Services/ActivityDescriber.cs`](diffhunk://#diff-5fa8a50dcf7bdbfa4329f0d4c331bf7bd444335eb9989d0d5ca78e657d60f643R186-R199): Updated the `DescribeInputPropertyAsync` method to include logic for handling conditional and state dropdown inputs.
* [`src/modules/Elsa.Workflows.Core/Attributes/InputAttribute.cs`](diffhunk://#diff-8d0eabc99bcfbc2cc9674412342c32b3eab632f930dbae90302b6b8372e8496dR101-R111): Added properties to the `InputAttribute` class to support conditional inputs and state dropdowns.

These changes enhance the Elsa workflows by allowing more dynamic and state-dependent input configurations.

A example usage of the conditional inputs could look like this:

```csharp
using Elsa.Extensions;
using Elsa.Workflows;
using Elsa.Workflows.Attributes.Conditional;
using Elsa.Workflows.Models;

namespace Elsa.Server.Web;

public class CondActivity : CodeActivity
{

    [StateDropdownInput(
        ["double", "string"],
        DefaultValue = "double")]
    public Input<string> ApiName{ get; set; } = default;
    
    [ConditionalInput(["double"], Description = "Double")]
    public Input<double> DoubleInput {get; set;} = default;

    [ConditionalInput(["string"], Description = "String")]
    public Input<string> StringInput {get; set;} = default;

    protected override void Execute(ActivityExecutionContext context)
    {
        object? value = null;
        if (ApiName.Get(context) == "double")
            value = DoubleInput.Get(context);
        else if (ApiName.Get(context) == "string")
            value = StringInput.Get(context);
        Console.WriteLine(value);
    }
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6410)


LukasBenner:
@dotnet-policy-service agree company="Mitutoyo CTL Germany GmbH"

bobhauser:
I'm not sure if your use case is similar to the one we had or not... Our use case is that we have a RunScript task that offers a list of separately controlled scripts that can be selected to be run. Each script has a a different set of parameters which could be configured. The solution we came up with was to support maps of script parameter and Input. So for one script you could have a "param1" parameter that is an Input<string>, and a "param2" that is an Input<bool> - and a different script could have completely different arguments. This is the PR that implements this: https://github.com/elsa-workflows/elsa-core/pull/5429. We don't use Elsa Studio, so there is certainly a gap there.

Your case could be completely different, and if it is, please just ignore this comment.

LukasBenner:
Thanks for your comment @bobhauser.
I think there are indeed some similarities between these PRs.
But I think the usecases aren't exactly the same.
The idea behind this PR is to show inputs in the UI based on the state of the activity which can be altered by e.g. a StateDropdownInput.

PR #6413, opened 2025-02-15
Add IBusRegistrationContext to Transport Configuration Actions

CODEdire:
This PR updates MassTransit configuration for Azure Service Bus and RabbitMQ. Previously, the IBusRegistrationContext was not included in the configuration action and prevents configuration of additional MassTransit components that require the context being passed in.

- `Program.cs`: Replaces `ConfigureServiceBus` with `ConfigureTransportBus` as the old action is now obsolete.
- `AzureServiceBusFeature.cs` and `RabbitMqServiceBusFeature.cs`: Marks `ConfigureServiceBus` as obsolete, introduces `ConfigureTransportBus`. Added XML documentation to bring awareness to the alternative action and some clarification that this action runs in addition to Elsa's standard configuration.
- `MassTransitFeature.cs`: Updates `ConfigureInMemoryTransport` method documentation and updates `AddMassTransit` method within to include `MassTransitOptions` configuration for prefetch count. While not directly part of the change, it was made because `PrefetchCount` was marked as obsolete in the feature class.

_**NOTE:** A name of ConfigureTransportBus was chosen for the new action to avoid conflicts with the old ConfigureServiceBus. Given that both ASB and RMQ are both transports in the eyes of MassTransit, it made the most sense to go this route without doing something like ConfigureServiceBusWithContext or similar._

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6413)

### DIFF
--- a/src/apps/Elsa.Server.Web/Program.cs
+++ b/src/apps/Elsa.Server.Web/Program.cs
@@ -481,7 +481,7 @@ services
 
                 if (massTransitBroker == MassTransitBroker.AzureServiceBus)
                 {
-                    massTransit.UseAzureServiceBus(azureServiceBusConnectionString, serviceBusFeature => serviceBusFeature.ConfigureServiceBus = bus =>
+                    massTransit.UseAzureServiceBus(azureServiceBusConnectionString, serviceBusFeature => serviceBusFeature.ConfigureTransportBus = (context, bus) =>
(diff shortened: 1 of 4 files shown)


CODEdire:
@microsoft-github-policy-service agree

PR #6416, opened 2025-02-17
Add AddManyAsync method to activity execution stores

sfmskywalker:
Introduced a new AddManyAsync method across multiple activity execution stores to add collections of log records. This enhancement ensures consistency in handling bulk additions and aligns with existing store interfaces.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6416)

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - 'blueberry'
       - 'feature/*'
+      - 'enh/*'
   release:
     types: [ prereleased, published ]
 env:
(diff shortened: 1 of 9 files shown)


sfmskywalker:
> Should we not use this AddMany method for the WorkflowExecutionLogRecords as well?

That's already the case.

> How will this work for blocking activities like Delay and waiting for events? Will the state of the activity not be changed between workflow instance runs?

none of that is affected by this. We are only exposing an additional method for consumption by the monitor.

sfmskywalker:
Ah, yes you are correct. I incorrectly assumed archive-only inserts, but of course there will be suspension points that flush completed and running activity execution records, the latter needing to be updated when activities resume.

Good catch 👍🏻

raymonddenhaan (approved):

Issue #6421, opened 2025-02-19
ActivityExecutionContext.Children collection is not populated


bobhauser:
## Description
The ActivityExecutionContext has a property called `Children`. The intention is that when the `ParentActivityExecutionContext` property is set on an `ActivityExecutionContext` the current context is added to its parent's Children set. See: https://github.com/elsa-workflows/elsa-core/blob/2c394e23536c2bb48aa12f7e5e8f086f1ef8b5cf/src/modules/Elsa.Workflows.Core/Contexts/ActivityExecutionContext.cs#L101

That property is not, however, used to set the parent ActivityExecutionContext. Instead the `_parentActivityExecutionContext` is assigned to directly here https://github.com/elsa-workflows/elsa-core/blob/2c394e23536c2bb48aa12f7e5e8f086f1ef8b5cf/src/modules/Elsa.Workflows.Core/Contexts/ActivityExecutionContext.cs#L47

The result of this is that the ActivityExecutionContext.Children set is always empty, and functionality that depends on it does not work as expected. One example of this can be found in Flowchart.HasPendingWork.

PR #6423, opened 2025-02-19
ActivityExecutionContext Children property not populated

bobhauser:
I updated ActivityExecutionContext to set the parent context to use the ParentActivityExecutionContext setter, which populates the parent's Children set. I also updated a few places to use the Children set instead of searching through all of the workflow's activity contexts.

See https://github.com/elsa-workflows/elsa-core/issues/6421 for more details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6423)


sfmskywalker (approved):
Sweet 👍🏻

PR #6452, opened 2025-02-26
Stop DefaultBookmarkQueuePurger log spamming

bobhauser:
Based on default scheduling, DefaultBookmarkQueuePurger logs two low value information log entries every 10 seconds. I could filter all of the logging out with logging config, but that would also filter out this seemingly useful log entry too
`logger.LogInformation("Purged {Count} bookmark queue items.", items.Count);`

This is how our logs looked before this change
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6452)


sfmskywalker (approved):

sfmskywalker:
Makes sense 👍🏻

PR #6457, opened 2025-02-27
2.x Added data column compression for Entity Framework persistance

TechnoDezi:
On the v2 workflows when creating and running large multi-step workflows the data stored on ms sql in json format gets very large.

This change adds compression on the serialized json to reduce the data stored and decompresses it again during loading and serialization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6457)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ src/designer/elsa-workflows-studio/loader
 # Created by developers using GNU/Linux
 .directory
 launchSettings.json
+/src/persistence/Elsa.Persistence.EntityFramework/Elsa.Persistence.EntityFramework.Core/Properties/PublishProfiles/FolderProfile.pubxml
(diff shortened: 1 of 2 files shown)


sfmskywalker:
@TechnoDezi This improvement makes sense, but it would be better if compression could be enabled or disabled. In V3, we implemented this feature as follows:

1. In `Program.cs`, we configure an option to specify the compression codec, defaulting to `None`.
2. The `WorkflowInstance` class includes an additional column to store the compression codec used. When decompressing, this stored codec is used, ensuring consistency even if the current codec configuration has changed. This allows for codec switching or disabling without issues.

You might consider a similar approach, where instead of embedding the compression logic directly in the store class, it resides in a strategy class. This way, you could define a default strategy that performs no compression and another that applies the compression method from this PR. The name of the strategy used during compression would then be stored alongside the record in a separate column, ensuring the correct strategy is applied during decompression.

sfmskywalker (changes requested):
It's great to have the option to compress the data column contents, but this should be configurable: the default should be no compression. Ideally, the selected compression code should be configurable from Startup/Program, just like Elsa 3.

sfmskywalker:
Triage: Elsa 2 support track (`2.x` base) — leaving open per Crew Lead PR-triage policy (changes requested).

PR #6460, opened 2025-03-02
Improved support for named workflow instances

sfmskywalker:
Introduced the ability to set and persist custom names for workflow instances. Updated relevant classes, services, and tests to ensure proper handling of the `Name` property.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6460)


src/modules/Elsa.Workflows.Management/Mappers/WorkflowStateMapper.cs
```diff
@@ -39,14 +37,12 @@ public void Apply(WorkflowState source, WorkflowInstance target)
...
         target.IsExecuting = source.IsExecuting;
         target.CorrelationId = source.CorrelationId;
+        target.Name = source.Name;
         target.IncidentCount = source.Incidents.Count;
         target.IsSystem = source.IsSystem;
         target.UpdatedAt = source.UpdatedAt;
         target.FinishedAt = source.FinishedAt;
         target.WorkflowState = source;
-        
-        if (source.Properties.TryGetValue<string>(SetName.WorkflowInstanceNameKey, out var name))
```
rolandgriesser, line 48: @sfmskywalker Is there a specific reason why you removed that part? Or at least didn't use it as a fallback?
All my old workflow instances used the `SetName` activity to set the name, and so each call to the `Map` function will set the `Name` to null as it isn't considering the old value but the new value was never set.

sfmskywalker: The WorkflowState should already store the name, which is why I removed this line, but I did not consider the old value. I will revert this change and add a comment explaining the need to keep this. Thank you for raising this issue.

sfmskywalker: https://github.com/elsa-workflows/elsa-core/issues/6748

sfmskywalker: Fixed via https://github.com/elsa-workflows/elsa-core/pull/6767

PR #6467, opened 2025-03-04
Harmonized launchSettings to be inline with appSettings

MatthijsKrempel:
In the appSettings for the ServerAndStudio.Web project the baseUrl is configured as 8080.
This PR will align the launchSettings to be inline with that configuration so that it does not throw errors on startup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6467)


sfmskywalker (approved):
Cheers @MatthijsKrempel !

PR #6477, opened 2025-03-08
Catch exceptions and log instead of crashing

### LINKED ISSUE #6473: Cleanup job deserialization failure causes Elsa 3.0 to CrashLoop
## Description
I have an Elsa 3.0 instance that's running that has a cleanup job that failed to deserialize one of the Workflow Instances.  Because it can't deserialize the data, it is unable to complete the cleanup and crashes Elsa in a crash loop.

## Steps to Reproduce
You can cause this issue by making any invalid JSON change to the `data` in a WorkflowInstance table in the database. I had it happen when a workflow faulted because of how an exception got stored but you can cause it any way you'd like.

Make sure the WorkflowInstance is one that's old enough that it will get cleaned up by the Cleanup job.

Restart Elsa and it will crash loop.

## Expected Behavior
It should be able to cleanup workflow instances without having to successfully deserialize the data in them.

## Actual Behavior
```
Unhandled exception. System.Text.Json.JsonException: The JSON value could not be converted to System.Collections.Generic.ICollection`1[Elsa.Workflows.Memory.Variable]. Path: $.activityExecutionContexts.$values[1].dynamicVariables.$values | LineNumber: 0 | BytePositionInLine: 1107.
   at Elsa.Workflows.Serialization.ReferenceHandlers.CustomPreserveReferenceResolver.AddReference(String referenceId, Object value)
   at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, TCollection& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.Metadata.JsonPropertyInfo`1.ReadJsonAndSetMember(Object obj, ReadStack& state, Utf8JsonReader& reader)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, TCollection& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.Metadata.JsonPropertyInfo`1.ReadJsonAndSetMember(Object obj, ReadStack& state, Utf8JsonReader& reader)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.Deserialize(Utf8JsonReader& reader, ReadStack& state)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 utf8Json, JsonTypeInfo`1 jsonTypeInfo, Nullable`1 actualByteCount)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 json, JsonTypeInfo`1 jsonTypeInfo)
   at Elsa.Workflows.Serialization.Serializers.JsonWorkflowStateSerializer.Deserialize(String serializedState)
   at Elsa.EntityFrameworkCore.Modules.Management.EFCoreWorkflowInstanceStore.OnLoadAsync(ManagementElsaDbContext managementElsaDbContext, WorkflowInstance entity, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Store`2.QueryAsync(Func`2 query, Func`4 onLoading, Boolean ignoreQueryFilters, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Store`2.QueryAsync(Func`2 query, Func`4 onLoading, Boolean ignoreQueryFilters, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Store`2.QueryAsync(Func`2 query, Func`4 onLoading, CancellationToken cancellationToken)
   at Open.Linq.AsyncExtensions.Extensions.ToList[TSource](Task`1 source)
   at Elsa.EntityFrameworkCore.Modules.Management.EFCoreWorkflowInstanceStore.FindManyAsync[TOrderBy](WorkflowInstanceFilter filter, PageArgs pageArgs, WorkflowInstanceOrder`1 order, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Modules.Management.EFCoreWorkflowInstanceStore.FindManyAsync(WorkflowInstanceFilter filter, PageArgs pageArgs, CancellationToken cancellationToken)
   at Elsa.Retention.Jobs.CleanupJob.ExecuteAsync(CancellationToken cancellationToken)
   at Elsa.Retention.CleanupRecurringTask.ExecuteAsync(CancellationToken stoppingToken)
   at Elsa.Common.Multitenancy.TaskExecutor.ExecuteInternalAsync(ITask task, Func`1 action, CancellationToken cancellationToken)
   at Elsa.Common.Multitenancy.TaskExecutor.ExecuteInternalAsync(ITask task, Func`1 action, CancellationToken cancellationToken)
   at Elsa.Common.Multitenancy.TaskExecutor.ExecuteTaskAsync(ITask task, CancellationToken cancellationToken)
   at Elsa.Common.Multitenancy.EventHandlers.StartRecurringTasks.<>c__DisplayClass5_1.<<TenantActivatedAsync>b__0>d.MoveNext()
```

yinzara:
Modified both WorkflowInstance and WorkflowDefinition state loading to prevent throwing exceptions if the state fails to load successfully.

Also modified the CleanupJob to catch exceptions thrown by cleaning up and logging as errors and continuing instead of crashing.

Fixes #6473

### DIFF
--- a/src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
+++ b/src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
@@ -13,12 +13,13 @@ using Elsa.Common.Entities;
 using Elsa.Workflows;
 using Elsa.Workflows.Management;
 using JetBrains.Annotations;
+using Microsoft.Extensions.Logging;
 
 namespace Elsa.EntityFrameworkCore.Modules.Management;
 
 /// <inheritdoc />
 [UsedImplicitly]
-public class EFCoreWorkflowDefinitionStore(EntityStore<ManagementElsaDbContext, WorkflowDefinition> store, IPayloadSerializer payloadSerializer)
+public class EFCoreWorkflowDefinitionStore(EntityStore<ManagementElsaDbContext, WorkflowDefinition> store, IPayloadSerializer payloadSerializer, ILogger<EFCoreWorkflowDefinitionStore> logger)
     : IWorkflowDefinitionStore
 {
     /// <inheritdoc />
@@ -178,8 +179,15 @@ public class EFCoreWorkflowDefinitionStore(EntityStore<ManagementElsaDbContext, 
         var data = new WorkflowDefinitionState(entity.Options, entity.Variables, entity.Inputs, entity.Outputs, entity.Outcomes, entity.CustomProperties);
         var json = (string?)managementElsaDbContext.Entry(entity).Property("Data").CurrentValue;
 
-        if (!string.IsNullOrWhiteSpace(json))
-            data = payloadSerializer.Deserialize<WorkflowDefinitionState>(json);
+        try
+        {
+            if (!string.IsNullOrWhiteSpace(json))
+                data = payloadSerializer.Deserialize<WorkflowDefinitionState>(json);
+        }
+        catch (Exception exp)
+        {
+            logger.LogWarning(exp, "Could not deserialize workflow definition state: {DefinitionId}. Reverting to default state", entity!.DefinitionId);
+        }
 
         entity.Options = data.Options;
         entity.Variables = data.Variables;
--- a/src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowInstanceStore.cs
+++ b/src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowInstanceStore.cs
@@ -11,6 +11,7 @@ using Elsa.Workflows.Management.Models;
 using Elsa.Workflows.Management.Options;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Open.Linq.AsyncExtensions;
 
@@ -26,6 +27,7 @@ public class EFCoreWorkflowInstanceStore : IWorkflowInstanceStore
     private readonly IWorkflowStateSerializer _workflowStateSerializer;
     private readonly ICompressionCodecResolver _compressionCodecResolver;
     private readonly IOptions<ManagementOptions> _options;
+    private readonly ILogger<EFCoreWorkflowInstanceStore> _logger;
 
     /// <summary>
     /// Constructor.
@@ -34,12 +36,14 @@ public class EFCoreWorkflowInstanceStore : IWorkflowInstanceStore
         EntityStore<ManagementElsaDbContext, WorkflowInstance> store,
         IWorkflowStateSerializer workflowStateSerializer,
         ICompressionCodecResolver compressionCodecResolver,
-        IOptions<ManagementOptions> options)
+        IOptions<ManagementOptions> options,
+        ILogger<EFCoreWorkflowInstanceStore> logger)
     {
         _store = store;
         _workflowStateSerializer = workflowStateSerializer;
         _compressionCodecResolver = compressionCodecResolver;
         _options = options;
+        _logger = logger;
     }
 
     /// <inheritdoc />
@@ -198,12 +202,18 @@ public class EFCoreWorkflowInstanceStore : IWorkflowInstanceStore
         var compressionAlgorithm = (string?)managementElsaDbContext.Entry(entity).Property("DataCompressionAlgorithm").CurrentValue ?? nameof(None);
         var compressionStrategy = _compressionCodecResolver.Resolve(compressionAlgorithm);
 
-        if (!string.IsNullOrWhiteSpace(json))
+        try
         {
-            json = await compressionStrategy.DecompressAsync(json, cancellationToken);
-            data = _workflowStateSerializer.Deserialize(json);
+            if (!string.IsNullOrWhiteSpace(json))
+            {
+                json = await compressionStrategy.DecompressAsync(json, cancellationToken);
+                data = _workflowStateSerializer.Deserialize(json);
+            }
         }
-
+        catch (Exception exp)
+        {
+            _logger.LogWarning(exp, "Exception while deserializing workflow instance state: {InstanceId}. Reverting to default state", entity.Id);
+        }
         entity.WorkflowState = data;
     }
 
--- a/src/modules/Elsa.Retention/Jobs/CleanupJob.cs
+++ b/src/modules/Elsa.Retention/Jobs/CleanupJob.cs
@@ -71,7 +71,14 @@ public class CleanupJob(
 
                     await foreach (var entities in collector.GetRelatedEntitiesGeneric(page.Items).WithCancellation(cancellationToken))
                     {
-                        await cleanupService.Cleanup(entities);
+                        try
+                        {
+                            await cleanupService.Cleanup(entities);
+                        }
+                        catch (Exception exp)
+                        {
+                            _logger.LogError(exp, "Failed to clean up {Type} because exception thrown", collectorService.Key.Name);
+                        }
                     }
                 }
                 


src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
```diff
@@ -178,8 +179,15 @@ private ValueTask OnLoadAsync(ManagementElsaDbContext managementElsaDbContext, W
...
-        if (!string.IsNullOrWhiteSpace(json))
-            data = payloadSerializer.Deserialize<WorkflowDefinitionState>(json);
+        try
+        {
+            if (!string.IsNullOrWhiteSpace(json))
+                data = payloadSerializer.Deserialize<WorkflowDefinitionState>(json);
+        }
+        catch (Exception exp)
+        {
+            logger.LogWarning(exp, "Could not deserialize workflow definition state: {DefinitionId}. Reverting to default state", entity!.DefinitionId);
```
sfmskywalker, line 189: I'm wondering if it might be better to throw the exception instead of swallowing it here - unless it breaks the application and not just the workflow - is that what's happening when an error occurs here?

af-git-64: Hello, we're experiencing the same issue, and I can confirm that it completely breaks the application. Resolving this should be a top priority, as the retention module is unusable without it—especially when the database contains old instances from previous versions that cannot be deserialized.

sfmskywalker: Does that require exceptions to be caught here, or only in the retention clean up job?

yinzara: While I agree catching it within the retention module would solve the problem outlined in the initial issue, I think that catching these exceptions in the stores is also appropriate and solves other defects.  I regularly get issues where I get a 400 error while loading workflows because of some small serialization issue.  The only place we can prevent a bad serialization from crashing an entire request is by catching and warning (not throwing) in the `OnLoadAsync` functions.

src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowInstanceStore.cs
```diff
@@ -198,12 +202,18 @@ private async ValueTask OnLoadAsync(ManagementElsaDbContext managementElsaDbCont
...
 
-        if (!string.IsNullOrWhiteSpace(json))
+        try
         {
-            json = await compressionStrategy.DecompressAsync(json, cancellationToken);
-            data = _workflowStateSerializer.Deserialize(json);
+            if (!string.IsNullOrWhiteSpace(json))
+            {
+                json = await compressionStrategy.DecompressAsync(json, cancellationToken);
+                data = _workflowStateSerializer.Deserialize(json);
```
yinzara, line 210: This is actually the only really necessary fix to make it so that the Cleanup script can delete WorkflowInstance(s) that have bad data in them.  Even the try/catch below in the CleanupJob only prevents the CleanupJob from crashing the entire Elsa server, but it won't allow it to cleanup bad instances without this fix.

src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
yinzara: Technically the fix here with the WorkflowDefinition store is not strictly necessary to fix the bug outlined in #6473, only the fix below in the WorkflowInstanceStore is required.  This is really a "nice to have" :-)

src/modules/Elsa.Retention/Jobs/CleanupJob.cs
```diff
@@ -71,7 +71,14 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
...
                     await foreach (var entities in collector.GetRelatedEntitiesGeneric(page.Items).WithCancellation(cancellationToken))
                     {
-                        await cleanupService.Cleanup(entities);
+                        try
+                        {
+                            await cleanupService.Cleanup(entities);
+                        }
+                        catch (Exception exp)
+                        {
+                            _logger.LogError(exp, "Failed to clean up {Type} because exception thrown", collectorService.Key.Name);
```
yinzara, line 80: This fix only prevents the CleanupJob from crashing the entire Elsa server should the Cleanup service throw an exception.  However it does not fix the problem where a bad WorkflowIhstance (with undeserializable data) cannot be deleted. The above fix in WorkflowInstanceStore is required for that.

sfmskywalker:
Let's catch the exception in the CleanupJon to prevent the entire host from crashing, but let's revisit handling serialization exceptions of workflow definitions separately. My concern is that if we catch and swallow the exceptions, even though we log these, it might lead to issues that are too subtle to notice. It would be better to see the current user experience, and design the right solution from there.

src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowDefinitionStore.cs
sfmskywalker: > I believe it's sufficient to catch the exception within the retention module. Catching it in the definition store or instance store could introduce unintended side effects.
> 
> The exception should be caught at the point where the workflow instances are retrieved: `var page = await workflowInstanceStore.FindManyAsync(filter, pageArgs, cancellationToken); `
> 
> However, this approach is not ideal either, as it disregards all instances retrieved on that page, even if only a single instance is corrupted. Ideally, exceptions should be caught individually for each retrieved instance to prevent one corrupted instance from affecting the others. However, this would introduce additional overhead.

Yeah exactly. We want to be able to load multiple workflow instances in bulk. However, once they are received from the DB, the EF Core implementation of the workflow instance store then goes through each record individually to deserialize the workflow state, so here we could certainly implement an exception handler.

This would allow bulk actions like cancel and export to succeed for non-problematic workflow instances at least. However, we still need a way to make clear to the user that there are problems with some of the workflow instances.

src/modules/Elsa.EntityFrameworkCore/Modules/Management/WorkflowInstanceStore.cs
sfmskywalker: True, so maybe let's add an exception handler here. If an error occurred here, it wouldn't make the workflow instance useful anyway, so we might just as well handle the exception for now, swallow it, and log an error. 
Then the CleanupJob and other bulk operations can still continue normally.

sfmskywalker (approved):
After another round of review and testing, I think we should take this PR as is. The only change I made was changing the log level from warning to error in case deserialization fails during workflow definition load.

Issue #6479, opened 2025-03-09
Activities can run multiple times when using FlowJoin(WaitAny) with backward connection


bobhauser:
## Description
Although the changes in https://github.com/elsa-workflows/elsa-core/pull/6406 fixed many issues related to flowchart activity scheduling, there is a subtle case where some activities were executed multiple times when they should have only been executed once.

The case involves a FlowSwitch, a FlowJoin(WaitAny), and a backward connection. Because FlowScope only keeps track of whether the last connection visit was followed it is possible that `FlowScope.GetConnectionLastVisitFollowed` returns the result of the previous iteration through the loop. So if the last iteration through the loop was followed it may schedule the FlowJoin again activity even though the current iteration though the loop is NOT followed. 

## Steps to Reproduce

The following unit test will reproduce the issue

```
    public async Task WaitLoopBugTest(FlowJoinMode joinMode)
    {
        var workflow = new TestWorkflow(workflowBuilder =>
        {
            var loopVariable = new Variable<int>("LoopCount", 0);

            var start = new Start();
            var loopbackSwitch = new FlowSwitch()
            {
                Cases = {
                    new FlowSwitchCase("DoLoopback", new Expression("JavaScript", "getVariable('LoopCount') < 3")),
                },
                Mode = new(SwitchMode.MatchFirst)
            };
            var a = new WriteLine("A");
            var incrementLoop = new SetVariable()
            {
                Variable = loopVariable,
                Value = new Models.Input<object?>(new Expression("JavaScript", "getVariable('LoopCount') + 1"))
            };
            var join = new FlowJoin()
            {
                Mode = new(joinMode)
            };
            var b = new WriteLine("B");
            var end = new End();


            workflowBuilder.Root = new Flowchart
            {
                Variables =
                    {
                        loopVariable
                    },
                Activities =
                    {
                        start,
                        loopbackSwitch,
                        a,
                        incrementLoop,
                        join,
                        b,
                        end
                    },
                Connections =
                {
                    new(start, loopbackSwitch),
                    new(new Endpoint(loopbackSwitch, "DoLoopback"), new Endpoint(a)),
                    new(new Endpoint(loopbackSwitch, "DoLoopback"), new Endpoint(incrementLoop)),
                    new(new Endpoint(loopbackSwitch, "Default"), new Endpoint(b)),
                    new(a, join),
                    new(incrementLoop, join),
                    new(join, loopbackSwitch),
                    new(b, end),
                }
            };
        });

[...]

PR #6480, opened 2025-03-09
Fixes issue with FlowJoin(WaitAny) where activities can be executed multiple times

bobhauser:
This fixes an issue where activities are executed more times than expected. See https://github.com/elsa-workflows/elsa-core/issues/6479 for more details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6480)


sfmskywalker (approved):
Thanks @bobhauser 👍🏻

PR #6491, opened 2025-03-12
Correct the issue with DeleteAsync not executing due to a premature p…

Green7:
The DeleteAsync method was terminating prematurely because the presence of query parameters was checked before filters were applied. As a result, the deletion queries never executed since the query parameters were always empty before applying the filters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6491)


Green7:
@dotnet-policy-service agree

sfmskywalker (approved):

sfmskywalker:
Nice catch @Green7 !

PR #6498, opened 2025-03-14
Allow to configure JobRunner worker count

truthz03:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6498)

### DIFF
--- a/src/common/Elsa.Mediator/HostedServices/JobRunnerHostedService.cs
+++ b/src/common/Elsa.Mediator/HostedServices/JobRunnerHostedService.cs
@@ -1,6 +1,8 @@
 using Elsa.Mediator.Contracts;
+using Elsa.Mediator.Options;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Elsa.Mediator.HostedServices;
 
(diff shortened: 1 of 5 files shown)


src/common/Elsa.Mediator/HostedServices/JobRunnerHostedService.cs
```diff
@@ -9,23 +9,24 @@ namespace Elsa.Mediator.HostedServices;
...
 {
-    private const int WorkerCount = 4;
+    private readonly int _workerCount;
     private readonly IJobsChannel _jobsChannel;
     private readonly ILogger<JobRunnerHostedService> _logger;
 
     /// <inheritdoc />
-    public JobRunnerHostedService(IJobsChannel jobsChannel, ILogger<JobRunnerHostedService> logger)
-    {
+    public JobRunnerHostedService(int workerCount, IJobsChannel jobsChannel, ILogger<JobRunnerHostedService> logger)
```
sfmskywalker, line 17: Would it perhaps be simpler if we just inject `IOptions<MediatorOptions>`? That way, DI registration would not have to change.

truthz03: Yes. That will be also ok. I did it like that because the other services with workers was implemented this way.

If you want I can change it to use IOptions but then I will also change the other 2 services to make them all equal :)

sfmskywalker: Ah, yes I see. You did the right thing, but if you agree that it's simpler to just inject the IOptions thing, then it would be great of you could apply this to the other 2 services as well. Thank you :)

truthz03: @sfmskywalker 
Ok. I changed it now

sfmskywalker (approved):

PR #6502, opened 2025-03-15
Improve OTEL Spans and Tags for Workflow and Activity Execution  

sfmskywalker:
This PR enhances OpenTelemetry (OTEL) spans and tags to align with best practices.  

#### **Operations**  
- Updated workflow and activity execution operation names and kinds:  
  - **Workflow Execution**:  
    - **Operation Name**: `elsa.workflow.execution`  
    - **Operation Kind**: `Server`  
  - **Activity Execution**:  
    - **Operation Name**: `elsa.activity.execution`  
    - **Operation Kind**: `Internal`  

#### **Resources**  
- Workflow and activity execution resources are now defined individually instead of being grouped together.  
- Resource names now follow an action/verb pattern similar to HTTP requests, e.g.:  
  - **Workflow Execution**: `execute workflow HelloWorld`  
  - **Activity Execution**: `execute activity WriteLine`  

#### **Error Handling**  
- Errors are now recorded using OTEL Events with the standard `"exception"` event and tag instead of being stored at the span level.  

#### **Trace Decoupling**  
- Introduced a new setting, `StartNewTrace`, for the `DispatchWorkflow` and `BulkDispatchWorkflows` activities.  
- When enabled, this setting starts a new root trace for dispatched workflows, reducing the hierarchy of spans in Datadog.  
- The new trace is linked to the previous one to maintain a connection between workflows without creating deep nested traces.  

This update improves observability, trace management, and adherence to OTEL best practices.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6502)


src/modules/Elsa.Workflows.Runtime/Activities/BulkDispatchWorkflows.cs
```diff
@@ -78,6 +78,12 @@ public BulkDispatchWorkflows([CallerFilePath] string? source = default, [CallerL
         Description = "Wait for the dispatched workflows to complete before completing this activity.",
         DefaultValue = true)]
     public Input<bool> WaitForCompletion { get; set; } = new(true);
+    
+    /// <summary>
+    /// Indicates whether a new trace context should be started for the workflow execution.
+    /// </summary>
+    [Input(Description = "Start a new trace context when using Open Telemetry.", Category = "Open Telemetry")]
+    public Input<bool> StartNewTrace { get; set; }
```
raymonddenhaan, line 86: I don't see the added benefit of making this a setting. This will only make things more complex while not adding that much benefit if any, but giving the possibility to make checking traces nearly impossible. If you still want this in, please set the default value to true since that is the setting that is needed to prevent too many traces being combined to a single trace.

sfmskywalker: The benefit would be to have a full picture in a single trace, which I think should be the default. Only when that trace starts to become unwieldy due to many nested executions, would it be good to opt out.

raymonddenhaan (approved):

PR #6503, opened 2025-03-16
FEAT - Automatic SQL Expression Parameterization

### LINKED ISSUE #6500: Adding support for SQL Parameters
## Feature Request

Adding support for SQL Parameters in the new SQL Activity, this will prevent SQL injection in case the parameters where supplied by a user.
It will be a good idea to use UI Component similar to Switch Case but modify it to support SQL Parameters.

KnibbsyMan:
As raised by #6500, variables could originate from outside the Studio (from an external source) and therefore could expose the SQL queries to SQL Injection. To address this, the following has been implemented & changed:

- Variables are now inserted via `{{ }}` rather than `@`. I.e. `{{Variables.SomeVariable}}` or `{{Input.SomeInput}}`. This simplifies the SqlEvaluator logic and removes the issue with `@` being used within the query itself.
- Inserted variables/inputs/etc are automatically parameterised to prevent SQL Injection. No front end user code required.
- The parameter string format used in parameterising can be overridden in the base classes should another client need access to this, i.e. potentially oracle. Default configuration is `@p0`.
- Simplified client implementations to reduce duplication of code.
- Tested implementation against all four available clients - SqlServer, Sqlite, MySql & PostgreSQL.

This PR closes #6500 - Thanks for a great suggestion @ghazisarhan!

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6503)

### DIFF
--- a/src/modules/Elsa.Sql.MySql/MySqlClient.cs
+++ b/src/modules/Elsa.Sql.MySql/MySqlClient.cs
@@ -1,74 +1,16 @@
 ﻿using MySql.Data.MySqlClient;
 using Elsa.Sql.Client;
-using System.Data;
-using Elsa.Sql.Models;
+using System.Data.Common;
 
 namespace Elsa.Sql.MySql;
 
-public class MySqlClient : BaseSqlClient, ISqlClient
+/// <summary>
+/// MySql client implementation.
+/// </summary>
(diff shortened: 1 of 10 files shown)


sfmskywalker (approved):
Awesome 👍🏻

PR #6531, opened 2025-03-26
Fix DbContext pooling

sfmskywalker:
Fixed DbContext pooling support.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6531)

### DIFF
--- a/src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
+++ b/src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
@@ -56,13 +56,8 @@ public abstract class PersistenceFeatureBase<TFeature, TDbContext> : FeatureBase
 
         Action<IServiceProvider, DbContextOptionsBuilder> setup = (sp, opts) =>
         {
-            opts.EnableSensitiveDataLogging();
-
-#if NET9_0_OR_GREATER
             opts.ConfigureWarnings(w => w.Ignore(RelationalEventId.PendingModelChangesWarning));
-            
             DbContextOptionsBuilder(sp, opts);
-#endif
         };
 
         if (UseContextPooling)
--- a/src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
+++ b/src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
@@ -4,6 +4,7 @@ using Elsa.Features.Abstractions;
 using Elsa.Features.Attributes;
 using Elsa.Features.Services;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Elsa.EntityFrameworkCore;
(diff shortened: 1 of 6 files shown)


src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
```diff
@@ -50,13 +51,24 @@ public override void ConfigureHostedServices()
...
     public override void Apply()
     {
-        if(DbContextOptionsBuilder == null)
+        if (DbContextOptionsBuilder == null)
             throw new InvalidOperationException("The DbContextOptionsBuilder must be configured.");
-        
+
+        Action<IServiceProvider, DbContextOptionsBuilder> setup = (sp, opts) =>
+        {
+            opts.EnableSensitiveDataLogging();
```
raymonddenhaan, line 59: Do we want this enabled by default? If so should we not be able to disable this?

src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
```diff
@@ -50,13 +51,24 @@ public override void ConfigureHostedServices()
...
+
+#if NET9_0_OR_GREATER
+            opts.ConfigureWarnings(w => w.Ignore(RelationalEventId.PendingModelChangesWarning));
+            
+            DbContextOptionsBuilder(sp, opts);
```
raymonddenhaan, line 64: Should this not be active for dotnet8? if not we can make this entire setup action dotnet9 only

src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
sfmskywalker: It's a good question, I was wondering the same thing. I haven't changed it because it was already like this and the scope of this PR is to get DB Context Pooling working again. Maybe let's keep this PR focused on that, and we create another PR for exposing this setting through the Feature class, or perhaps as part of an existing or new option class.

src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
sfmskywalker: Maybe, I don't know. It was taken as-is and moved into this location. Let's consider this separate from this PR.

sfmskywalker: Actually, I remember now. The reason this was added is because the behavior changed with the introduction of EF Core 9. So, it probably doesn't hurt doing this for .NET 8, which would certainly look nicer in terms of code.

raymonddenhaan (approved):

src/modules/Elsa.EntityFrameworkCore.Common/PersistenceFeatureBase.cs
sfmskywalker: On second thought, let's just remove this setting altogether - if we want to enable it, we already have access to the `DbContextOptionsBuilder` delegate anyway, from where we can not only control this aspect, but every other aspect as well. In reality, sensitive logging should be enabled only for troubleshooting.

Thanks for calling it out, because having this enabled by default might do more harm than good.

PR #6564, opened 2025-04-07
2.x Fix array enumeration in JsonFilter

Slettan:
Use Enumerate(TemplateContext) instead of Enumerate() (which is obsolete and returns an empty array)

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6564)


Slettan:
@dotnet-policy-service agree

sfmskywalker (approved):

Issue #6591, opened 2025-04-15
3.4.0-rc1 FlowJoin+WaitAny  unexpected behavior


Nordran:
## Description
Unexpected behavior of FlowJoin+WaitAny activity in cyclic workflow after first cycle.

## Steps to Reproduce
1. Start attached workflow by get request to /joins/start
2. Receive response. Click 1
3 Recieve response. Click 2
4 No response recieved

Execution of flow stoped since FlowJoin wasn't activated.

[test-join (1).json](https://github.com/user-attachments/files/19761077/test-join.1.json)

![Image](https://github.com/user-attachments/assets/28625361-4ce5-45e1-994e-af1e674b22a3)

## Expected Behavior
FlowJoin has been activated and we continue to another cycle

## Actual Behavior
FlowJoin hasn't been started second time. 

## Environment
3.4.0-rc1

## Additional Context
That happens because canceled activities and their outbound connections are not counted as visited, so FlowScope.ShouldIgnoreConnection returns true and FlowChart.ScheduleJoinActivityAsync doesn't schedule FlowJoin.

PR #6592, opened 2025-04-15
Fix unexpected FlowJoin(WaitAny) behavior in cyclic flow.

Nordran:
Closes https://github.com/elsa-workflows/elsa-core/issues/6591
Visited counting mechanism now applies to canceled branches.
Additional changes were done to ScheduleJoinActivityAsync to avoid repeating scheduling of FlowJoin by counting mechanism propagation

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6592)


Nordran:
@dotnet-policy-service agree

Nordran:
@sfmskywalker i've noticed that my changes to ScheduleJoinActivityAsync  broke FlowJoin WaitAll scenario since FlowJoin remains in Running state after first call and other branches are not able to call it anymore. With last commit FlowJoin(WaitAll) will remain in pending state until it's ready to be completed

sfmskywalker:
Got it. Thanks for catching that :)

sagargolu:
@Nordran & @sfmskywalker 

I have tried to fix this by putting WaitAll condition to check. in If condition, this might solve that problem of WaitAll in below code.

 if (joinContext is not { Status: ActivityStatus.Running })
        {}
        
        if ((joinContext != null && joinContext.ActivityState.ContainsKey("Mode") && joinContext.ActivityState["Mode"].ToString() == "WaitAll") || joinContext is not { Status: ActivityStatus.Running })
{}

PR #6606, opened 2025-04-22
Start a new workflow trace when triggered from UI

raymonddenhaan:
Adds an option to use a dummy activity as a hack to prevent other middleware of reinstating the parent activity

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6606)

### DIFF
--- a/src/modules/Elsa.OpenTelemetry/Features/OpenTelemetryFeature.cs
+++ b/src/modules/Elsa.OpenTelemetry/Features/OpenTelemetryFeature.cs
@@ -2,16 +2,34 @@ using Elsa.Features.Abstractions;
 using Elsa.Features.Services;
 using Elsa.OpenTelemetry.Contracts;
 using Elsa.OpenTelemetry.Handlers;
+using Elsa.OpenTelemetry.Options;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Elsa.OpenTelemetry.Features;
 
 public class OpenTelemetryFeature(IModule module) : FeatureBase(module)
 {
+    /// <summary>
+    /// When active this will create a new root Activity if the current Activity has a remote parent.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


src/modules/Elsa.OpenTelemetry/Features/OpenTelemetryFeature.cs
```diff
@@ -2,16 +2,34 @@
...
+    /// <summary>
+    /// When active this will create a new root Activity if the current Activity has a remote parent.
+    /// </summary>
+    public bool UseNewRootActivityForRemoteParent { get; set; } = true;
+    
+    /// <summary>
+    /// Determines if instead of a empty parent, a dummy parent Activity should be used to create a new Root Activity.
+    /// </summary>
+    /// <remarks>This is needed when middleware is active that uses the previous Parent Activity despite creating an Activity based on an empty Parent.</remarks>
+    public bool UseDummyParentActivityAsRootSpan { get; set; } = false;
```
sfmskywalker, line 21: Should we include a comment that these settings exist for the purpose of mitigating a limitation with the datadog autoinstrumentation?

sfmskywalker (approved):

src/modules/Elsa.OpenTelemetry/Features/OpenTelemetryFeature.cs
raymonddenhaan: I didn't mention Datadog directly, however I tried to use the message in the remarks tag to get this point across. Is there anything you are missing in this?

sfmskywalker: Unless I misunderstood, this effect is directly tied to what Datadog's auto instrumentation is doing, so for me it would be helpful to have that context.

PR #6637, opened 2025-05-09
Add IResilienceStrategy Abstraction with Category Matching and Expression-Based Configuration

sfmskywalker:
This PR introduces a new `IResilienceStrategy` abstraction to enable configurable resilience strategies across different activity types in Elsa Workflows. The strategy selection supports category-based matching, making it possible to associate strategies with specific types of activities.

In addition to selecting a strategy from a predefined list, users can switch to Expression mode to dynamically define the strategy using JavaScript expressions. Examples include:

* Returning a strategy ID:

  ```javascript
  return "123"; // Return the ID of a registered strategy.
  ```

* Returning a strategy as a plain object:

  ```javascript
  return {
      "$type": "HttpResilienceStrategy",
      "Id": "3",
      "DisplayName": "Retry 5 times",
      "RetryCount": 5,
      "BackoffFactor": 3.0
  }
  ```

* Instantiating a strategy directly in code:

  ```javascript
  const strategy = new HttpResilienceStrategy();
  strategy.RetryCount = 10;
  strategy.BackoffFactor = 3.0;
  return strategy;
  ```

Out of the box, a `ConfigurationResilienceStrategySource` implementation is included. This allows defining strategies via configuration files like `appsettings.json`:

```json
{
  "Resilience": {
    "Strategies": [
      {
        "$type": "HttpResilienceStrategy",
        "Id": "1",
        "DisplayName": "Retry 3 times",
        "RetryCount": 3,
        "BackoffFactor": 2.0
      },
      {
        "$type": "HttpResilienceStrategy",
        "Id": "2",
        "DisplayName": "Retry 10 times",
        "RetryCount": 10,
        "BackoffFactor": 2.0
      }
    ]
  }
}
```

These features provide a highly flexible mechanism to define and apply custom resilience behavior in workflows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6637)


src/modules/Elsa.Http/Activities/SendHttpRequestBase.cs
```diff
@@ -180,15 +182,17 @@ private async Task TrySendAsync(ActivityExecutionContext context)
 
         return;
 
-        async Task<HttpResponseMessage> SendRequestAsync()
+        async Task<HttpResponseMessage> SendRequestAsync(ActivityExecutionContext activityExecutionContext)
         {
+            // Keep this for backward compatibility.
```
raymonddenhaan, line 187: I agree we should keep this in, but should this take precedence over a selected resilience strategy?

src/modules/Elsa.Resilience.Core/Attributes/ResilienceCategoryAttribute.cs
```diff
@@ -0,0 +1,7 @@
+namespace Elsa.Resilience;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Property)]
```
raymonddenhaan, line 3: Why allow a property to be targeted? Shouldn't this only be targeted at Activities?

src/modules/Elsa.Resilience.Core/Services/ResilienceStrategyCatalog.cs
```diff
@@ -0,0 +1,31 @@
...
+    public async Task<IResilienceStrategy?> GetAsync(string id, CancellationToken cancellationToken = default)
+    {
+        var strategies = await ListAsync(cancellationToken);
+        return strategies.FirstOrDefault(x => x.Id == id);
+    }
+
+    private async Task<IEnumerable<IResilienceStrategy>> GetStrategiesInternalAsync()
+    {
+        var strategies = new List<IResilienceStrategy>();
+        foreach (var provider in _providers) strategies.AddRange(await provider.GetStrategiesAsync());
```
raymonddenhaan, line 28: How do we want to handle Strategies with the same ID or name for from either the same or different providers?

src/modules/Elsa.Resilience.Core/Contracts/IResilientActivityInvoker.cs
```diff
@@ -0,0 +1,8 @@
+using Elsa.Workflows;
+
+namespace Elsa.Resilience;
+
+public interface IResilientActivityInvoker
+{
+    Task<T> InvokeAsync<T>(IResilientActivity activity, ActivityExecutionContext context, Func<Task<T>> action, CancellationToken cancellationToken = default);
```
raymonddenhaan, line 7: Might be good to add a comment here that when no Strategy can be found the action will be executed without one.

raymonddenhaan (approved):

src/modules/Elsa.Common/Extensions/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        return JsonSerializer.Serialize(element, options);
+    }
+
+    private static JsonElement ConvertToJsonElement(IConfigurationSection section)
+    {
+        using var doc = JsonDocument.Parse(JsonSerializer.Serialize(ToObject(section)));
+        return doc.RootElement.Clone();
+    }
+
+    private static object? ToObject(IConfigurationSection section)
```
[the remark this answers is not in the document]
sfmskywalker: Actually, ToObject can return any of the following:

- A string
- A dictionary
- A list of strings or dictionaries

src/modules/Elsa.Http/Activities/SendHttpRequestBase.cs
```diff
@@ -180,15 +182,17 @@ private async Task TrySendAsync(ActivityExecutionContext context)
...
             if (resiliencyEnabled)
```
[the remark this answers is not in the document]
sfmskywalker: Yes good idea.

src/modules/Elsa.Http/Activities/SendHttpRequestBase.cs
sfmskywalker: TLDR; not worth the trouble.

Not necessarily, but if we were to change this, it seems to me that the code becomes a bit hairy: we'd have to explicitly try and resolve the active strategy to use, and, if not null, execute using this strategy. This code already lives nicely encapsulated in the ResilientActivityInvoker, so I'd rather not copy that out just for a setting that is really only currently used by users of 3.4.0 preview. 

However, if you see a clean way to do it, then yes by all means we should reverse the precedence.

src/modules/Elsa.Resilience.Core/Attributes/ResilienceCategoryAttribute.cs
sfmskywalker: Originally as per the technical design, I wanted to expose a ResilienceStrategyId property on resilient activities, and then bind that property's category to the strategy types matching the same category using this attribute.

But with the current implementation, no such property is used, so we actually don't need to target properties anymore.

I'll remove it 👍🏻

src/modules/Elsa.Resilience.Core/Contracts/IResilientActivityInvoker.cs
sfmskywalker: Agreed.

src/modules/Elsa.Resilience.Core/Services/ResilienceStrategyCatalog.cs
sfmskywalker: Nice catch. I think that either we leave this up to the user, or we prefix IDs with e.g. the type name of the resilience strategy source, hashed or otherwise. Tricky part with that might be that we must ensure the type names never get renamed, or else.

Another idea might be to use an attribute that uniquely identifies the source type. Obviously, this still doesn't guarantee uniqueness if two independent developers happen to pick the same identifier, but given those small odds, at least we could throw an exception.

Finally, if we did want to ensure developers won't have to deal with each other in this manner, we could put the burden with the application developer and require them to manually register the sources, where they should then also provide some sort of prefix. I'm not a fan of this because it makes it more complicated for the application developer.

What do you think might be the best solution?

sfmskywalker: I opted for the second approach where we can control the prefix of the identities using a attribute. This prevents us from renaming types without realizing we'd be breaking people's workflows without requiring application developers to explicitly provide a prefix or having to be concerned with unique identities across strategy sources.

Let me know what you think.

PR #6643, opened 2025-05-12
Enhances workflow runtime resilience and signaling

### LINKED ISSUE #6642: BulkDispatchWorkflows May Not Complete Due to Bookmark Queue Deadlock and Swallowed Exceptions
There is a bug causing workflows using `BulkDispatchWorkflows` to intermittently not complete. This occurs under the following conditions:

* A parent workflow dispatches multiple child workflows using `BulkDispatchWorkflows`.
* The `ChildCompleted` activity is set.
* `BulkDispatchWorkflows` is configured to wait until all child workflows have completed.

**Observed Behavior:**
Sometimes the `BulkDispatchWorkflows` activity never completes, preventing the parent workflow from progressing.

**Root Causes Identified:**

1. **Deadlock in Bookmark Queue Processor:**

   * When a child workflow completes before the parent workflow instance has been persisted to the database.
   * The bookmark pointing to the parent is already persisted, but the parent workflow does not yet exist in the DB.
   * This results in an exception during bookmark resumption.
   * The exception is silently swallowed with no trace, causing the background processor to stall.

2. **Bookmark Queue Batch Skipping:**

   * The background worker processes bookmarks in batches.
   * If a bookmark resumption fails (e.g., due to a missing workflow instance in the DB), the remainder of that batch is skipped.
   * This causes additional bookmarks to be ignored, contributing to workflows not completing.

**Expected Behavior:**

* Exceptions during bookmark processing should be logged and handled explicitly.
* The bookmark queue processor should continue processing the remainder of the batch even if a bookmark fails to resume.
* `BulkDispatchWorkflows` should reliably complete once all child workflows have finished, regardless of execution timing.

**Suggested Fixes:**

* Add logging and error handling for failed bookmark resumptions.
* Modify batch processing logic to isolate and skip only failed bookmarks, not the entire remainder of the batch.
* Fix deadlock in bookmark queue processor.

sfmskywalker:
Improves the reliability and efficiency of the workflow runtime by:

- Refactors the `BookmarkQueueSignaler` to use channels for more robust signaling.
- Enhances the `BookmarkQueueWorker` to handle exceptions and continue processing.
- Adds a new `WorkflowInstanceNotFoundException` for clearer error handling when workflow instances are not found.
- Introduces a handler for `WorkflowInstanceSaved` to trigger bookmark queue processing upon workflow instance creation.
- Updates the BulkDispatchWorkflows activity to use null defaults.
- Clarifies comments related to bookmark and workflow instance checks during bookmark resumption.

Fixes #6642

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6643)

### DIFF
--- a/src/modules/Elsa.Workflows.Runtime/Services/BookmarkQueueWorker.cs
+++ b/src/modules/Elsa.Workflows.Runtime/Services/BookmarkQueueWorker.cs
@@ -7,7 +7,7 @@ namespace Elsa.Workflows.Runtime;
 public class BookmarkQueueWorker : IBookmarkQueueWorker
 {
     private readonly RateLimitedFunc<CancellationToken, Task> _rateLimitedProcessAsync;
-    private CancellationTokenSource _cts = default!;
+    private CancellationTokenSource _cts = null!;
     private bool _running;
     private readonly IBookmarkQueueSignaler _signaler;
     private readonly IServiceScopeFactory _scopeFactory;
@@ -18,7 +18,7 @@ public class BookmarkQueueWorker : IBookmarkQueueWorker
         _signaler = signaler;
         _scopeFactory = scopeFactory;
         _logger = logger;
-        _rateLimitedProcessAsync = Debouncer.Debounce<CancellationToken, Task>(ProcessAsync, TimeSpan.FromMilliseconds(500));
+        _rateLimitedProcessAsync = Throttler.Throttle<CancellationToken, Task>(ProcessAsync, TimeSpan.FromMilliseconds(500));
     }
 
     public void Start()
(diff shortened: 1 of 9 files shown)


src/modules/Elsa.Workflows.Runtime/Services/BookmarkQueueWorker.cs
```diff
@@ -47,8 +47,19 @@ private async Task AwaitSignalAsync()
...
+            try
+            {
+                await _signaler.AwaitAsync(_cts.Token);
+                await _rateLimitedProcessAsync.InvokeAsync(_cts.Token);
+            }
+            catch (OperationCanceledException)
+            {
+                break; // Stop() was called
+            }
+            catch (Exception ex)
```
raymonddenhaan, line 59: Is it guaranteed that the cancellation token will be triggered within a short timeframe or does this create a potential infinite loop?

sfmskywalker: The token will be triggered only when the application stops. Until then, this is for all intents and purposes an infinite loop.

raymonddenhaan: Will this part be executed on a background worker? Because that might lead to a possible exhaustion of background workers when multiple activities will be in an infinite loop here.

raymonddenhaan (approved):

src/modules/Elsa.Workflows.Runtime/Services/BookmarkQueueWorker.cs
sfmskywalker: Yes, this will be called from a background worker. I am not sure I see how this loop will lead to an exhaustion of background workers. Let me call you.

sfmskywalker: As discussed on the call, these are hosted background workers and not background workers used by mediator which has a finite set of workers. I.e. all good.

PR #6646, opened 2025-05-14
View suspended workflow in studio fix

MaxBrooks114:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6646)


sfmskywalker:
@MaxBrooks114 Does this have something to do with "older" workflow instances by any chance from earlier versions?

MaxBrooks114:
> @MaxBrooks114 Does this have something to do with "older" workflow instances by any chance from earlier versions?

No this is any suspended workflow using the current version. Though it should be noted that these workflows are programmatic, with the cron trigger that starts the workflow being edited in the designer to fire and test the workflow at the desired time.

sfmskywalker (approved):

PR #6648, opened 2025-05-15
fix: code fixed to allow loading all the descriptors in the UI

sagargolu:
@sfmskywalker 
This bug fix addresses the inconsistency in loading workflows marked as "Usable As Activity" in the Activity Panel.

Previously, when a user loaded a main workflow that included a sub-workflow with multiple versions (e.g., the third version), the system was sometimes unable to locate it in the activity descriptors. This issue was caused by a Distinct filter that excluded the intended version.

The fix resolves both the loading inconsistency and the version-mismatch problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6648)


sfmskywalker (approved):

sfmskywalker:
Many thanks! @sagargolu

PR #6660, opened 2025-05-19
Fix async callback awaiting in HTTP endpoint

sfmskywalker:
## Summary
- await HTTP resume callbacks to avoid losing async continuations
- make `WaitForHttpRequest` async and update callers

## Testing
- `dotnet test --no-build` *(fails: `dotnet` not found)*

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6660)

### DIFF
--- a/src/modules/Elsa.Http/Activities/HttpEndpointBase.cs
+++ b/src/modules/Elsa.Http/Activities/HttpEndpointBase.cs
@@ -20,10 +20,10 @@ public abstract class HttpEndpointBase<TResult> : Trigger<TResult>
     {
     }
 
-    protected override void Execute(ActivityExecutionContext context)
+    protected override async ValueTask ExecuteAsync(ActivityExecutionContext context)
     {
         var options = GetOptions();
-        context.WaitForHttpRequest(options, HttpRequestReceivedAsync);
+        await context.WaitForHttpRequest(options, HttpRequestReceivedAsync);
     }
 
     protected override IEnumerable<object> GetTriggerPayloads(TriggerIndexingContext context)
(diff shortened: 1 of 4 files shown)


src/modules/Elsa.Http/Activities/HttpEndpointBase.cs
```diff
@@ -20,10 +20,10 @@ protected virtual void OnHttpRequestReceived(ActivityExecutionContext activityEx
     {
     }
 
-    protected override void Execute(ActivityExecutionContext context)
+    protected override async ValueTask ExecuteAsync(ActivityExecutionContext context)
```
raymonddenhaan, line 23: Since this function is made async it should be awaited when called

raymonddenhaan (changes requested):

src/modules/Elsa.Http/Activities/HttpEndpointBase.cs
sfmskywalker: Correct. As far as I can see, this is the case. Please point me to where it's called without being awaited.

raymonddenhaan: Since the function signature changed I would expect that the calling function also needed to be changed. I didn't see that there were separate calls for async and sync functions within the Activity base class

raymonddenhaan (approved):

src/modules/Elsa.Http/Activities/HttpEndpointBase.cs
sfmskywalker: Ok, all good then.

PR #6751, opened 2025-06-24
Feat/6732 zip archive activities

lukhipolito-nexxbiz:
### Summary
This pull request introduces new modules, `Elsa.IO` and `Elsa.IO.Compression`, to extend the Elsa platform with IO and compression functionalities. It adds features for resolving content types to streams and creating ZIP archives, along with supporting infrastructure and integration into the existing solution.

#### IO Module (`Elsa.IO`):
* Added `IContentResolver` and `IFileExtensionResolver` interfaces for resolving content to streams and determining file extensions. [[1]](diffhunk://#diff-f5c2b13a37fd7acf47ee6b4cd498b8a95d737bd21584061b502960a8fb682e5cR1-R15) [[2]](diffhunk://#diff-93613d09ee75116a39d1abad4aa4244ea9426c56faa61dbcb3aa7e36dcec5a53R1-R15)
* Implemented the `ContentResolver` class using a strategy pattern to handle various content types such as byte arrays, streams, file paths, URLs, and base64 strings.

#### Compression Module (`Elsa.IO.Compression`):
* Added the `CreateZipArchive` activity to create ZIP archives from diverse content types, including streams, files, and base64 strings.
* Created the `ZipEntry` model to represent individual entries in a ZIP archive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6751)


lukhipolito-nexxbiz:
@dotnet-policy-service agree [company="nexxbiz"]

lukhipolito-nexxbiz:
@dotnet-policy-service agree company="nexxbiz"

src/modules/Elsa.IO/Features/IOFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <summary>
+/// A feature that installs IO services for resolving various content types to streams.
+/// </summary>
+public class IOFeature(IModule module) : FeatureBase(module)
+{
+    /// <inheritdoc />
+    public override void Apply()
+    {
+        // Add HttpClient for URL content strategy
+        Services.AddHttpClient();
```
sfmskywalker: It’s better to use a named HTTP client so we can apply distinct configurations for different modules. For example, one module might require resilience policies like retries or circuit breakers, while another might need to attach authorization headers. If all modules share the same unnamed HTTP client, their handlers and settings will be combined, potentially leading to unpredictable and hard-to-debug behavior. Named clients provide clear boundaries and make it easier to reason about each module’s HTTP behavior independently.

src/modules/Elsa.IO/Features/IOFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        // Register strategies in order of priority
```
sfmskywalker: Avoid relying on DI registration order for resolution priority.
Instead, let's have each strategy define its own `Priority`, making the intended order explicit and easier to maintain.
This also prevents bugs when new strategies are added or registration order changes from other modules.

Tip: use `float` for the Priority type. This way, any module is always able to "squeeze" in another strategy.

src/modules/Elsa.IO/Services/ContentResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+    private readonly ILogger<ContentResolver> _logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ContentResolver"/> class.
+    /// </summary>
+    public ContentResolver(IEnumerable<IContentResolverStrategy> strategies, ILogger<ContentResolver> logger)
+    {
+        _strategies = strategies;
+        _logger = logger;
+    }
```
sfmskywalker: If you remove the XML comment from the constructor, Rider will suggest using a Primary Constructor - please do so, as it will minimize the boilerplate code by removing the private fields. E.g.:

image

With primary constructor:

```csharp
public class ContentResolver(IEnumerable<IContentResolverStrategy> strategies, ILogger<ContentResolver> logger) : IContentResolver
{
    /// <inheritdoc />
    public async Task<Stream> ResolveContentAsync(object content, CancellationToken cancellationToken = default)
    {
        var strategy = strategies.FirstOrDefault(s => s.CanHandle(content));

        if (strategy == null)
        {
            throw new ArgumentException($"Unsupported content type: {content?.GetType()?.Name ?? "null"}");
        }

        return await strategy.ResolveAsync(content, cancellationToken);
    }
}
```

src/modules/Elsa.IO/Services/ContentResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Elsa.IO.Services;
+
+/// <summary>
+/// Resolves various content types to streams using a strategy pattern.
+/// </summary>
+public class ContentResolver : IContentResolver
+{
+    private readonly IEnumerable<IContentResolverStrategy> _strategies;
+    private readonly ILogger<ContentResolver> _logger;
```
sfmskywalker: This is an unused field, so let's remove it.

src/modules/Elsa.IO/Services/ContentResolver.cs
sfmskywalker: Perhaps an improvement for the AI development workflow is to figure out how to instruct GitHub Copilot Agent with agent.md (or agents.md?) with instructions for code style, such as:

- Use modern C# syntax
- Use primary constructors when possible.
- etc.

src/modules/Elsa.IO/Services/ContentResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /// <inheritdoc />
+    public async Task<Stream> ResolveContentAsync(object content, CancellationToken cancellationToken = default)
+    {
+        var strategy = _strategies.FirstOrDefault(s => s.CanHandle(content));
+
+        if (strategy == null)
+        {
+            throw new ArgumentException($"Unsupported content type: {content?.GetType()?.Name ?? "null"}");
```
sfmskywalker: The null-coalescing operator ? on content implies that `content` is potentially null. If that's true, then let's use the ? symbol on the type definition of the content parameter:

```csharp
object? content
```

src/modules/Elsa.IO/Services/ContentResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Initializes a new instance of the <see cref="ContentResolver"/> class.
+    /// </summary>
+    public ContentResolver(IEnumerable<IContentResolverStrategy> strategies, ILogger<ContentResolver> logger)
+    {
+        _strategies = strategies;
+        _logger = logger;
+    }
+
+    /// <inheritdoc />
+    public async Task<Stream> ResolveContentAsync(object content, CancellationToken cancellationToken = default)
```
sfmskywalker:
```suggestion
    public async Task<Stream> ResolveContentAsync(object? content, CancellationToken cancellationToken = default)

```

src/modules/Elsa.IO/Services/FileExtensionResolver.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Elsa.IO.Services;
+
+/// <summary>
+/// Resolves appropriate file extensions for various content types.
+/// </summary>
+public class FileExtensionResolver : IFileExtensionResolver
+{
+    /// <summary>
+    /// Ensures that a filename has an appropriate file extension based on its content type.
```
sfmskywalker: The current implementation tightly couples content type detection with the local file system by directly calling `File.Exists`. This introduces a few problems:

1. Testability: Any unit tests for this method must either mock the file system or rely on actual files being present, which makes tests brittle and harder to maintain.
2. Extensibility: The logic assumes that the only way to determine an extension from a file path is via the local file system. This limits future enhancements such as supporting remote file sources, virtual files, or files stored in a database or cloud storage.

A better approach would be to abstract the file existence and extension retrieval logic behind an interface (e.g., `IFileExtensionResolverStrategy`). This way, we can plug in alternative strategies for different environments, including tests.

src/modules/Elsa.IO/Services/Strategies/FilePathContentStrategy.cs
```diff
@@ -0,0 +1,98 @@
+namespace Elsa.IO.Services.Strategies;
+
+/// <summary>
+/// Strategy for handling file path content by reading from the filesystem.
+/// </summary>
+public class FilePathContentStrategy : IContentResolverStrategy
+{
+    /// <inheritdoc />
+    public bool CanHandle(object content)
```
sfmskywalker: The `CanHandle` and `ResolveAsync` methods seem to be doing the same file path validations, which gives us a good opportunity to make this entire class more DRY.

src/modules/Elsa.IO/Services/Strategies/IContentResolverStrategy.cs
```diff
@@ -0,0 +1,22 @@
...
+/// Defines a strategy for resolving specific content types to streams.
+/// </summary>
+public interface IContentResolverStrategy
+{
+    /// <summary>
+    /// Determines if this strategy can handle the specified content.
+    /// </summary>
+    /// <param name="content">The content to check.</param>
+    /// <returns>True if this strategy can handle the content, false otherwise.</returns>
+    bool CanHandle(object content);
```
sfmskywalker: Let's keep the two method names similar to each other. If the main method is `Resolve`, then we can rename this to `CanResolve`.

src/modules/Elsa.IO/Services/Strategies/UrlContentStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+    private readonly ILogger<UrlContentStrategy> _logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UrlContentStrategy"/> class.
+    /// </summary>
+    public UrlContentStrategy(ILogger<UrlContentStrategy> logger, IHttpClientFactory? httpClientFactory = null)
+    {
+        _logger = logger;
+        _httpClientFactory = httpClientFactory;
+    }
```
sfmskywalker: Turn into primary constructor, remove private fields.

src/modules/Elsa.IO/Services/Strategies/UrlContentStrategy.cs
```diff
@@ -0,0 +1,50 @@
...
+        var url = (string)content;
+
+        if (_httpClientFactory == null)
+        {
+            throw new InvalidOperationException("HTTP client factory is not available. Cannot download from URL.");
+        }
+
+        try
+        {
+            var httpClient = _httpClientFactory.CreateClient();
```
sfmskywalker: Use a named HTTP client (see earlier comment about this topic).

src/modules/Elsa.IO/Services/ContentResolver.cs
lukhipolito-nexxbiz: I believe we should just define an Agent Guideline md and add it to the context of the agent ("whenever you work on this repo, first read the agent guidelines", for example).
I can give it a shot, but maybe not for this PR

sfmskywalker: Agreed, out of scope for this PR.

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -0,0 +1,183 @@
...
+    [JsonConstructor]   
+    public CreateZipArchive(string? source = null, int? line = null) : base(source, line)
+    {
+    }
+
+    /// <summary>
+    /// The filename for the ZIP archive.
+    /// </summary>
+    [Input(Description = "The filename for the ZIP archive.")]
+    public Input<string?> Filename { get; set; } = null!;
```
sfmskywalker: Since this activity only produces a ZIP stream, it doesn't make much sense to ask for a filename just to return it as output. Even generating and returning a filename isn't particularly useful, as the activity doesn't actually do anything with it. In other words, filename generation falls outside the scope of this activity's responsibility. So, how about we remove it?

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -0,0 +1,183 @@
...
+        Description = "The entries to include in the ZIP archive. Can be byte[], Stream, file path, file URL, base64 string, ZipEntry objects, or arrays of these types",
+        UIHint = InputUIHints.MultiLine
+    )]
+    public Input<object?> Entries { get; set; } = null!;
+
+    /// <summary>
+    /// The filename that was used for this ZIP archive.
+    /// </summary>
+    [Output(Description = "The filename that was used for this ZIP archive.")]
+    public Output<string> ResultFilename { get; set; } = new();
```
sfmskywalker: See my previous comment about the filename: how about we remove it?

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -0,0 +1,183 @@
...
+        Stream contentStream;
+
+        if (entryContent is ZipEntry zipEntry)
+        {
+            contentStream = await resolver.ResolveContentAsync(zipEntry.Content, context.CancellationToken);
+        }
+        else
+        {
+            contentStream = await resolver.ResolveContentAsync(entryContent, context.CancellationToken);
+        }
```
sfmskywalker: Instead of explicitly checking the type of `entryContent`, it's preferable for the resolver strategy to return a normalized model that already includes the name to use for the entry. This eliminates the need for type checks (e.g., against `ZipEntry`) and allows us to take advantage of other content types that can also provide a meaningful name.

For instance, the URL resolver strategy could extract a file name from the `Content-Disposition` response header if available, while the File resolver strategy could simply return the file name.

To normalize this across strategies, we could introduce a `BinaryContent` class:

```csharp
class BinaryContent
{
    string? Name { get; set; }
    string? ContentType { get; set; }
    IDictionary<string, object> Headers { get; set; } // Optional
    Stream Stream { get; set; }
}
```

With this in place, the ZipEntry resolver strategy could look like this:

```csharp
class ZipEntryResolverStrategy(IServiceProvider serviceProvider) : IContentResolverStrategy
{
    async Task<BinaryContent> ResolveAsync(object content)
    {
        var zipEntry = (ZipEntry)content;
        var resolver = serviceProvider.GetRequiredService<IContentResolver>(); // Using service locator to avoid circular constructor dependencies

        var resolved = await resolver.ResolveAsync(zipEntry);
        return new BinaryContent
        {
            Name = zipEntry.EntryName,
            ContentType = resolved.ContentType,
            Headers = resolved.Headers,
            Stream = resolved.Stream
        };
    }
}
```

Other strategies would follow a similar pattern but skip recursion where unnecessary. For example, the URL resolver can directly return a `BinaryContent` instance:

```csharp
public async Task<BinaryContent> ResolveAsync(object content, CancellationToken cancellationToken = default)
{
    var url = (string)content;

    var httpClient = _httpClientFactory.CreateClient();
    var response = await httpClient.GetAsync(url, cancellationToken);
    var stream = await response.Content.ReadAsStreamAsync(cancellationToken);

    return new BinaryContent
    {
        Name = response.GetFileName(),
        ContentType = response.Content.Headers.ContentType?.MediaType,
        Stream = stream
    };
}
```

The URL example demonstrates how we can use a `GetFileName` extension method to extract the filename from the `Content-Disposition` header when present.

One area to further refine is how we handle recursive resolution. There may be other models in the future that wrap another content object (similar to `ZipEntry`), so we’ll want to apply the same resolution logic recursively in those cases as well.

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        entryName = fileExtensionResolver.EnsureFileExtension(zipEntry.EntryName ?? "temp", zipEntry.Content);
+        var extension = Path.GetExtension(entryName);
+
+        if (!entryName.EndsWith(extension, StringComparison.OrdinalIgnoreCase))
+        {
+            entryName += extension;
+        }
+        return entryName;
+    }
```
sfmskywalker: If the suggestion with the ZipEntry resolver strategy is implemented, this entire entry name logic could probably be moved there as well, cleaning up this activity class nicely.

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -0,0 +1,183 @@
...
+    {
+        var filename = Filename.Get(context) ?? DefaultArchiveName;
+
+        if (!filename.EndsWith(ZipExtension, StringComparison.OrdinalIgnoreCase))
+        {
+            filename += ZipExtension;
+        }
+
+        ResultFilename.Set(context, filename);
+    }
```
sfmskywalker: If we remove the FileName and ResultFileName properties, we can remove this method as well.

src/modules/Elsa.IO/Common/Constants.cs
```diff
@@ -0,0 +1,15 @@
+namespace Elsa.IO.Common;
+
+public static class Constants
+{
+    public const string IOHttpClientName = "IOFeatureHttpClient";
```
sfmskywalker: This made me realize that we need a third module: Elsa.IO.Http - and move all HTTP-related aspects from Elsa.IO to there.

src/modules/Elsa.IO/Features/IOFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <summary>
+/// A feature that installs IO services for resolving various content types to streams.
+/// </summary>
+public class IOFeature(IModule module) : FeatureBase(module)
+{
+    /// <inheritdoc />
+    public override void Apply()
+    {
+        // Add HttpClient for URL content strategy
+        Services.AddHttpClient(Constants.IOHttpClientName);
```
sfmskywalker: This would be part of the new Elsa.IO.Http module mentioned in my previous comment.

src/modules/Elsa.IO/Services/Strategies/UrlContentStrategy.cs
```diff
@@ -0,0 +1,41 @@
+using Elsa.IO.Common;
+using Microsoft.Extensions.Logging;
+
+namespace Elsa.IO.Services.Strategies;
+
+/// <summary>
+/// Strategy for handling URL content by downloading from HTTP/HTTPS URLs.
+/// </summary>
+public class UrlContentStrategy(ILogger<UrlContentStrategy> logger, IHttpClientFactory? httpClientFactory = null) : IContentResolverStrategy
```
sfmskywalker: When moving this to the newly proposed Elsa.IO.Http module, let's not inject nullable services - from there, the HttpClientFactory will always exist. Which also means no null0checking necessary further down below in the Resolve method.

src/modules/Elsa.IO.Http/Elsa.IO.Http.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    </PropertyGroup>
```
sfmskywalker: We're missing a package description and package tags.

src/modules/Elsa.IO.Http/Extensions/ModuleExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+/// <summary>
+/// Provides extensions to install the <see cref="IOHttpFeature"/> feature.
+/// </summary>
+public static class ModuleExtensions
+{
+    /// <summary>
+    /// Install the <see cref="IOHttpFeature"/> feature.
+    /// </summary>
+    public static IModule UseIOHttp(this IModule module, Action<IOHttpFeature>? configure = null)
```
sfmskywalker: It's better if we could call this on the IO feature, rather than as its own feature. For example, rather than having to do this:

```csharp
elsa.UseIO()
elsa.UseIOHttp()
```

We should be able to do this:

```csharp
elsa.UseIO(io => io.UseHttp())
```

The IO HTTP feature is a sub-feature of the IO feature, so to speak.

src/modules/Elsa.IO.Http/Extensions/ModuleExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        module.Configure(configure);
+        return module;
+    }
```
sfmskywalker:
```suggestion
    public static IOFeature UseHttp(this IOFeature io, Action<IOHttpFeature>? configure = null)
    {
        io.Module.Configure(configure);
        return io;
    }

```

src/modules/Elsa.IO/Extensions/ContentTypeExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+        if (mimeType.StartsWith("file/") || mimeType.StartsWith("@file/"))
+        {
+            var extension = mimeType[(mimeType.IndexOf('/') + 1)..];
+            if (!string.IsNullOrWhiteSpace(extension))
+                return "." + extension;
+        }
+        
+        return ".bin";
+    }
+}
```
sfmskywalker:
```suggestion
namespace Elsa.IO.Extensions;

using System;
using System.Collections.Generic;
using System.IO;
using System.Linq;

public static class ContentTypeExtensions
{
    private static readonly Dictionary<string, string> MimeMapping = new(StringComparer.OrdinalIgnoreCase);
    private static readonly Dictionary<string, HashSet<string>> ContentTypeToExtensionsMap = new(StringComparer.OrdinalIgnoreCase);
    
    static ContentTypeExtensions()
    {
        // Define all mappings in a single place
        AddMapping(".txt", "text/plain");
        AddMapping(".html", "text/html");
        AddMapping(".htm", "text/html");
        AddMapping(".css", "text/css");
        AddMapping(".js", "application/javascript");
        AddMapping(".json", "application/json");
        AddMapping(".xml", "application/xml");
        AddMapping(".jpg", "image/jpeg");
        AddMapping(".jpeg", "image/jpeg");
        AddMapping(".png", "image/png");
        AddMapping(".gif", "image/gif");
        AddMapping(".svg", "image/svg+xml");
        AddMapping(".pdf", "application/pdf");
        AddMapping(".doc", "application/msword");
        AddMapping(".docx", "application/vnd.openxmlformats-officedocument.wordprocessingml.document");
        AddMapping(".xls", "application/vnd.ms-excel");
        AddMapping(".xlsx", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet");
        AddMapping(".ppt", "application/vnd.ms-powerpoint");
        AddMapping(".pptx", "application/vnd.openxmlformats-officedocument.presentationml.presentation");
        AddMapping(".zip", "application/zip");
        AddMapping(".csv", "text/csv");
    }

    private static void AddMapping(string extension, string contentType)
    {
        // Map extension to content type
        MimeMapping[extension] = contentType;

        // Map content type to extension(s)
        if (!ContentTypeToExtensionsMap.TryGetValue(contentType, out var extensions))
        {
            extensions = new(StringComparer.OrdinalIgnoreCase);
            ContentTypeToExtensionsMap[contentType] = extensions;
        }
        
        extensions.Add(extension);
    }

    public static string GetExtensionFromContentType(this string? contentType)
    {
        if (string.IsNullOrEmpty(contentType))
            return ".bin";
            
        if (contentType.EndsWith("/pdf") || contentType == "application/pdf")
            return ".pdf";
        
        if (ContentTypeToExtensionsMap.TryGetValue(contentType, out var extensions) && extensions.Any())
        {
            // Return the first extension for this content type
            return extensions.First();
        }
        
        return DetermineExtensionFromMimeType(contentType);
    }

    public static string GetContentTypeFromExtension(this string filePath)
    {
        var extension = filePath.GetFileExtension();
        
        return MimeMapping.GetValueOrDefault(extension, "application/octet-stream");
    }
    
    public static string GetFileExtension(this string filePath)
    {
        return Path.GetExtension(filePath).ToLowerInvariant();
    }
    
    private static string DetermineExtensionFromMimeType(string mimeType)
    {
        if (mimeType.Contains("/pdf"))
            return ".pdf";
        if (mimeType.Contains("image/"))
            return ".img";
        if (mimeType.Contains("text/"))
            return ".txt";
        if (mimeType.Contains("audio/"))
            return ".audio";
        if (mimeType.Contains("video/"))
            return ".video";
        
        if (mimeType.StartsWith("file/") || mimeType.StartsWith("@file/"))
        {
            var extension = mimeType[(mimeType.IndexOf('/') + 1)..];
            if (!string.IsNullOrWhiteSpace(extension))
                return "." + extension;
        }
        
        return ".bin";
    }
}
```

src/modules/Elsa.IO/Services/Strategies/StreamContentStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+        if (stream is FileStream fileStream)
+        {
+            name = Path.GetFileName(fileStream.Name);
+        }
+        
+        var result = new BinaryContent
+        {
+            Stream = stream,
+            Name = name ?? "stream",
+            Extension = string.Empty
```
sfmskywalker: What if the filename carries a file extension? Shouldn't we store that extension in the Extension field?
Perhaps even better: instead of storing the file extension (if there is any) separately, perhaps it should simply be part of the Name property. This might make other parts of the code simpler. In other words, we assume that the Name contains the file extension, if any. To easily get the extension, we use the Path.GetExtension method where the extension is needed (e.g. to determine the content type).

src/modules/Elsa.IO/Services/Strategies/TextContentStrategy.cs
```diff
@@ -0,0 +1,32 @@
...
+    public Task<BinaryContent> ResolveAsync(object content, CancellationToken cancellationToken = default)
+    {
+        var textContent = (string)content;
+        var textBytes = Encoding.UTF8.GetBytes(textContent);
+        var stream = new MemoryStream(textBytes);
+        
+        return Task.FromResult(new BinaryContent
+        {
+            Name = "text.txt",
+            Extension = ".txt",
```
sfmskywalker: Case in point: we can simply keep Name and remove Extension - anywhere we need the extension, we'd do `Path.GetExtension(Name)`.

lukhipolito-nexxbiz:
@sfmskywalker All comments have been addressed. Now it properly handles different extensions.
Additionally Elsa.IO.Http module is not a dependency of Elsa.IO.Compression anymore

PR #6783, opened 2025-07-10
Implementing better base64 evaluation + appropriate zip entry naming

lukhipolito:
This pull request introduces enhancements to the `CreateZipArchive` activity for handling duplicate entry names, updates the logic for validating Base64 strings, and refines code behavior to improve reliability and maintainability.

### Enhancements to Base64 Validation:
* Refined the `IsBase64String` method to validate padding positions and count, ensuring padding is correctly placed at the end and all trailing characters after the first '=' are also '='.
* Added checks to prevent false positives for short strings that consist only of lowercase letters and digits, improving the robustness of Base64 validation.

### Improvements to Zip Archive Creation:
* Changed the `ZipArchiveMode` from `Create` to `Update` in `CreateZipStreamFromEntries`, allowing read operations inside the entries besides just creating them.
* Added logic to generate unique entry names in `ProcessZipEntry` to prevent overwriting existing entries with the same name, following Windows naming conventions. This includes the new helper method `GetUniqueEntryName` for determining the next available name based on existing entries. [[1]](diffhunk://#diff-a78fb3ecdf96a52f900846139b12f877c93f2be449c511d3dd20fff814498c6fR129-R131) [[2]](diffhunk://#diff-a78fb3ecdf96a52f900846139b12f877c93f2be449c511d3dd20fff814498c6fR143-R193)

This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6783)

### DIFF
--- a/src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
+++ b/src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
@@ -80,7 +80,7 @@ public class CreateZipArchive : CodeActivity<Stream>
 
         try
         {
-            using var zipArchive = new ZipArchive(zipStream, ZipArchiveMode.Create, leaveOpen: true);
+            using var zipArchive = new ZipArchive(zipStream, ZipArchiveMode.Update, leaveOpen: true);
             var entryIndex = 0;
 
             var compressionLevel = CompressionLevel.Get(context);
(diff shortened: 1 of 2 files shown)


src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -143,52 +143,52 @@ private static async Task ProcessZipEntry(
...
-    
-        // Find the highest index used for this filename pattern
-        int highestIndex = 0;
+        var filenameWithoutExtension = Path.GetFileNameWithoutExtension(originalName);
+        var extension = Path.GetExtension(originalName);
+        
+        var originalExists = false;
+        var highestIndex = 0;
         
-        // Check for the original name and any name with pattern "name(n).ext"
```
sfmskywalker: It would be good to keep this comment somewhere to explain what this function is doing with the parenthesis thing. Also, what sort of pattern is this? Is it an arbitrary scenario we encountered, or is this a common thing in the world of zip archives?

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
```diff
@@ -143,52 +143,52 @@ private static async Task ProcessZipEntry(
...
+                    var indexStr = entryNameWithoutExtension.Substring(
+                        filenameWithoutExtension.Length + 1,
                         closingParenIndex - filenameWithoutExtension.Length - 1);
-                    
+    
                     if (int.TryParse(indexStr, out int index))
                     {
                         highestIndex = Math.Max(highestIndex, index);
                     }
                 }
```
sfmskywalker: It would be easier to read the overall function if this number extraction logic were encapsulated into its own function.

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
lukhipolito: this is the Windows-style way of dealing with equal filenames. Since we are using a microsoft library to create the zip files, I thought that using this standard for file and entry naming also makes sense.

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
lukhipolito: you are right. done it

src/modules/Elsa.IO.Compression/Activities/CreateZipArchive.cs
lukhipolito: I added a comment there explaining what is the standard

sfmskywalker (approved):

PR #6786, opened 2025-07-10
Update RawStringContent encoding in JsonContentFactory

MaxBrooks114:
Modified the instantiation of `RawStringContent` to use a new `UTF8Encoding` instance with `encoderShouldEmitUTF8Identifier` set to `false`, affecting the handling of the UTF-8 byte order mark (BOM) in serialized JSON content. Fixes a bug with content length being different than expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6786)


sfmskywalker:
@MaxBrooks114 Please review Copilot's comments, then I'd be happy to merge it. Thanks for the fix!

PR #6792, opened 2025-07-14
Refactors workflow reference updates

sfmskywalker:
Streamlines workflow reference updating to improve dependency resolution and prevent concurrency issues.

- Introduces recursive dependency resolution to ensure all references are consistently updated.
- Prevents concurrent updates to avoid race conditions and data corruption.
- Simplifies activity descriptor creation with a dedicated factory.
- Employs topological sorting to process workflows in the correct order based on dependencies.


FransVanEk:
i wonder if it if implemented fully and if this is correct considering your pr description of sort. 

i see the implementation of the new way of filtering but i wonder if applied correctly. perhaps you can talk me through the implementation a bit

src/modules/Elsa.Workflows.Management/Contracts/IWorkflowReferenceQuery.cs
```diff
@@ -0,0 +1,15 @@
+namespace Elsa.Workflows.Management;
+
+/// <summary>
+/// Finds all latest versions of workflow definitions that reference a specific workflow definition.
+/// </summary>
+public interface IWorkflowReferenceQuery
```
FransVanEk: in the implementation of this  we should prevent loading of workflow definition ids that were already being loaded for 2 reasons 

- performance 
- circular referencing

sfmskywalker: Great catch. Fixed.

PR #6803, opened 2025-07-17
Added null checks to ActionFactory

RalfvandenBurg:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6803)


sfmskywalker:
I'm wondering if handling these null values isn't going to hide the underlying problem of why the descriptor was null in the first place.

RalfvandenBurg:
@sfmskywalker I've added a test to get this issue. MaterializeWorkflowPriorToPopulateRegistriesAsync 
the commented code await host.PopulateRegistriesAsync(); fixes it. but that isn't automatically done out of the box. do you know why that is?

RalfvandenBurg:
i think, i understand as to why this can happen. 
activityRegistryPopulator.PopulateRegistryAsync loops thru all IEnumerable<IActivityProvider> providers.
(thus also the WorkflowDefinitionActivityProvider).
This will attempt to load the definitions from the store. so there 2 ways that this can go wrong.
1. If the order of providers is returned in the wrong order (i think you can not rely on the order). 
2. The first time because the _activitydescriptors are set at the end of the method (Interlocked.Exchange(ref _activityDescriptors, activityDescriptors); which shall not be reached since this is an exception when materializing the workflowdefinition.

PR #6841, opened 2025-08-05
Improving behaviour of Run Asynchronously property for the TaskActivity Attribute

lukhipolito:
This pull request updates how the system determines whether workflow activities should run asynchronously, introducing better support for nullable values and attribute-based defaults. The changes primarily affect the handling of the `RunAsynchronously` property, ensuring more flexible and accurate behavior for different activity types.

### Improvements to asynchronous execution handling

* Changed the `RunAsynchronously` property in `Activity` and related extension methods to use `bool?` instead of `bool`, allowing for explicit null values and improved defaulting logic. [[1]](diffhunk://#diff-10e7b59fb4df17c4453722a96d8e334bac93234a00d8ce801fe737e5cfd38c90L71-R71) [[2]](diffhunk://#diff-9837aedf227716c992139cd9706f0b4c86282611660bde423202fc362880d950L32-R42)
* Updated the logic in `ActivityFactory` to use a new `GetNullableBoolean` helper for reading `runAsynchronously`, falling back to the `TaskActivityAttribute` value or `false` if not specified. [[1]](diffhunk://#diff-77eb1a045a5be96c43841e93c99d43e2371c98b6674a05307fa4dde94f669ba5L37-R49) [[2]](diffhunk://#diff-77eb1a045a5be96c43841e93c99d43e2371c98b6674a05307fa4dde94f669ba5R212-R231)
* Added the `GetTaskRunAsynchronously` method to `BackgroundActivityInvokerMiddleware`, which checks the attribute and property for task activities to decide if they should run asynchronously.

### Code consistency and attribute usage

* Added missing `using Elsa.Workflows.Attributes;` to files that now rely on activity attributes for defaulting logic. [[1]](diffhunk://#diff-77eb1a045a5be96c43841e93c99d43e2371c98b6674a05307fa4dde94f669ba5R5) [[2]](diffhunk://#diff-11381be89f80cb98323f6489578970d4d03f250d2e3d499187f42df60a45c09fR4)

### Minor improvements

* Updated the `GetCanStartWorkflow` extension method to use an explicit generic type for property retrieval, improving type safety.

This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6841)


src/modules/Elsa.Workflows.Core/Services/ActivityFactory.cs
```diff
@@ -197,4 +209,24 @@ private static bool GetBoolean(JsonElement element, string propertyName)
...
+            }
+
+            if (element.TryGetProperty(propertyName.Camelize(), out var property) 
+                && (bool?)property.GetValue() is { } propValue)
+                return propValue;
+        }
+
+        return null;
+    }
 }
```
sfmskywalker: I think you should be able to extract a reusable method from this that can be used by two other methods to get a boolean and a nullable boolean.

sfmskywalker (commented):
Looks great. Let's make the `GetNullableBoolean` and `GetBoolean` reuse a new method that captures the core property value extraction logic to make it a bit more DRY. Then merge.

FransVanEk (approved):

src/modules/Elsa.Workflows.Core/Services/ActivityFactory.cs
lukhipolito-nexxbiz: I updated it, take a look

src/modules/Elsa.Workflows.Core/Extensions/ActivityPropertyExtensions.cs
```diff
@@ -39,7 +39,7 @@ public static class ActivityPropertyExtensions
     /// By default, activities with an <see cref="ActivityKind"/> of <see cref="Action"/>, <see cref="Task"/> or <see cref="Trigger"/>
     /// will execute synchronously, while activities of the <see cref="ActivityKind.Job"/> kind will execute asynchronously.
     /// </summary>
-    public static void SetRunAsynchronously(this IActivity activity, bool? value) => activity.CustomProperties[RunAsynchronouslyPropertyName[0]] = value;
+    public static void SetRunAsynchronously(this IActivity activity, bool? value) => activity.CustomProperties[RunAsynchronouslyPropertyName[0]] = value ?? false;
```
sfmskywalker, line 42: It doesn't seem right that we allow to pass in a null value but then store it as false. If we pass in null, null should be stored. Alternatively, the value should be typed as non-nullable.

lukhipolito-nexxbiz: Well the reasoning behind this decision is that I depend on the nullability to determine whether it has just been initialized and has never been assigned.
That is what the factory method uses in order to overwrite that value based on the TaskActivity attribute.
Once it has been assigned by any other means, it is not useful to have it null anymore.

sfmskywalker (approved):

PR #6848, opened 2025-08-07
Downgrading JetBrains.Annotations for fixing the packages build

lukhipolito:
The latest version of `JetBrains.Annotation package` was causing a problem in the package build pipeline, in the test step, causing it to run forever and get the pipeline execution cancelled for timeout.

This PR fixes the elsa 3 package build by downgrading the `JetBrains.Annotation` package.


KnibbsyMan:
Hi @lukhipolito, I will check this tonight to see if I can get my local tests to now pass. Good find!

PR #6872, opened 2025-08-22
Introduces modular logging framework

sfmskywalker:
Refactors the logging system to a modular framework:

- Introduces configurable and extensible log sinks.
- Updates dependencies and codebase to align with the new framework.
- Adds Console and Serilog sink implementations.
- Improves logging configuration handling through appsettings.json.
- Adds Log activity to workflows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6872)


src/modules/Elsa.Logging.Console/ConsoleLogSinkFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+            builder.AddCategoryFilters(options);
+
+            var min = options.MinLevel ?? LogLevel.Information;
+
+            switch (options.Formatter.ToLowerInvariant())
+            {
+                case "simple":
+                    builder.AddSimpleConsole(o =>
+                    {
+                        if (options.TimestampFormat is not null) o.TimestampFormat = options.TimestampFormat;
```
FransVanEk, line 38: why is this in multiple places but with different options that are allowed. can we make it more dry?

src/apps/Elsa.Server.Web/Elsa.Server.Web.csproj
```diff
@@ -8,8 +8,11 @@
         <ProjectReference Include="..\..\modules\Elsa.EntityFrameworkCore.Oracle\Elsa.EntityFrameworkCore.Oracle.csproj" />
         <ProjectReference Include="..\..\modules\Elsa.EntityFrameworkCore.PostgreSql\Elsa.EntityFrameworkCore.PostgreSql.csproj" />
         <ProjectReference Include="..\..\modules\Elsa.Kafka\Elsa.Kafka.csproj" />
+        <ProjectReference Include="..\..\modules\Elsa.Logging.Console\Elsa.Logging.Console.csproj" />
```
FransVanEk, line 11: what is the reasoning that they are only in certain apps not in all? i see that for other items as well, is there a logic to it?

src/apps/Elsa.ServerAndStudio.Web/Program.cs
```diff
@@ -142,6 +143,7 @@
             .UseWebhooks(webhooks => webhooks.ConfigureSinks = options => builder.Configuration.GetSection("Webhooks:Sinks").Bind(options))
             .UseWorkflowsApi()
             .UseCompression()
+            .UseLoggingFramework()
```
FransVanEk: i am missing any configuration for the sinks, is that intended?

src/modules/Elsa.Logging/Features/LoggingFeature.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        Module.ConfigureHostedService<LogEntryBackgroundWorker>();
+    }
+
+    public override void Apply()
+    {
+        Services
+            .AddScoped<ILogSinkProvider, ConfigurationLogSinkProvider>()
+            .AddScoped<ILogSinkProvider, StaticLogSinkProvider>()
+            .AddScoped<ILogSinkRouter, LogSinkRouter>()
```
FransVanEk, line 54: why scoped, is this not engine wide? and never changes?  if it can be single the background worker can be optimized not to fetch it each time

src/modules/Elsa.Logging.Console/ConsoleLogSinkFactory.cs
sfmskywalker: It's arguably easier to reason about when looking at an individual formatter option, because now, looking at any individual formatter name instantly shows what options are applicable to it. Now, this clarity does come at the cost of repetition. But to answer your second question, we sure can.

src/apps/Elsa.Server.Web/Elsa.Server.Web.csproj
sfmskywalker: I use this app as a workbench. Hence, I added the feature here to test-drive it.

src/apps/Elsa.ServerAndStudio.Web/Program.cs
sfmskywalker: Not my intention - I took this in when merging from someone's branch. I'll remove it, because without any configuration, we'll just see the Log activity with no sinks.

src/modules/Elsa.Logging.Console/ConsoleLogSinkFactory.cs
sfmskywalker: Actually, it turns out that each argument is of a different type, making it not straightforward to reuse the same assignments.

- `AddSimpleConsole(o) // o is of type SimpleConsoleFormatterOptions`
- `AddSystemdConsole(o)` // o is of type ConsoleFormatterOptions`
- `AddConsole(o)` // o is of type ConsoleLoggerOptions`

None of these options classes are related, i.e. they do not share a common base.

src/modules/Elsa.Logging/Features/LoggingFeature.cs
sfmskywalker: Providers being scoped allows an individual provider to inject other scoped dependencies, like database connections.

However, being scoped does not automatically mean they will be resolved each time a log entry gets emitted. All of the providers are resolved exactly once from the `LogSinkCataglog`, which itself is a singleton.
The catalog invokes each provider once to get their log sinks. These log sinks are then held in a private field of the catalog.

MariusVuscanNx (approved):

PR #6890, opened 2025-09-03
Otel improvements

MariusVuscanNx:
- Included workflow instance id, defintion id, definition version in the activity trace
- Introduced extensions in metadata and included retry data in the trace
- Included incident data in the span attributes

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6890)

### DIFF
--- a/src/modules/Elsa.OpenTelemetry/Middleware/OpenTelemetryTracingActivityExecutionMiddleware.cs
+++ b/src/modules/Elsa.OpenTelemetry/Middleware/OpenTelemetryTracingActivityExecutionMiddleware.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using System.Text;
 using Elsa.Common;
 using Elsa.Extensions;
 using Elsa.OpenTelemetry.Contracts;
@@ -93,23 +94,25 @@ public class OpenTelemetryTracingActivityExecutionMiddleware(ActivityMiddlewareD
         }
     }
 
-public string ToOtelFormat(string input)
-{
-    if (string.IsNullOrWhiteSpace(input))
-        return string.Empty;
-
-    var sb = new StringBuilder(input.Length);
-
-    foreach (var c in input)
+    public string ToOtelFormat(string input)
     {
-        if (char.IsWhiteSpace(c))
-            sb.Append('_');
-        else
-            sb.Append(char.ToLowerInvariant(c));
-    }
+        if (string.IsNullOrWhiteSpace(input))
+            return string.Empty;
 
-    return sb.ToString();
-}
+        var humanized = input.Humanize();
+
+        var sb = new StringBuilder(humanized.Length);
+
+        foreach (var c in humanized)
+        {
+            if (char.IsWhiteSpace(c))
+                sb.Append('_');
+            else
+                sb.Append(char.ToLowerInvariant(c));
+        }
+
+        return sb.ToString();
+    }
 }
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


src/modules/Elsa.OpenTelemetry/Middleware/OpenTelemetryTracingActivityExecutionMiddleware.cs
```diff
@@ -72,6 +73,31 @@ public async ValueTask InvokeAsync(ActivityExecutionContext context)
...
+            }
+        }
+    }
+
+    public string ToOtelFormat(string input)
+    {
+        var humanized = input.Humanize();
+        var correlationId = humanized.Replace(" ", "_").ToLowerInvariant();
+        return correlationId;
     }
```
sfmskywalker, line 101: I'm not sure, but maybe it's worthwhile optimizing this with a single string allocation instead of two, considering that this code executes for each activity execution.

```csharp
public string ToOtelFormat(string input)
{
    if (string.IsNullOrWhiteSpace(input))
        return string.Empty;

    var sb = new StringBuilder(input.Length);

    foreach (var c in input)
    {
        if (char.IsWhiteSpace(c))
            sb.Append('_');
        else
            sb.Append(char.ToLowerInvariant(c));
    }

    return sb.ToString();
}
```

sfmskywalker: I'm not sure, but maybe it's worthwhile optimizing this with a single string allocation instead of two, considering that this code executes for each activity execution.

```suggestion
public string ToOtelFormat(string input)
{
    if (string.IsNullOrWhiteSpace(input))
        return string.Empty;

    var sb = new StringBuilder(input.Length);

    foreach (var c in input)
    {
        if (char.IsWhiteSpace(c))
            sb.Append('_');
        else
            sb.Append(char.ToLowerInvariant(c));
    }

    return sb.ToString();
}
```

src/modules/Elsa.Workflows.Core/Extensions/ActivityExecutionContextExtensions.cs
```diff
@@ -436,6 +438,22 @@ public static IEnumerable<ActivityExecutionContext> GetDescendants(this Activity
         }
     }
 
+    /// <summary>
+    /// Sets extension data in the metadata. Represents specific data that is exposed generically for an activity.
+    /// </summary>
+    public static void SetExtensionsMetadata(this ActivityExecutionContext context, Dictionary<string, object?> data)
+    {
+        context.Metadata[ExtensionsMetadataKey] = data;
+    }
```
sfmskywalker: It would be better if this ExtensionsMetadata owns the dictionary, rather than relying on the caller to provide the dictionary, considering that other activities might also want to add extension data. If they were  to call this method, they will be overriding existing values.

Instead, let this extension method try to read an existing value at the ExtensionsMetadataKey, and if it doesn't exist, create it and add it to Metadata.

The SetExtensionsMetadata therefore should only have to receive a key and a value to store in the extensions dictionary.

sfmskywalker:
```suggestion
    public static void SetExtensionsMetadata(this ActivityExecutionContext context, string key, object? value)
    {
        var extensionsDictionary = context.GetExtensionsMetadata();
        
        if(extensionsDictionary == null) extensionsDictionary = new();
        
        extensionsDictionary[key] = value;
        
        context.Metadata[ExtensionsMetadataKey] = extensionsDictionary;
        
    }
```

sfmskywalker (changes requested):
Let's revisit the SetExtensionsMetadata to make sure it can work for other activities as well.

src/modules/Elsa.OpenTelemetry/Middleware/OpenTelemetryTracingActivityExecutionMiddleware.cs
MariusVuscanNx: I still kept the var humanized = input.Humanize(); so that it converts pascal case into a sentence with spaces

sfmskywalker (approved):

PR #6895, opened 2025-09-05
Fix race condition when sending same stimuli

sfmskywalker:
- Adds `IWorkflowResumer` and its implementation for workflow resumption.
- Marks `BookmarkResumer` and related interfaces as obsolete.
- Refactors dependent services to use `WorkflowResumer`.
- Enhances `ResumeBookmarkRequest` to include `ActivityInstanceId`.
- Updates logging and queue handling logic to align with the new resumption approach.


sfmskywalker:
> So the key is to use the activity instance id, to make sure that only one is picked and not multiple items?

The actual fix is in `WorkflowResumer`, where we acquire a distributed lock for the stimulus being processed.  
The activity instance ID change is unrelated to the fix itself, but it’s an important API improvement: an activity handle (a kind of union type) can represent an activity in multiple ways: hash, activity ID, or activity node ID, none of which are used internally in this context. In other words, the change corrects the false impression that activity handles are supported here (they are not).

PR #6904, opened 2025-09-10
Fixed content type check from execute endpoint

MariusVuscanNx:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6904)

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -8,6 +8,7 @@ on:
       - 'perf/*'
       - 'patch/*'
       - 'feat/*'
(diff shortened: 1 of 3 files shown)


.github/workflows/packages.yml
```diff
@@ -8,6 +8,7 @@ on:
       - 'perf/*'
       - 'patch/*'
       - 'feat/*'
+      - 'feature/*'
```
lukhipolito-nexxbiz: @MariusVuscanNx Please revert this change, it was already set as default to feat

PR #6928, opened 2025-09-25
Update nullable argument defaults and enhance ForEach tests with additional edge cases.

sfmskywalker:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6928)


FransVanEk:
The only question i have , is it a conscious choice, to allow changes in the collection  while iterating? i don't see any downsides except the potential rick or running forever...

sfmskywalker:
I went back and forth with this. On the one hand, I would prefer the ForEach activity to behave more closely to C#’s `foreach` which doesn’t allow changes in the enumerable. But that’s not how it currently works. So should the test cover the existing behavior, even if we might decide to change it? If we do, we could be breaking people’s workflows that might be relying on the ability to change the data source within iterations. Having the current behavior captured by a test communicates that we currently support this behavior. I can imagine us making this more strict, allowing users to opt in into “strict mode”. Having this test will prevent us then from accidentally breaking the current behavior.

Maybe there’s another angle to it that I haven’t considered?

PR #6959, opened 2025-10-10
Unit testing coverage for WriteLine activity

lukhipolito-nexxbiz:
Coverage for WriteLine activity

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6959)

### DIFF
--- a/doc/qa/test-guidelines.md
+++ b/doc/qa/test-guidelines.md
@@ -173,7 +173,7 @@ When in doubt, add the minimal unit tests plus one integration test that reprodu
 
 1. **For activity unit tests, prefer returned state from [`ExecuteActivityAsync`](../../test/unit/Elsa.Activities.UnitTests/Helpers/ActivityTestHelper.cs).** Always inspect on the returned context — it is deterministic for synchronous workflows.
 2. **Resume bookmarks explicitly.** Do not wait for external schedulers — call the engine's resume/trigger APIs in your test to continue execution.
-3. **For integration tests, Locate instances deterministically.** Use an instance id returned by [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) or attach a `CorrelationId` test variable and query [`IWorkflowInstanceStore.FindByCorrelationIdAsync(...)`](../../src/modules/Elsa.Workflows.Management/Contracts/IWorkflowInstanceStore.cs). Avoid using "latest" queries.
+3. **For integration tests, locate instances deterministically.** Use an instance id returned by [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) or attach a `CorrelationId` test variable and query [`IWorkflowInstanceStore.FindByCorrelationIdAsync(...)`](../../src/modules/Elsa.Workflows.Management/Contracts/IWorkflowInstanceStore.cs). Avoid using "latest" queries.
 
 ---
 


test/unit/Elsa.Activities.UnitTests/Console/WriteLineTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        // Assert
+        mockTextWriter.Received(1).WriteLine(expectedText);
+    }
+
+    [Fact]
+    public async Task Should_Use_Default_Provider_When_None_Configured()
+    {
+        // Arrange
+        const string expectedText = "Default provider test";
```
sfmskywalker: The term `expectedText` creates the expectation that it will be asserted somewhere - but that doesn't happen. Perhaps it would be better to give it a more neutral name, such as `textToWrite`.

test/unit/Elsa.Activities.UnitTests/Console/WriteLineTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        // Act
+        await ActivityTestHelper.ExecuteActivityAsync(writeLine, services =>
+        {
+            services.AddSingleton(mockProvider);
+        });
+
+        // Assert
+        mockTextWriter.Received(1).WriteLine(expectedText);
+    }
```
sfmskywalker: These tests are out of scope because they test the functionality of the various input types, rather than the functionality of the WriteLine activity. Given the WriteLine activity implementation, the only three relevant tests should be:

- Does the WriteLine write a given input to the TextWriter of the configured out stream provider?
- Does the WriteLine correctly handle null value?
- Does the WriteLine handle the scenario correctly when there's no `IStandardOutStreamProvider` registered with DI?

Arguably, you could test the various constructor overloads to cover that the provided value is actually stored in the Text property, but that might be taking it too far. Still, if you were to do that, that would be the extent of the test - no need to then also test their evaluated result to end up in the out stream - that was already covered by the first test.

doc/qa/test-guidelines.md
```diff
@@ -168,10 +171,9 @@ When in doubt, add the minimal unit tests plus one integration test that reprodu
 
 ## Deterministic patterns to avoid flaky tests
 
-1. **Prefer returned state from [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs).** Always inspect [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) results first — it is deterministic for synchronous workflows.
+1. **For activities unit tests, prefer returned state from [`ExecuteActivityAsync`](../../test/unit/Elsa.Activities.UnitTests/Helpers/ActivityTestHelper.cs).** Always inspect on the returned context — it is deterministic for synchronous workflows.
```
sfmskywalker:
```suggestion
1. **For activity unit tests, prefer returned state from [`ExecuteActivityAsync`](../../test/unit/Elsa.Activities.UnitTests/Helpers/ActivityTestHelper.cs).** Always inspect on the returned context — it is deterministic for synchronous workflows.
```

doc/qa/test-guidelines.md
```diff
@@ -168,10 +171,9 @@ When in doubt, add the minimal unit tests plus one integration test that reprodu
...
 2. **Resume bookmarks explicitly.** Do not wait for external schedulers — call the engine's resume/trigger APIs in your test to continue execution.
-3. **Locate instances deterministically.** Use an instance id returned by [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) or attach a `CorrelationId` test variable and query [`IWorkflowInstanceStore.FindByCorrelationIdAsync(...)`](../../src/modules/Elsa.Workflows.Management/Contracts/IWorkflowInstanceStore.cs). Avoid using "latest" queries.
-4. **Use short polling where necessary.** If you must poll the instance store (e.g., testing asynchronous controllers), use a short interval and a deterministic timeout (helper code snippets in examples above).
+3. **For integration tests, Locate instances deterministically.** Use an instance id returned by [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) or attach a `CorrelationId` test variable and query [`IWorkflowInstanceStore.FindByCorrelationIdAsync(...)`](../../src/modules/Elsa.Workflows.Management/Contracts/IWorkflowInstanceStore.cs). Avoid using "latest" queries.
```
sfmskywalker:
```suggestion
3. **For integration tests, locate instances deterministically.** Use an instance id returned by [`RunAsync`](../../src/modules/Elsa.Workflows.Core/Contracts/IWorkflowRunner.cs) or attach a `CorrelationId` test variable and query [`IWorkflowInstanceStore.FindByCorrelationIdAsync(...)`](../../src/modules/Elsa.Workflows.Management/Contracts/IWorkflowInstanceStore.cs). Avoid using "latest" queries.
```

PR #6966, opened 2025-10-14
Feat/unit test coverage switch

lukhipolito-nexxbiz:
Added unit test coverage for Switch Activity

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6966)


test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+                new("Case 1", Expression.LiteralExpression(false), Substitute.For<IActivity>())
+            },
+            Default = mockActivity
+        };
+
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
```
sfmskywalker: I don't see how this test asserts that the Default activity executes when no cases match.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+                new("Third", Expression.LiteralExpression(true), Substitute.For<IActivity>())
+            }
+        };
+
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
+    }
```
sfmskywalker: All of these tests can be merged into a single `[Theory]` where you provide data to the test method. Will greatly reduce repetative code.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+                new("Second True", Expression.LiteralExpression(true), Substitute.For<IActivity>())
+            }
+        };
+
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
+    }
```
sfmskywalker, line 105: How is this test asserting that the first case is executed? I only see an assertion that no exception was thrown.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+            },
+            Default = Substitute.For<IActivity>()
+        };
+
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
+    }
```
sfmskywalker: A good enhancement might be to assert the desired behavior when providing a null condition, which is, the scheduling of the Default activity.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+                new("Valid case", Expression.LiteralExpression(true), Substitute.For<IActivity>())
+            }
+        };
+
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
+    }
```
sfmskywalker: This is basically the same test as the second test, the only difference being that here, there is one `true` case instead of a `false` one. Better to merge these two into a theory.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+        // Arrange
+        var switchActivity = new Switch();
+        
+        // Test that the Cases collection is initialized
+        Assert.NotNull(switchActivity.Cases);
+        Assert.Empty(switchActivity.Cases);
+        
+        // Test that we can add cases
+        switchActivity.Cases.Add(new SwitchCase("Test", Expression.LiteralExpression(true), Substitute.For<IActivity>()));
+        Assert.Single(switchActivity.Cases);
```
sfmskywalker: This is testing the behavior of the `List<SwitchCase> class, which should be out of scope. If anything, this test might verify that whatever is added to the Cases collection, these cases are used in the ExecuteAsync method.

test/unit/Elsa.Activities.UnitTests/Branching/SwitchTests.cs
```diff
@@ -0,0 +1,349 @@
...
+        
+        // Test setting different modes
+        switchActivity.Mode = new Input<SwitchMode>(SwitchMode.MatchAny);
+        
+        // Act & Assert - Should not throw any exceptions
+        var exception = await Record.ExceptionAsync(async () => 
+            await ActivityTestHelper.ExecuteActivityAsync(switchActivity));
+        
+        Assert.Null(exception);
+    }
```
sfmskywalker, line 193: The activity not throwing an exception when MatchAny is set does not prove that the activity handles the default mode properly.

sfmskywalker (approved):

Issue #6971, opened 2025-10-15
Incorrect use of TypeConverter.IsValid for ConvertTo in ObjectConverter


bobhauser:
## Description
Within the `ObjectConverter.ConvertTo` method, there is logic intended to leverage registered `TypeConverter`s to convert from a source type to a target type:

```
var sourceTypeConverter = TypeDescriptor.GetConverter(underlyingSourceType);

if (sourceTypeConverter.CanConvertTo(underlyingTargetType))
{
    var isValid = targetTypeConverter.IsValid(value);

    if (isValid)
        return sourceTypeConverter.ConvertTo(value, underlyingTargetType);
}
```

This approach was intended to guard the conversion by first checking if the value is “valid” for conversion, then invoking `ConvertTo`.

### Issues with the current implementation

1. Wrong converter is used for `IsValid`

    - The previous code incorrectly called `targetTypeConverter.IsValid(value)`, it likely was meant  to call `sourceTypeConverter.IsValid(value)`

2. `TypeConverter.IsValid` does not support `ConvertTo` scenarios

    - The `IsValid` method is primarily intended to validate values for `ConvertFrom`, not `ConvertTo`.

### Resolution
The fix removes the `IsValid` check entirely for `ConvertTo` and instead attempts the conversion directly, while catching and ignoring any exceptions:

```
if (sourceTypeConverter.CanConvertTo(underlyingTargetType))
{
    // TypeConverter.IsValid is not supported for ConvertTo, so we have to try and ignore any exceptions.
    try
    {
        return sourceTypeConverter.ConvertTo(value, underlyingTargetType);
    }
    catch
    {
        // Ignore and try other conversion strategies.
    }
}
```
#### Benefits of this approach:

- Allows all valid TypeConverter.ConvertTo implementations to execute without being blocked by an unsupported IsValid call.
- Ensures that the conversion logic is robust and falls back gracefully to alternative conversion strategies if ConvertTo throws.
- Fixes bugs where conversions previously failed for types with registered TypeConverters even though ConvertTo would have succeeded.

## Steps to Reproduce

See the following test in associated PR to for an example test showing a conversion which previous failed (it successes with the fix applied) `Elsa.Workflows.Core.UnitTests.ObjectConversion.ConvertTo_WithRegisteredPersonTypeConverter_ConvertsFromPersonToJson`

## Environment
[...]

PR #6972, opened 2025-10-15
Fix ObjectConverter.ConvertTo to avoid using sourceTypeConverter.IsValid

bobhauser:
This resolves the following issue: https://github.com/elsa-workflows/elsa-core/issues/6971

This PR targets branch `patch/3.5.2`. Normally I would target `main` but there are some changes in the 3.5 stream that are not in main yet. Also, this is likely something that should be included in the next 3.5 patch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6972)


sfmskywalker (approved):
Awesome!

Issue #6979, opened 2025-10-16
Cannot configure workflow to persist internal state without persisting inputs/outputs


bobhauser:
## Description
Workflows cannot currently be configured to globally persist activity internal state independently of inputs and outputs. The existing persistence default resolution logic ties internal state persistence to the same default hierarchy used for inputs and outputs, effectively forcing them to share the same default setting.

As a result, if a workflow is configured not to persist inputs and outputs, the internal state will also not be persisted, even if explicitly configured to do so. This prevents valid configuration scenarios where internal state persistence should be enabled while other persistence modes remain disabled.

Although this was definitely a change in behavior, I am not suggesting that the default behavior should match the old behavior, where internal state was always persisted and inputs and outputs were configurable. Instead, it should simply be possible to configure a workflow to behave like it did previously.

## Expected Behavior
It should be possible to configure workflows so that internal state persistence can be controlled independently of input and output persistence. A workflow should be able to persist its internal state even when inputs and outputs are excluded, allowing finer-grained control over what data is saved without requiring all persistence modes to share the same default configuration.

## Actual Behavior
It is not currently possible to configure, at the workflow level, for internal state to be persisted (`Include`) when the workflow’s default persistence mode is set to `Exclude`. While this configuration can be applied at the individual activity level, there is no way to enable it globally for all activities in the workflow. The `internalState` setting defined in the workflow’s `logPersistenceConfig` section is ignored, causing internal state persistence to follow the workflow’s default mode instead of its own explicit configuration.

PR #6980, opened 2025-10-16
Allow internal state to be globally configured independently of inputs/outputs

bobhauser:
This PR fixes an issue where workflows could not persist internal state globally when the workflow default persistence mode was set to Exclude. The internalState setting defined at the workflow level was ignored, causing internal state to inherit the workflow default for inputs and outputs rather than its own configuration.

Issue: https://github.com/elsa-workflows/elsa-core/issues/6979

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6980)


bobhauser:
Closing this PR, will open a new one against branch `patch/3.5.2`

PR #6982, opened 2025-10-17
Allow internal state to be globally configured independently of inputs/outputs

bobhauser:
This PR fixes an issue where workflows could not persist internal state globally when the workflow default persistence mode was set to Exclude. The internalState setting defined at the workflow level was ignored, causing internal state to inherit the workflow default for inputs and outputs rather than its own configuration.

Issue: https://github.com/elsa-workflows/elsa-core/issues/6979

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6982)


sfmskywalker:
Thanks for the contribution. Closing because this targets an obsolete 3.x base branch that is no longer maintained.

Please retarget or reopen against current `main`, or the latest active `release/*` branch if the change still applies. Happy to review a fresh PR on a current base.

PR #6985, opened 2025-10-18
Restore Active Branch-Only Joining

sfmskywalker:
A previous change was made to address what was thought to be a bug, but it turns out the behavior was actually correct and intentional.

The join rule is designed to consider **only activated branches**.

In the attached example, the `Decision` activity activates the **False** branch, meaning the **True** branch remains inactive. As a result, the `Join` activity correctly ignores the inactive branch and executes as intended.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/6985)

### DIFF
--- a/doc/adr/0006-formalizing-join-behaviors.md
+++ b/doc/adr/0006-formalizing-join-behaviors.md
@@ -21,14 +21,15 @@ There was a need to improve the join implementation and formalize the expected j
 
 ### Join Modes
 
-There are two join modes available:
+There are three join modes available:
 
 1. **WaitAll** - The join waits for all active inbound branches to complete before proceeding
 2. **WaitAny** - The join proceeds as soon as any active inbound branch completes
+3. **WaitAllInbound** - The join waits for all defined inbound connections to be followed before proceeding (stricter than WaitAll)
 
 ### Explicit Joins
 
-When using the `Join` activity, users can explicitly select either `WaitAll` or `WaitAny` mode.
+When using the `Join` activity, users can explicitly select between `WaitAll`, `WaitAny`, or `WaitAllInbound` mode.
 
 ### Implicit Joins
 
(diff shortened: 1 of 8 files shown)


rolandgriesser:
Hi @sfmskywalker - I understand the reasoning and it's good to have a clear semantic now - although I wonder, how I would implement something like this with that behavior: (fyi, it didn't work before either, but I wanted to test this with 3.6 as you've mentioned that joining behavior was reworked there as far as I remember).
image
- The "Balancer not fictional" tests for a boolean value, if it's not set, I want my workflow to stop
- The "Join" is configured with "WaitAll"
- The three bottom branches would load some data/do things in parallel
- As the first branch never gets activated if the boolean value is false, the "Join" would continue if the other branches are activated and finish. Up to now, I was wondering if this is a bug and might be fixed in 3.6, as I thought "WaitAll" would really only continue if all branches finish
- (The step after the "Join" will check the boolean value again and finish the workflow to mitigate this. I'm aware that I also could just get rid of the parallel branches and serialize them, but I do have other workflows with this pattern as well, that are more complex than this. Also, I'm not quite sure if the current behavior also sometimes leads to race-conditions, e.g. the "Finish" in the "Decision-Else"-Path sometimes took a while to execute, so the Join already continued - which sounds like unintended behavior as well)

I'm fine with my current workarounds - I was just wondering if this will use-case will be covered as well in the future, e.g. by configuring the Join behavior even more detailed.

sfmskywalker:
Great question. I think that for Elsa 3.6, we should have all scenarios covered with the following merge modes (the new term for join mode):

- **Default**: waits for all **active inbound paths** to complete.
- **Converge**: waits for all inbound paths to complete. I.e., **all** inbound paths must be activated before the join completes.
- **Race**: continues as soon as one inbound path completes and ignores/cancels all other inbound paths.
- **Stream**: continues any inbound path.

This way, the workflow designer should offer full control over the desired merging behavior.
The only issue I see currently is that of visualization. Consider the following two executions:

image

image

The issue is that we can't immediately see how execution would flow just from the workflow structure.
So perhaps the best way to solve for that is to have explicit activities like the Join activity, similar to what BPMN offers.
I realize this is outside the scope of your question and feedback, but I am curious to hear any opinions on this matter as well.

sfmskywalker:
[This PR](https://github.com/elsa-workflows/elsa-core/pull/6993) proposes an addition to the set of merge modes, allowing you to control exactly whether a join needs to wait for all inbound branches (Converge), or only activated inbound branches (Merge). For 3.6.0.

sfmskywalker:
Brief status update: we are internally discussing the change of the WaitAll behavior to now wait for **active branches** only, rather than **all** branches previous to 3.5.

Essentially, as I now understand it, the change in interpretation was required to enable the **inclusive OR** construct mentioned in [this issue's thread](https://github.com/elsa-workflows/elsa-core/issues/4433).

I think that perhaps we made a mistake by changing the interpretation of WaitAll, and instead, should consider adding a third option, e.g. WaitAllActive, to allow the user to control whether they want to await for active branches only, or all branches.

That would maintain existing behavior before 3.5.
Ultimately, we are considering to add explicit node types to model OR, XOR etc. visually, which is recommended.

bobhauser:
I suggest that this PR be abandoned in favour of this new PR: https://github.com/elsa-workflows/elsa-core/pull/7006 Restore pre-3.5 Join WaitAll behavior. It implements the `WaitAllActive` join mode suggested by @sfmskywalker in his previous comment.

@rolandgriesser - I believe that the change in the PR I mentioned about addresses the very valid issue you raised. I believe that the example you specified in your earlier comment is addressed by this [InlineData test](https://github.com/elsa-workflows/elsa-core/pull/7006/files#diff-9148b25183c3922d81337f040e3cf0d3be2864d87970752719d970701c1e1e14R350). It involves a WaitAll join where all of the inbound connections are not followed (a.k.a. activated). I would appreciate if you could confirm this for me.

PR #7001, opened 2025-10-21
Unit tests for For activity

lukhipolito-nexxbiz:
Unit test coverage for For activity

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7001)


test/unit/Elsa.Activities.UnitTests/Looping/ForTests.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        // Act
+        var context = await fixture.ExecuteAsync();
+
+        // Assert
+        Assert.True(context.HasScheduledActivity(mockBody));
+        
+        var currentValue = context.GetActivityOutput(() => forActivity.CurrentValue);
+        Assert.Equal(100, currentValue);
+    }
```
sfmskywalker: Redundant test. This is essentially a duplicate of scenarios already covered in the first test (`ExecutesValidLoopConfigurations_ScheduledChildActivity`)

test/unit/Elsa.Activities.UnitTests/Looping/ForTests.cs
```diff
@@ -0,0 +1,255 @@
...
+using Elsa.Workflows.Behaviors;
+
+namespace Elsa.Activities.UnitTests.Looping;
+
+public class ForTests
+{
+    [Theory]
+    [InlineData(1, 3, 1)] // Ascending loop
+    [InlineData(5, 1, -1)] // Descending loop  
+    [InlineData(10, 12, 1)] // UpdatesCurrentValueEachIteration
```
sfmskywalker: What does this comment mean? How does this test differ from the first scenario (Ascending loop)?

test/unit/Elsa.Activities.UnitTests/Looping/ForTests.cs
```diff
@@ -0,0 +1,255 @@
...
+    [InlineData(0, 4, 1)] // HandlesFloatingPointSteps equivalent
```
sfmskywalker: What does this comment mean? How does this test differ from the first scenario (Ascending loop)?

test/unit/Elsa.Activities.UnitTests/Looping/ForTests.cs
```diff
@@ -0,0 +1,255 @@
...
+    [InlineData(10, 15, 1)] // PreservesIterationCountAccuracy
```
sfmskywalker: What does this comment mean? How does this test differ from the first scenario (Ascending loop)?

test/unit/Elsa.Activities.UnitTests/Looping/ForTests.cs
```diff
@@ -0,0 +1,255 @@
...
+            Step = new Input<int>(1),
+            Body = mockBody
+        };
+        var fixture = new ActivityTestFixture(forActivity);
+
+        // Act
+        var context = await fixture.ExecuteAsync();
+
+        // Assert
+        bool shouldSchedule = ShouldExecuteLoop(start, end, 1);
```
sfmskywalker: **Pass expected values as theory arguments** is generally better practice for unit tests. Here's why:

## Advantages of explicit expected values:
1. **Test data is self-documenting** - You see inputs AND expected outputs at a glance without jumping to helper methods
2. **No hidden logic** - The test doesn't depend on correctly implementing the same logic as the production code
3. **Easier to maintain** - Changing one test case doesn't require understanding the helper method
4. **Prevents false positives** - If your helper has the same bug as the production code, all tests pass incorrectly

## When helper methods are acceptable:
- Very complex expected value calculations that would clutter test data
- When you're testing against a known-good reference implementation
- When the calculation is truly independent of the code under test

## For your case:

Since `ShouldExecuteLoopWithBounds` mirrors the For activity's internal logic, you risk having the same bugs in both places. Compare:

```csharp
// Current - obscures expectation
[InlineData(0, 4, 1)]
bool shouldSchedule = ShouldExecuteLoop(start, end, 1);

// Better - clear and explicit
[InlineData(0, 4, 1, true)]  // Zero start, should execute
```


**Recommendation**: Replace with explicit boolean parameters like you already did in `LoopDecisionLogic_ValidatesCorrectly` (line 206-225). That test is actually better structured than the ones using helper methods.

PR #7016, opened 2025-10-27
Feat/unit test coverage writefilehttpresponse

lukhipolito-nexxbiz:
Unit test coverage for Write File Http Response

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7016)

### DIFF
--- a/test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
+++ b/test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
@@ -700,7 +700,7 @@ public class WriteFileHttpResponseTests
                 {
                     var entryName = $"file{fileIndex++}.txt";
                     var entry = archive.CreateEntry(entryName, CompressionLevel.Optimal);
-                    // ReSharper disable once ConvertToUsingDeclaration
+                    
                     using (var entryStream = entry.Open())
                     {
                         var contentBytes = Encoding.UTF8.GetBytes(fileContent.ToString() ?? "");


test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
```diff
@@ -0,0 +1,744 @@
...
+        // Act
+        var context = await ExecuteActivityAsync(activity, httpContext);
+
+        // Assert
+        Assert.True(context.IsCompleted);
+        Assert.Equal(StatusCodes.Status204NoContent, httpContext.Response.StatusCode);
+    }
+
+    [Fact]
+    public async Task Should_Return_ApplicationZip_When_Content_Is_Empty_Array()
```
FransVanEk, line 160: is that so? not just null

test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
```diff
@@ -0,0 +1,744 @@
...
+        
+        activity.Content = new Input<object>(testContent);
+        // Filename not set, should default to "file.bin"
+
+        // Act
+        var context = await ExecuteActivityAsync(activity, httpContext);
+
+        // Assert
+        Assert.True(context.IsCompleted);
+        Assert.Contains("filename=file.bin", httpContext.Response.Headers.ContentDisposition.ToString());
```
FransVanEk, line 277: is this the real default name?

test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
```diff
@@ -0,0 +1,744 @@
...
+    }
+
+    [Fact]
+    public async Task Should_Create_Bookmark_When_No_HttpContext_Available()
+    {
```
FransVanEk, line 282: A bookmark for? 

is this the current implementation?

test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
lukhipolito-nexxbiz: yeah that is the current behaviour

test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
lukhipolito-nexxbiz: Yes as the Activity sets it

test/unit/Elsa.Activities.UnitTests/Http/WriteFileHttpResponseTests.cs
lukhipolito-nexxbiz: I am not sure if that is correct, but that is how it works currently:
Whenever there is no context it cannot write so creates a bookmark, the next time it is called uses the context and resumes the bookmark, then it writes normally

FransVanEk: ok , put it on the list of strange behaviors :-)

PR #7029, opened 2025-11-03
Fixed ParentWorkflowInstanceId not being set

crashkonijn:
Hello,

Whilst working with sub-workflows and trying to gather their data I noticed the `ParentWorflowInstanceId` value, however this value is always null.

After a lot of digging it appears that the value is being set on the `DispatchWorkflowDefinitionRequest`, but is not being passed to the `DispatchWorkflowDefinitionCommand`.

This PR adds this value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7029)


crashkonijn:
@dotnet-policy-service agree company="Autotaalglas"

PR #7031, opened 2025-11-03
Unit test coverage Flow Join activity

lukhipolito-nexxbiz:
Unit test coverage for Flow Join activity

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7031)


test/unit/Elsa.Activities.UnitTests/Branching/FlowJoinTests.cs
```diff
@@ -0,0 +1,291 @@
...
+            Flowchart.UseTokenFlow = originalValue;
+        }
+    }
+
+    [Theory]
+    [InlineData(true, FlowJoinMode.WaitAny)]
+    [InlineData(true, FlowJoinMode.WaitAll)]
+    [InlineData(false, FlowJoinMode.WaitAny)]
+    [InlineData(false, FlowJoinMode.WaitAll)]
+    public async Task Should_Handle_Complex_Flowchart_Scenarios(bool useTokenFlow, FlowJoinMode joinMode)
```
sfmskywalker: This looks more like an integration test than a unit test. Better to do this from the integration tests project.

sfmskywalker:
Can we make the tests more DRY?

test/unit/Elsa.Activities.UnitTests/Branching/FlowJoinTests.cs
lukhipolito-nexxbiz: Indeed, I was unsure at first, but I'll move it

test/integration/Elsa.Activities.IntegrationTests/Branching/FlowJoinTests.cs
```diff
@@ -0,0 +1,228 @@
...
+        // Act
+        await RunFlowchartAsync(_services, flowchart);
+
+        // Assert
+        Assert.Contains("Start", _output.Lines);
+        
+        if (useTokenFlow)
+        {
+            // In token flow mode, FlowJoin acts as a no-op
+            // Both branches should execute and continue to AfterJoin
```
sfmskywalker: I'm not sure this is entirely correct - even with token mode, the JoinMode property of the FlowJoin activity is significant, because the value is used to derive the `MergeMode`.

test/integration/Elsa.Activities.IntegrationTests/Branching/FlowJoinTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            Assert.Contains("AfterJoin", _output.Lines);
+        }
+        else
+        {
+            // In counter flow mode:
+            // - WaitAny should allow continuation after first inbound connection
+            // - WaitAll should wait for all inbound connections (in this case just one)
+            // Since there's only one inbound connection, both should behave similarly
+            Assert.Contains("AfterJoin", _output.Lines);
+        }
```
sfmskywalker: 1. These two assertions are exactly the same.
2. Which makes sense, because the expected behavior of the Join activity is the same regardless of the algorithm used.

Given that, it would be better to just do the assertion, regardless of the `useTokenFlow` flag.

sfmskywalker (approved):

Issue #7077, opened 2025-11-19
Workflow instance reappears after deletion


sfmskywalker:
### Steps to Reproduce

1. Import [this JSON workflow file](https://github.com/user-attachments/files/23636922/delay-bug-fail-to-remove.json) via Elsa Studio and start it.

2. Try to remove the workflow instances. Sometimes this may succeed, but more often than not, the workflow instance will reappear.

### Symptoms

Deleting a workflow instance causes it to reappear shortly afterward.
In the observed case, a workflow contained a very short delay (`00:00:00.01`).
After deletion:

* The workflow instance and definition were removed from the database.
* **A bookmark remained**, e.g.:

```
"bc1dc652fec843e0" "Elsa.Delay" "5AA4AB21EBB35F0F988AE5F34CF1D145C06D8CA8454CC3A2EFCA3F07C93D72CE" "249c0372dede013" "4d19dbf0d5f10200" "2025-11-13 13:30:13.321016+01" "{""resumeAt"":""2025-11-13T12:30:13.3309712+00:00"",""_type"":""Elsa.Scheduling.Bookmarks.DelayPayload, Elsa.Scheduling""}" "Elsa.Delay"
```

Because the bookmark still referenced a workflow that no longer existed, **the engine failed on startup and restarted repeatedly**.

---

### Root Cause

Deletion performed through the API removes the workflow instance **only from the database**, while a running in-memory instance continues to execute.
When this in-memory instance next persists its state, it **writes its data back to the database**, effectively resurrecting the deleted instance.
Bookmarks created by the running instance also remain, causing startup failures.

---

#### 1. Delegate deletion to the workflow runtime

The API should call a runtime-level `DeleteInstanceAsync` method instead of deleting directly from storage.
Each runtime (distributed, ProtoActor, etc.) will be responsible for safely stopping and deleting its instances.

#### 2. Use runtime-specific coordination

* **Distributed runtime**: perform deletion under the same distributed lock used for execution, preventing concurrent DB writes.
* **ProtoActor runtime**: send a control message instructing the actor to stop and delete itself.

#### 3. Add a “Deleted” terminal state (soft delete)

Mark workflow instances as `Deleted` instead of hard-deleting immediately.
Schedulers and runtimes treat this as terminal and must not resume or persist such instances.
A background cleanup task may permanently remove old deleted records.

#### 4. (Optional) Introduce optimistic concurrency

Add a revision/version field to workflow instances.
[...]

PR #7088, opened 2025-11-20
Improves workflow deletion and task scheduling

sfmskywalker:
Addresses issues related to workflow instance deletion, trigger indexing, and task scheduling.

- Ensures proper cancellation and cleanup of running workflows during deletion by using the workflow runtime.
- Implements exception handling for trigger deletion to prevent failures when loading workflows with missing materializers.
- Introduces thread safety to the local scheduler to avoid race conditions during concurrent scheduling.
- Handles exceptions for resume workflow tasks to prevent errors when scheduled workflow instances are deleted.

Relates to #7077

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7088)

### DIFF
--- a/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/TriggerIndexing/Fakes/FailingMaterializer.cs
+++ b/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/TriggerIndexing/Fakes/FailingMaterializer.cs
@@ -1,0 +1,18 @@
+using Elsa.Workflows.Activities;
+using Elsa.Workflows.Management;
+using Elsa.Workflows.Management.Entities;
+
+namespace Elsa.Workflows.IntegrationTests.Scenarios.TriggerIndexing.Fakes;
+
+/// <summary>
+/// A custom materializer that throws an exception when materialization is attempted.
+/// This simulates the "Provider not found" scenario.
+/// </summary>
+public class FailingMaterializer : IWorkflowMaterializer
+{
+    public const string MaterializerName = "FailingMaterializer";
+    public string Name => MaterializerName;
+
+    public ValueTask<Workflow> MaterializeAsync(WorkflowDefinition definition, CancellationToken cancellationToken = default) =>
+        ValueTask.FromException<Workflow>(new("Provider not found"));
+}
--- a/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/TriggerIndexing/Fakes/WorkingMaterializer.cs
+++ b/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/TriggerIndexing/Fakes/WorkingMaterializer.cs
@@ -1,0 +1,22 @@
+using Elsa.Workflows.Activities;
+using Elsa.Workflows.Management;
+using Elsa.Workflows.Management.Entities;
+
+namespace Elsa.Workflows.IntegrationTests.Scenarios.TriggerIndexing.Fakes;
+
+/// <summary>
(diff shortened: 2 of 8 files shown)


src/modules/Elsa.Scheduling/Tasks/ResumeWorkflowTask.cs
```diff
@@ -1,38 +1,40 @@
...
+            var runWorkflowInstanceRequest = new RunWorkflowInstanceRequest
+            {
+                Input = request.Input,
+                Properties = request.Properties,
+                ActivityHandle = request.ActivityHandle,
+                BookmarkId = request.BookmarkId
+            };
+            await workflowClient.RunInstanceAsync(runWorkflowInstanceRequest, cancellationToken);
+        }
+        catch (WorkflowInstanceNotFoundException ex)
```
FransVanEk, line 33: i assume there is not a way to find out if the workflow still exists without a exception.

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/BulkDelete/Endpoint.cs
```diff
@@ -19,14 +20,45 @@ public override void Configure()
...
+            DefinitionIds = baseFilter.DefinitionIds,
+            WorkflowStatus = WorkflowStatus.Running
+        };
+
+        var runningInstanceIds = await workflowInstanceStore.FindManyIdsAsync(runningFilter, cancellationToken);
+        var count = 0L;
+
+        foreach (var instanceId in runningInstanceIds)
+        {
+            var client = await workflowRuntime.CreateClientAsync(instanceId, cancellationToken);
```
FransVanEk, line 44: does this client need to be created for each instance or can it be re-used? thinking ahead like the number of instances could get bigger.

src/modules/Elsa.Scheduling/Tasks/ResumeWorkflowTask.cs
sfmskywalker: Technically, we could add a method to check to see if it exists, but that would then always incur 2 DB calls instead of 1. Given that this usually doesn't happen, unless workflow instances have been deleted and background triggers are about to fire, it's better to catch the exception. Alternatively, we could consider adding a method that returns a result object rather than throwing an exception.

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/BulkDelete/Endpoint.cs
sfmskywalker: The workflow client is designed to manage exactly one workflow instance. I think this is fine, because the client is an extremely light object - lighter than your average string, which gets instantiated many more times during workflow execution.

test/integration/Elsa.Common.IntegrationTests/Serialization/ObjectFormatterTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        var result = testList.Format();
+
+        Assert.Equal("[10,20,30]", result);
+    }
+
+    [Fact(DisplayName = "Null returns null")]
+    public void NullReturnsNull()
+    {
+        object? testValue = null;
+        var result = testValue.Format();
```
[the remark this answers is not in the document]
sfmskywalker: Yes, the point is to test the Format extension method to properly deal with null values.

PR #7089, opened 2025-11-21
Refactors extension methods to instance methods

sfmskywalker:
Refactors extension methods in core workflow modules to instance methods.

This improves encapsulation and readability by:
- Reducing the number of extension methods.
- Making the code more object-oriented and easier to understand.
- Improving code discoverability.

- - -
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7089)

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Extensions/ActivityExecutionContextExtensions.InputOutput.cs
+++ b/src/modules/Elsa.Workflows.Core/Extensions/ActivityExecutionContextExtensions.InputOutput.cs
@@ -21,7 +21,7 @@ public static class ActivityExecutionContextExtensions
 
             var outputs = outputDescriptors.ToDictionary(x => x.Name, x =>
             {
-                if (!x.IsSerializable)
+                if (x.IsSerializable == false)
                     return "(not serializable)";
 
                 var cachedValue = activity.GetOutput(expressionExecutionContext, x.Name);


src/modules/Elsa.Workflows.Core/Activities/Flowchart/Extensions/ActivityExecutionContextExtensions.cs
```diff
@@ -43,64 +43,67 @@ public static class ActivityExecutionContextExtensions
...
+        /// </summary>
+        internal bool HasPendingWork()
         {
-            var ownerInstanceId = workItem.Owner?.Id;
+            var flowchart = (Activities.Flowchart)context.Activity;
+            var workflowExecutionContext = context.WorkflowExecutionContext;
+            var activityIds = flowchart.Activities.Select(x => x.Id).ToList();
+            var children = context.Children;
+            var hasRunningActivityInstances = children.Where(x => activityIds.Contains(x.Activity.Id)).Any(x => x.Status == ActivityStatus.Running);
+            var hasUnconsumedTokens = flowchart.GetTokenList(context).Any(x => x is { Consumed: false, Blocked: false });
```
FransVanEk: could this be enhanced.  like { Consumed: false, Blocked: false } like a Static factory method, so weis is more clear and easy to extend if there where changes to the statusses

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Extensions/ActivityExecutionContextExtensions.cs
```diff
@@ -43,64 +43,67 @@ public static class ActivityExecutionContextExtensions
...
+            var hasFaulted = context.HasFaultedChildren();
 
-            if (ownerInstanceId == null)
-                return false;
+            var hasPendingWork = workflowExecutionContext.Scheduler.List().Any(workItem =>
```
FransVanEk: like this method to be moved to the scheduler to be reusable

src/modules/Elsa.Workflows.Core/Extensions/ActivityExecutionContextExtensions.InputOutput.cs
```diff
@@ -5,34 +5,37 @@ namespace Elsa.Workflows;
...
-            if (x.IsSerializable == false)
-                return "(not serializable)";
+            var activity = context.Activity;
+            var expressionExecutionContext = context.ExpressionExecutionContext;
+            var activityDescriptor = context.ActivityDescriptor;
+            var outputDescriptors = activityDescriptor.Outputs;
+
+            var outputs = outputDescriptors.ToDictionary(x => x.Name, x =>
+            {
+                if (x.IsSerializable == false)
```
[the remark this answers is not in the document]
lukhipolito-nexxbiz: I don't like this suggestion, I think `== false` is better because it is easier to read

src/modules/Elsa.Workflows.Core/Extensions/ExpressionExecutionContextExtensions.cs
```diff
@@ -66,309 +66,311 @@ public static IDictionary<object, object> CreateTriggerIndexingPropertiesFrom(Wo
...
-            {
-                var variables = activityExecutionContext.Variables;
 
-                foreach (var variable in variables)
-                    yield return variable;
+            if (context.TryGetWorkflowExecutionContext(out var workflowExecutionContext))
+            {
+                if (workflowExecutionContext.Workflow.ResultVariable != null)
+                    yield return workflowExecutionContext.Workflow.ResultVariable;
             }
```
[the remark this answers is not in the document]
sfmskywalker: I prefer the current format, seems more readable to me.

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Extensions/ActivityExecutionContextExtensions.cs
sfmskywalker: That's a good idea 👍🏻

sfmskywalker: Done

src/modules/Elsa.Workflows.Core/Activities/Flowchart/Extensions/ActivityExecutionContextExtensions.cs
sfmskywalker: Yes, nice. Done 👍🏻

src/modules/Elsa.Workflows.Core/Extensions/ActivityExecutionContextExtensions.InputOutput.cs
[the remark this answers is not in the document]
sfmskywalker: That's true. At the same time, it's very common to use the exclamation mark in this way. Personally, I prefer it, plus it's more consistent with the code base. However, the suggested code change won't compile, because `IsSerializable` is a nullable boolean and the compiler does not support the negation symbol against that.

j03y-nxxbz (commented):
It is quite a large PR with not only refactoring of extension methods it seems. It makes it hard to review. But from what I can see, no strange patterns. So hereby approved

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/BulkDelete/Endpoint.cs
FransVanEk, line 31: does this removes the running ones first and then in bulk remove them from the persistence layer? 

and the count is in memory deletes plus  the persistence ones?

src/modules/Elsa.Workflows.Core/Extensions/WorkflowExtensions.cs
```diff
@@ -8,19 +8,22 @@ namespace Elsa.Extensions;
...
+    extension(Workflow workflow)
     {
-        if (workflow.CreatedWithModernTooling())
-            modernToolingAction();
-        else
-            legacyToolingAction();
+        /// <summary>
+        /// Returns a boolean indicating whether the workflow was created with modern tooling.
+        /// </summary>
+        public bool CreatedWithModernTooling() => workflow.WorkflowMetadata.ToolVersion?.Major >= 3;
```
FransVanEk, line 16: how can this be hardcode deep inside of the code.... why not centralize it ?

src/modules/Elsa.Workflows.Core/Extensions/WorkflowExecutionContextExtensions.cs
```diff
@@ -12,132 +12,133 @@ namespace Elsa.Extensions;
...
 
-    public static object? GetOutputByActivityId(this WorkflowExecutionContext workflowExecutionContext, string activityId, string? outputName = null)
-    {
-        var outputRegister = workflowExecutionContext.GetActivityOutputRegister();
-        return outputRegister.FindOutputByActivityId(activityId, outputName);
-    }
+        /// <summary>
+        /// Returns true if all activities have completed or canceled, false otherwise.
+        /// </summary>
+        public bool AllActivitiesCompleted() => workflowExecutionContext.ActivityExecutionContexts.All(x => x.IsCompleted);
```
FransVanEk, line 119: i was wondering if this is correct, have completed or canceled.... where is the cancel filter, or is canceled also completed?

src/modules/Elsa.Workflows.Core/Extensions/OutputExtensions.cs
```diff
@@ -18,52 +18,58 @@ public static class OutputExtensions
...
-    /// Gets the target type of the specified variable type, if any, linked to the output.
-    /// </summary>
-    public static Type? GetTargetType(this Output? output, ActivityExecutionContext context)
+    extension<T>(Output<T>? output)
     {
-        var memoryBlockReference = output?.MemoryBlockReference();
-        
-        if (memoryBlockReference is null)
-            return null;
+        /// <summary>
```
FransVanEk, line 23: removing the this , will cause a lot of breaking changes when people used the extensions... is this what we want?

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowInstances/BulkDelete/Endpoint.cs
sfmskywalker: Yes, exactly like that.

src/modules/Elsa.Workflows.Core/Extensions/WorkflowExtensions.cs
sfmskywalker: I think this method was added to introduce backwards compatibility when some change was introduced. I don't like this strategy and so for me it's OK to be hidden away.

src/modules/Elsa.Workflows.Core/Extensions/WorkflowExecutionContextExtensions.cs
sfmskywalker: It's correct. The `IsCompleted` is a computed property:

```csharp
public bool IsCompleted => Status is ActivityStatus.Completed or ActivityStatus.Canceled;
```

src/modules/Elsa.Workflows.Core/Extensions/OutputExtensions.cs
sfmskywalker: Removing `this` will not break anyone, because the static methods are nested within the new `extension` keyword. This cleans up the code elegantly without breaking the public interface.

More details about this new keyword: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods

PR #7093, opened 2025-11-21
Adds Publish Event Activity Tests

sfmskywalker:
This change is [Reviewable](https://reviewable.io/reviews/elsa-workflows/elsa-core/7093)


test/unit/Elsa.Activities.UnitTests/Event/PublishEventTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        var publisher = Substitute.For<IEventPublisher>();
+
+        // Act
+        await ExecuteAsync(activity, publisher);
+
+        // Assert
+        await publisher.Received(1).PublishAsync(
+            eventName,
+            null,
+            Arg.Any<string?>(),
```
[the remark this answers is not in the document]
sfmskywalker: Great catch! You're absolutely right, this is a legitimate issue.

The current implementation only normalizes empty strings (`""`), not whitespace. This is inconsistent behavior that could lead to bugs when users accidentally pass whitespace-only correlation IDs.

I'll implement your suggested fix:

1. Add a NullIfWhiteSpace() extension method to StringExtensions
2. Update PublishEvent.ExecuteAsync to use it instead of NullIfEmpty()
3. Add test coverage for the whitespace case

This ensures whitespace correlation IDs are properly normalized to null, making the behavior more intuitive and consistent.

test/component/Elsa.Workflows.ComponentTests/Scenarios/Activities/Primitives/Event/PublishEventTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    [Fact]
+    public async Task PublishEvent_LocalEvent_CompletesWorkflow()
+    {
+        // Act
+        var result = await _workflowRunner.RunAndAwaitWorkflowCompletionAsync(WorkflowDefinitionHandle.ByDefinitionId(PublishAndConsumeEventWorkflow.DefinitionId, VersionOptions.Published));
+
+        // Assert
+        Assert.Equal(WorkflowSubStatus.Finished, result.WorkflowExecutionContext.SubStatus);
```
j03y-nxxbz, line 39: Should there also be an Assert.Equal(WorkflowStatus.Finished, result.WorkflowExecutionContext.Status)?

test/component/Elsa.Workflows.ComponentTests/Scenarios/Activities/Primitives/Event/PublishEventTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    [Fact]
+    public async Task PublishEvent_WithPayload_CompletesSuccessfully()
+    {
+        // Act
+        var result = await _workflowRunner.RunAndAwaitWorkflowCompletionAsync(
+            WorkflowDefinitionHandle.ByDefinitionId(PublishGlobalEventWorkflow.DefinitionId, VersionOptions.Published));
+
+        // Assert
+        Assert.Equal(WorkflowSubStatus.Finished, result.WorkflowExecutionContext.SubStatus);
```
j03y-nxxbz: Same here as my previous comment?

j03y-nxxbz: In addition, should we do any assert on the Payload? I could only deduce a Payload exists when looking at the workflow definition itself, but I am not seeing what we expect to happen with the payload.

test/component/Elsa.Workflows.ComponentTests/Scenarios/Activities/Primitives/Event/PublishEventTests.cs
sfmskywalker: Not necessarily - the main status is derived from the sub status. That logic itself, however, would be good to have covered via some other unit tests (it may or may not be there already).

test/component/Elsa.Workflows.ComponentTests/Scenarios/Activities/Primitives/Event/PublishEventTests.cs
sfmskywalker: That's a great idea, yes we should definitely have a test case that asserts a sent payload is received.

sfmskywalker: Done 👍🏻

PR #7199, opened 2026-01-20
Bugfix: handle uncaught exception thrown in ReadSyntheticInputs

j03y-nxxbz:
Direct cause: typeName was missing in synthetic input property
Root cause: worklfow used as activity defined input with reserved name 'Metadata'

Why fix was needed? Because the uncaught exception was preventing other workflows from being published. Even though this exception occurs, it should not prevent publishing other workflows.

Solution introduced here:
- Collect exceptions that prevent an activity from being in the expected state. In this case, we expected all syntehtic input properties to be valid, but they are not.
- Return the list of exceptions to the caller, along with an IActivity instance that does not throw exceptions
- The caller, in this case ActivityJsonConverter, can log any exceptions, whilst not breaking the loop with uncaught exceptions.

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Serialization/Converters/ActivityJsonConverter.cs
+++ b/src/modules/Elsa.Workflows.Core/Serialization/Converters/ActivityJsonConverter.cs
@@ -39,7 +39,6 @@ public class ActivityJsonConverter(
         }
 
         var clonedOptions = GetClonedOptions(options);
-        
         // If the activity type is not found, create a NotFoundActivity instead.
         if (activityDescriptor == null)
         {


src/modules/Elsa.Dsl.ElsaScript/Compiler/ElsaScriptCompiler.cs
```diff
@@ -202,8 +202,9 @@ private async Task<IActivity> CompileActivityInvocationAsync(ActivityInvocationN
         else
         {
             // No positional arguments, use default constructor
-            var activityConstructorContext = new ActivityConstructorContext(activityDescriptor, ActivityActivator.Create);
-            activity = activityDescriptor.Constructor(activityConstructorContext);
+            var activityConstructorContext = new ActivityConstructorContext(activityDescriptor, (t) => new(ActivityActivator.Create(t)));
+            var activityConstructionResult = activityDescriptor.Constructor(activityConstructorContext);
```
FransVanEk, line 206: what i don't understand is why we have this intermediate result and we don't actually do anything with properties like has HasExceptions etc. 

although i agree with the setup , it feels some logic on error handling is missing here

src/modules/Elsa.Workflows.Core/Serialization/Converters/ActivityJsonConverter.cs
```diff
@@ -38,13 +39,15 @@ public override IActivity Read(ref Utf8JsonReader reader, Type typeToConvert, Js
         }
 
         var clonedOptions = GetClonedOptions(options);
-
+
```
sfmskywalker:
```suggestion

```

src/modules/Elsa.Dsl.ElsaScript/Compiler/ElsaScriptCompiler.cs
sfmskywalker: Actually, the exceptions are being logged in one of the callers.

PR #7311, opened 2026-02-19
Distributed locks to prevent multiple pods executing Reload and Refresh

lukhipolito-nexxbiz:
## Purpose

Addition of Distributed locks to prevent multiple pods from triggering Reload or Refresh at the same time.

---

## Scope

Select one primary concern:
- [X] New feature

---

## Description
Distributed lock with generic lock has been added to `IWorkflowDefinitionsReloader` interface as a decorator. The goal is to prevent multiple pods from ever executing this action at the same time.

The same was done for `IWorkflowDefinitionsRefresher`, with the exception that the key to the distributed lock uses a combination of definition Ids from the request to allow it to be called by different pods with with different Requests.

### Problem
Double or multiple triggering of change notifications and updates/publishing of workflows.

### Solution
Distributed locks ensure no race between pods and only one action that should be singleton accross pods is taken.

### DIFF
--- a/src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
+++ b/src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
@@ -14,44 +14,33 @@ public class DistributedWorkflowDefinitionsRefresher(IWorkflowDefinitionsRefresh
     IDistributedLockProvider distributedLockProvider,
     ILogger<DistributedWorkflowDefinitionsRefresher> logger) : IWorkflowDefinitionsRefresher
 {
-    
-    
     /// <summary>
     /// This ensures that only one instance of the application can refresh a set of workflow definitions at a time, preventing potential conflicts and ensuring consistency across distributed environments.
     /// </summary>
     public async Task<RefreshWorkflowDefinitionsResponse> RefreshWorkflowDefinitionsAsync(RefreshWorkflowDefinitionsRequest request, CancellationToken cancellationToken = default)
     {
-        if (request.DefinitionIds == null 
-            || request.DefinitionIds.Count == 0)
+        var isRefreshingAll = request.DefinitionIds == null || request.DefinitionIds.Count == 0;
+        var lockKey = isRefreshingAll
+            ? "WorkflowDefinitionsRefresher:All"
+            : $"WorkflowDefinitionsRefresher:{string.Join(",", request.DefinitionIds!.OrderBy(x => x))}";
+
+        await using var distributedLock = await distributedLockProvider.TryAcquireLockAsync(
+            lockKey,
+            TimeSpan.Zero,
+            cancellationToken);
+
+        if (distributedLock == null)
         {
-            const string refreshAllLockKey = "WorkflowDefinitionsRefresher:All";
-            await using var distributedAllLock = await distributedLockProvider.TryAcquireLockAsync(
-                refreshAllLockKey,
-                TimeSpan.Zero,
-                cancellationToken);
-            
-            if (distributedAllLock == null)
-            {
-                logger.LogInformation("Could not acquire lock for refreshing all workflow definitions. Another instance is already refreshing all workflow definitions");
-                return new RefreshWorkflowDefinitionsResponse(Array.Empty<string>(), Array.Empty<string>(), RefreshWorkflowDefinitionsStatus.AlreadyInProgress);
-            }
+            var logMessage = isRefreshingAll
+                ? "Could not acquire lock for refreshing all workflow definitions. Another instance is already refreshing all workflow definitions"
+                : "Could not acquire lock for refreshing workflow definitions. Another instance is already refreshing these workflow definitions";
+
+            logger.LogInformation(logMessage);
+
+            var failedDefinitionIds = isRefreshingAll ? Array.Empty<string>() : request.DefinitionIds!;
+            return new(Array.Empty<string>(), failedDefinitionIds, RefreshWorkflowDefinitionsStatus.AlreadyInProgress);
         }
-        else
-        {
-            var definitionIdsKey = string.Join(",", request.DefinitionIds!.OrderBy(x => x));
-            var lockKey = $"WorkflowDefinitionsRefresher:{definitionIdsKey}";
-            await using var distributedLock = await distributedLockProvider.TryAcquireLockAsync(
-                lockKey,
-                TimeSpan.Zero,
-                cancellationToken);
-            
-            if (distributedLock == null)
-            {
-                logger.LogInformation("Could not acquire lock for refreshing workflow definitions. Another instance is already refreshing these workflow definitions");
-                return new RefreshWorkflowDefinitionsResponse(Array.Empty<string>(), request.DefinitionIds!, RefreshWorkflowDefinitionsStatus.AlreadyInProgress);
-            }
-        }
-        
+
         return await inner.RefreshWorkflowDefinitionsAsync(request, cancellationToken);
     }
 }


src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <param name="request"></param>
+    /// <param name="cancellationToken"></param>
+    /// <returns></returns>
+    public async Task<RefreshWorkflowDefinitionsResponse> RefreshWorkflowDefinitionsAsync(RefreshWorkflowDefinitionsRequest request, CancellationToken cancellationToken = default)
+    {
+        if (request.DefinitionIds == null 
+            || request.DefinitionIds?.Count < 1)
+        {
+            return new RefreshWorkflowDefinitionsResponse(Array.Empty<string>(), Array.Empty<string>());
+        }
```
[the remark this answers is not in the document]
lukhipolito-nexxbiz: That's not right. I want to make sure to skip the execution, that is the reason why there is a distributed lock in place. This refreshing should only take place once per pod per list of definition ids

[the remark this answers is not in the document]
lukhipolito-nexxbiz: already answered above. This comment does not make sense. Distributed locks are supposed to prevent execution

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsRefresherDistributedLocking.cs
```diff
@@ -0,0 +1,45 @@
...
+[UsedImplicitly]
+public class WorkflowDefinitionsRefresherDistributedLocking(IWorkflowDefinitionsRefresher inner,
+    IDistributedLockProvider distributedLockProvider) : IWorkflowDefinitionsRefresher
+{
+    /// <summary>
+    /// This ensures that only one instance of the application can refresh a set of workflow definitions at a time, preventing potential conflicts and ensuring consistency across distributed environments.
+    /// </summary>
+    /// <param name="request"></param>
+    /// <param name="cancellationToken"></param>
+    /// <returns></returns>
```
sfmskywalker: Either provide content for XML comment elements, or remove empty ones.

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsRefresherDistributedLocking.cs
```diff
@@ -0,0 +1,45 @@
...
+using JetBrains.Annotations;
+using Medallion.Threading;
+
+namespace Elsa.Workflows.Runtime.Distributed;
+
+/// <summary>
+/// Decorator class that adds Distributed Locking to the Workflow Definitions Refresher.
+/// </summary>
+/// <param name="inner"></param>
+/// <param name="distributedLockProvider"></param>
```
sfmskywalker: Either provide content for XML comment elements, or remove empty ones.

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsRefresherDistributedLocking.cs
```diff
@@ -0,0 +1,45 @@
...
+[UsedImplicitly]
+public class WorkflowDefinitionsRefresherDistributedLocking(IWorkflowDefinitionsRefresher inner,
```
sfmskywalker: Please rename to `DistributedWorkflowDefinitionsRefresher`.


```suggestion
public class DistributedWorkflowDefinitionsRefresher(IWorkflowDefinitionsRefresher inner,
```

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsRefresherDistributedLocking.cs
```diff
@@ -0,0 +1,45 @@
+using Elsa.Workflows.Runtime.Requests;
+using Elsa.Workflows.Runtime.Responses;
+using JetBrains.Annotations;
+using Medallion.Threading;
+
+namespace Elsa.Workflows.Runtime.Distributed;
+
+/// <summary>
+/// Decorator class that adds Distributed Locking to the Workflow Definitions Refresher.
```
sfmskywalker:
```suggestion
/// Decorator class that adds distributed locking to the Workflow Definitions Refresher.
```

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
sfmskywalker, line 1: This file should be moved to the `Elsa.Workflows.Runtime.Distributed project.

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsReloader.cs
sfmskywalker, line 1: This file should be moved to the `Elsa.Workflows.Runtime.Distributed project.

src/modules/Elsa.Workflows.Runtime/Features/WorkflowRuntimeFeature.cs
```diff
@@ -264,8 +265,15 @@ public override void Apply()
             .AddScoped<IActivityExecutionMapper, DefaultActivityExecutionMapper>()
             .AddScoped<IWorkflowDefinitionStorePopulator, DefaultWorkflowDefinitionStorePopulator>()
             .AddScoped<IRegistriesPopulator, DefaultRegistriesPopulator>()
+            
             .AddScoped<IWorkflowDefinitionsRefresher, WorkflowDefinitionsRefresher>()
+            // Decorating with distributed locking to prevent multiple instances from refreshing the same workflow definitions at the same time.
```
sfmskywalker: The comment doesn't provide that much value - it's pretty clear what service is being registered. Even more so once this is moved to the Elsa.Workflows.Runtime.Distributed project, where this is implied from context.

src/modules/Elsa.Workflows.Runtime/Features/WorkflowRuntimeFeature.cs
```diff
@@ -264,8 +265,15 @@ public override void Apply()
...
+            .Decorate<IWorkflowDefinitionsRefresher, WorkflowDefinitionsRefresherDistributedLocking>()
+            
             .AddScoped<IWorkflowDefinitionsReloader, WorkflowDefinitionsReloader>()
+            // Decorating with distributed locking to prevent multiple instances from reloading workflow definitions at the same time.
```
sfmskywalker: The comment doesn't provide that much value - it's pretty clear what service is being registered. Even more so once this is moved to the Elsa.Workflows.Runtime.Distributed project, where this is implied from context.

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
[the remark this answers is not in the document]
sfmskywalker: That's a valid concern. Yes, we want only one pod to do the actual work, but the current design is problematic because to @greptile's point, the current code makes it look like the operation succeeded when it actually did nothing, which could lead to issues down the road.

It might be better if `RefreshWorkflowDefinitionsResponse` would carry a status code of some kind, perhaps an enum, with values like:

- Success
- Skipped

Perhaps, with a reason field explaining why it was skipped.

Still, this is not entirely pure either, because the decorated service itself has no notion of distributed locking, let alone the "Skipped" enum value.

Let's give this some more thought.

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsReloaderDistributedLocking.cs
```diff
@@ -0,0 +1,37 @@
+using JetBrains.Annotations;
+using Medallion.Threading;
+
+namespace Elsa.Workflows.Runtime.Distributed;
+
+/// <summary>
+/// Decorator class that adds Distributed Locking to the Workflow Definitions Reloader.
+/// </summary>
+/// <param name="inner"></param>
+/// <param name="distributedLockProvider"></param>
```
sfmskywalker: Empty XML comments should be removed or populated with a meaningful description.

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsReloaderDistributedLocking.cs
```diff
@@ -0,0 +1,37 @@
+using JetBrains.Annotations;
+using Medallion.Threading;
+
+namespace Elsa.Workflows.Runtime.Distributed;
+
+/// <summary>
+/// Decorator class that adds Distributed Locking to the Workflow Definitions Reloader.
```
sfmskywalker:
```suggestion
/// Decorator class that adds distributed locking to the Workflow Definitions Reloader.
```

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsRefresherDistributedLocking.cs
```diff
@@ -0,0 +1,45 @@
...
+        if (request.DefinitionIds == null 
+            || request.DefinitionIds.Count < 1)
+        {
+            return new RefreshWorkflowDefinitionsResponse(Array.Empty<string>(), Array.Empty<string>());
+        }
+        var definitionIdsKey = string.Join(",", request.DefinitionIds!.OrderBy(x => x));
+        var lockKey = $"WorkflowDefinitionsRefresher:{definitionIdsKey}";
+        await using var distributedLock = await distributedLockProvider.TryAcquireLockAsync(
+            lockKey,
+            TimeSpan.FromMinutes(1),
```
sfmskywalker: We should not wait for one minute before giving up on acquiring a lock, because we want to try and acquire a lock right now - if that doesn't succeed, it mean another replica has the lock, therefore we can just exit immediately.


```suggestion
            TimeSpan.Zero,
```

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsReloaderDistributedLocking.cs
```diff
@@ -0,0 +1,37 @@
...
+    
+    /// <summary>
+    /// This ensures that only one instance of the application can reload workflow definitions at a time, preventing potential conflicts and ensuring consistency across distributed environments.
+    /// </summary>
+    /// <param name="cancellationToken"></param>
+    public async Task ReloadWorkflowDefinitionsAsync(CancellationToken cancellationToken = default)
+    {
+        await using var distributedLock = await distributedLockProvider.TryAcquireLockAsync(
+            LockKey,
+            TimeSpan.FromMinutes(1),
```
sfmskywalker: We should not wait for one minute before giving up on acquiring a lock, because we want to try and acquire a lock right now - if that doesn't succeed, it mean another replica has the lock, therefore we can just exit immediately.


```suggestion
            TimeSpan.Zero,
```

src/modules/Elsa.Workflows.Runtime/Distributed/WorkflowDefinitionsReloaderDistributedLocking.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Elsa.Workflows.Runtime.Distributed;
+
+/// <summary>
+/// Decorator class that adds Distributed Locking to the Workflow Definitions Reloader.
+/// </summary>
+/// <param name="inner"></param>
+/// <param name="distributedLockProvider"></param>
+[UsedImplicitly]
+public class WorkflowDefinitionsReloaderDistributedLocking(
```
sfmskywalker: Please rename to `DistributedWorkflowDefinitionsReloader`

```suggestion
public class DistributedWorkflowDefinitionsReloader(
```

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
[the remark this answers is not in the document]
sfmskywalker: @lukhipolito-nexxbiz Why was this comment marked as resolved?

src/modules/Elsa.Workflows.Runtime/Elsa.Workflows.Runtime.csproj
```diff
@@ -23,4 +23,8 @@
         <ProjectReference Include="..\Elsa.Workflows.Management\Elsa.Workflows.Management.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Folder Include="Distributed\" />
+    </ItemGroup>
```
sfmskywalker: Can be removed.

src/modules/Elsa.Workflows.Runtime/Features/WorkflowRuntimeFeature.cs
```diff
@@ -267,12 +267,10 @@ public override void Apply()
             .AddScoped<IRegistriesPopulator, DefaultRegistriesPopulator>()
             
             .AddScoped<IWorkflowDefinitionsRefresher, WorkflowDefinitionsRefresher>()
-            // Decorating with distributed locking to prevent multiple instances from refreshing the same workflow definitions at the same time.
-            .Decorate<IWorkflowDefinitionsRefresher, WorkflowDefinitionsRefresherDistributedLocking>()
+            .Decorate<IWorkflowDefinitionsRefresher, DistributedWorkflowDefinitionsRefresher>()
```
sfmskywalker: Given that the distributed services have moved to the Elsa.Workflows.Runtime.Distributed module, it should be _that_ module that registers the distributed decorators, not this one.

src/modules/Elsa.Workflows.Runtime/Features/WorkflowRuntimeFeature.cs
```diff
@@ -267,12 +267,10 @@ public override void Apply()
...
             
             .AddScoped<IWorkflowDefinitionsReloader, WorkflowDefinitionsReloader>()
-            // Decorating with distributed locking to prevent multiple instances from reloading workflow definitions at the same time.
-            .Decorate<IWorkflowDefinitionsReloader, WorkflowDefinitionsReloaderDistributedLocking>()
+            .Decorate<IWorkflowDefinitionsReloader, DistributedWorkflowDefinitionsReloader>()
```
sfmskywalker: Given that the distributed services have moved to the Elsa.Workflows.Runtime.Distributed module, it should be that module that registers the distributed decorators, not this one.

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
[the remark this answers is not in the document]
sfmskywalker: I marked the comment as unresolved just now to make sure it's visible.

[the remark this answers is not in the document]
lukhipolito-nexxbiz: @sfmskywalker I believe if we use a separate key for empty or null definition Ids we can ensure that only one instance refreshes all.
In case key not found we could throw an exception, so the other instances skip the action.

What do you think?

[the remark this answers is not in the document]
sfmskywalker: That's not the issue. The issue is that of service API design; the method signature produces a response that will contain an empty list. If a caller were to receive this, they would interpret this as: "no workflows have been refreshed". And although this would be true, there would be no clear way of knowing why, because as far as the call site is concerned, it is working with an IWorkflowRefresher, and is used correctly when passing in a set of definitions to refresh - yet mysteriously enough, somehow, non of them are being refreshed.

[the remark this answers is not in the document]
lukhipolito-nexxbiz: I updated the PR with a commit that addresses the issue. Take a look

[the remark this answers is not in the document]
sfmskywalker: I think the cleanest path forward is to enrich the response type rather than throw exceptions.

The issue with the exception approach is that it leaks infrastructure concerns across a decorator boundary. `IWorkflowDefinitionsRefresher` is a domain-level abstraction; callers should not have to anticipate an `InvalidOperationException` caused by distributed lock contention. If we throw here, every caller either needs to know about this exception (breaking decorator transparency) or risk it bubbling up as an unhandled error. Neither is desirable.

Instead, I propose extending `RefreshWorkflowDefinitionsResponse` with a status:

```csharp
public enum RefreshWorkflowDefinitionsStatus
{
    Completed,
    AlreadyInProgress
}

public record RefreshWorkflowDefinitionsResponse(
    ICollection<string> Refreshed,
    ICollection<string> NotFound,
    RefreshWorkflowDefinitionsStatus Status = RefreshWorkflowDefinitionsStatus.Completed);
````

The decorator returns `AlreadyInProgress` when it cannot acquire the lock. The inner service always returns `Completed` and remains completely unaware of distributed locking.

At the API layer, `AlreadyInProgress` maps cleanly to `202 Accepted`, signaling: "the request was valid; the work is already underway elsewhere in the cluster."

This preserves decorator transparency, provides an explicit and actionable signal to callers, and avoids the silent-empty-success behavior that triggered this discussion. `AlreadyInProgress` is not a distributed-lock detail leaking into the domain, but it is a legitimate outcome in any concurrent system.

Could you update the PR accordingly?

File: src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs [hunk 16]
```diff
@@ -0,0 +1,50 @@
...
+        if (request.DefinitionIds == null 
+            || request.DefinitionIds.Count < 1)
+        {
+            const string refreshAllLockKey = "WorkflowDefinitionsRefresher:All";
+            await using var distributedAllLock = await distributedLockProvider.TryAcquireLockAsync(
+                refreshAllLockKey,
+                TimeSpan.Zero,
+                cancellationToken);
+
+            return distributedAllLock != null
```
j03y-nxxbz: I am confused, was it not agreed in the whole comments to not throw exceptions, instead return an enum AlreadyInProgress instead?

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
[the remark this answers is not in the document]
lukhipolito-nexxbiz: The PR has been updated. 
A status enum was added to the response of worflow refresher, and proper logic changes have been made to account for that in both the decorator and api caller

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowDefinitions/Refresh/Endpoint.cs
```diff
@@ -20,7 +20,14 @@ public override void Configure()
...
     {
         var result = await RefreshWorkflowDefinitionsAsync(request.DefinitionIds, cancellationToken);
-        await Send.OkAsync(new Response(result.Refreshed, result.NotFound), cancellationToken);
+        if (result.Status == RefreshWorkflowDefinitionsStatus.Completed)
+        {
+            await Send.OkAsync(new Response(result.Refreshed, result.NotFound), cancellationToken);
+        }
+        else
+        {
+            await Send.AcceptedAtAsync<Refresh>(responseBody: new Response(result.Refreshed, result.NotFound), cancellation: cancellationToken);
```
sfmskywalker, line 29: Not sure, but will this not redirect to some non-existing endpoint?

j03y-nxxbz (approved):

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    
+    
+    /// <summary>
+    /// This ensures that only one instance of the application can refresh a set of workflow definitions at a time, preventing potential conflicts and ensuring consistency across distributed environments.
+    /// </summary>
+    public async Task<RefreshWorkflowDefinitionsResponse> RefreshWorkflowDefinitionsAsync(RefreshWorkflowDefinitionsRequest request, CancellationToken cancellationToken = default)
+    {
+        if (request.DefinitionIds == null 
+            || request.DefinitionIds.Count < 1)
```
sfmskywalker:
```suggestion
            || request.DefinitionIds.Count == 0)
```

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
```diff
@@ -0,0 +1,57 @@
...
+            await using var distributedLock = await distributedLockProvider.TryAcquireLockAsync(
+                lockKey,
+                TimeSpan.Zero,
+                cancellationToken);
+            
+            if (distributedLock == null)
+            {
+                logger.LogInformation("Could not acquire lock for refreshing workflow definitions. Another instance is already refreshing these workflow definitions");
+                return new RefreshWorkflowDefinitionsResponse(Array.Empty<string>(), request.DefinitionIds!, RefreshWorkflowDefinitionsStatus.AlreadyInProgress);
+            }
```
sfmskywalker: If you first build the lock key, then this entire section can be de-duplicated.

src/modules/Elsa.Workflows.Api/Endpoints/WorkflowDefinitions/Refresh/Endpoint.cs
sfmskywalker: I just tested this, and what happens is that the response includes a Location header with a URL to itself (this endpoint). So, this is fine.

src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedWorkflowDefinitionsRefresher.cs
lukhipolito-nexxbiz: Altought similar, the sections are still different enough to be their own branch

lukhipolito-nexxbiz: nevermind I saw your refactor. Good one!

Issue #7738, opened 2026-06-19
[Discussion] Restore ability to publish workflows with empty/disabled scheduling triggers (Cron/Timer) — disable triggers properly instead of via empty required values


sfmskywalker:
## Context

In **Elsa 3.6**, required activity-input validation was tightened so that publishing a workflow now fails when a required property has an invalid value. A concrete consequence: a `Cron` activity with an **empty `CronExpression`** can no longer be published.

Before 3.6, an empty `CronExpression` was allowed at publish time. It (correctly) produced no working trigger, but it let users **publish the workflow anyway** — and some users relied on this as an informal way to **temporarily disable a Cron/Timer trigger** without deleting the activity.

A customer upgrading to 3.6 has hit this: workflows they could previously publish with a blank cron now fail to publish. This issue is to decide how we want to handle "disabled / not-yet-configured" triggers properly, rather than relying on empty values as a sentinel.

## How it works today (3.6, confirmed in code)

- `WorkflowDefinitionPublisher.PublishAsync` hard-gates on validation — **any** `WorkflowValidationError` blocks publish; there is no warning tier:
  ```csharp
  var validationErrors = (await workflowValidator.ValidateAsync(workflowGraph.Workflow, ct)).ToList();
  if (validationErrors.Any())
      return new(false, validationErrors, new([]));
  ```
- `TriggerIndexer` indexes a trigger for every activity where `Activity.GetCanStartWorkflow() && Activity is ITrigger`.
- `Cron.GetTriggerPayload` returns `new CronTriggerPayload(cronExpression)` even when the expression is blank.
- `CronTriggerPayloadValidator` parses it via Cronos; a blank/invalid expression throws `CronFormatException` → a `WorkflowValidationError` → **publish blocked**.

Relevant files:
- `src/modules/Elsa.Workflows.Management/Services/WorkflowDefinitionPublisher.cs`
- `src/modules/Elsa.Workflows.Runtime/Services/TriggerIndexer.cs`
- `src/modules/Elsa.Workflows.Runtime/Handlers/ValidateWorkflowRequestHandler.cs`
- `src/modules/Elsa.Scheduling/Activities/Cron.cs`
- `src/modules/Elsa.Scheduling/TriggerPayloadValidators/CronTriggerPayloadValidator.cs`

## Key insight: a disable switch already exists

[...]

Issue #7033, opened 2025-11-05
Couldn't store trigger exception


AnKa0:
## Description
We are unable to publish any workflow with a cron trigger in it, in case the cron expression is filled, because the following exception is thrown (stack trace attached below):
Quartz.JobPersistenceException: Couldn't store trigger 'Default.cf146dda767734a9' for 'Default.RunWorkflowJob' job: The job (Default.RunWorkflowJob) referenced by the trigger does not exist.

Adding a Timer activity and running the workflow results in the same exception.

## Steps to Reproduce
1. Create a new workflow with a cron trigger. 
2. Fill in the cron expression.
3. Press save and publish.

**Attachments**:
   - **Workflow JSON**: If your issue is reproducible with a specific workflow, please attach the workflow's JSON file. This will help us understand the exact process that leads to the problem.

[cron trigger](https://github.com/user-attachments/files/23357784/workflow-2.json)

[timer](https://github.com/user-attachments/files/23455127/workflow-1.json)

**Reproduction Rate**: Indicate how often the bug occurs when following these steps (e.g., "every time", "intermittently: about 50% of the time").
It had not been present before, but it suddenly became present in both the test environment and on development machines. It had been tested on a different, fully new, empty database as well - the issue is present that way too.

**Video/Screenshots**

<img width="1830" height="1597" alt="Image" src="https://github.com/user-attachments/assets/b0e62d28-d968-4e8b-a854-9631392dd236" />

## Expected Behavior
The workflow is published.

## Actual Behavior
The workflow is not published, because an exception stops the process.
<img width="1956" height="626" alt="Image" src="https://github.com/user-attachments/assets/ef5c5950-49bc-45e0-8e74-2cb18197b4ef" />

## Environment
- **Elsa Package Version**: 3.5.1
- **Operating System**: Windows 11, Linux

## Log Output
Quartz.JobPersistenceException: Couldn't store trigger 'Default.cf146dda767734a9' for 'Default.RunWorkflowJob' job: The job (Default.RunWorkflowJob) referenced by the trigger does not exist.
 ---> Quartz.JobPersistenceException: The job (Default.RunWorkflowJob) referenced by the trigger does not exist.
[...]

PR #7739, opened 2026-06-19
Allow publishing workflows with a blank/disabled Cron trigger

sfmskywalker:
## Summary

Restores the ability to **publish** a workflow that contains a `Cron` activity with an **empty `CronExpression`**, which regressed in 3.6 when required-input validation was tightened. Some users relied on blanking the expression to **temporarily disable** a Cron trigger without deleting the activity; since 3.6 that workflow fails to publish (`CronFormatException` from Cronos → `WorkflowValidationError` → publish blocked).

This is the **layer-1 / immediate-unblock** change discussed in #7738. It is intentionally narrow: a *blank* expression means "no trigger" (disabled), while a *non-blank but malformed* expression still fails validation (the genuinely valuable part of the 3.6 check).

## What changed

- **`Cron.GetTriggerPayloads`** — returns **no payloads** when `CronExpression` resolves to null/whitespace, so an unconfigured Cron produces no trigger (instead of a malformed `CronTriggerPayload("")`).
- **`TriggerIndexer`** — now distinguishes an *intentionally empty* payload set (→ index **no** trigger) from a *payload-generation failure* (→ keep today's null-payload fallback so genuine failures still surface as "Trigger should have a payload" during validation). Previously both collapsed to a fabricated null payload, which turned an unconfigured Cron into a hard publish error.
- **`CronTriggerPayloadValidator`** — skips validation for a blank expression (defense-in-depth + self-documenting).

Why this is important beyond publish: a blank `CronTriggerPayload` that slipped through would later be loaded by `UpdateTenantSchedules` and handed to the scheduler, throwing Cronos at schedule time (the #7033 family). Not indexing the trigger in the first place avoids that.

`Timer` is unaffected (non-nullable `TimeSpan` with a default, no validator).

## Tests

- `CronTriggerPayloadValidatorTests` (unit) — blank/whitespace/null → no errors; malformed → error; valid → no errors.
- `EmptyCronTriggerTests` (integration) — a blank Cron indexes **no** triggers; a valid Cron indexes exactly **one** with the expected payload.

Full runs green: `Elsa.Scheduling.UnitTests` (51), `Elsa.Workflows.Runtime.UnitTests` (27), `Scenarios/TriggerIndexing` (5, incl. pre-existing delete tests).

## Scope / follow-ups (see #7738)

This unblocks the customer now. The broader discussion — a first-class **disable-trigger** toggle (`CanStartWorkflow` already exists) and an optional explicit **publish-with-warnings** severity tier — is tracked in #7738 for a later minor. The separate Quartz `RunWorkflowJob` registration error (#7033) is **not** addressed here (different repo/subsystem).

Relates to #7738.

### DIFF
--- a/src/modules/Elsa.Scheduling/Activities/Cron.cs
+++ b/src/modules/Elsa.Scheduling/Activities/Cron.cs
@@ -53,10 +53,17 @@ public class Cron : EventGenerator
     }
 
     /// <inheritdoc />
-    protected override object GetTriggerPayload(TriggerIndexingContext context)
+    protected override IEnumerable<object> GetTriggerPayloads(TriggerIndexingContext context)
     {
-        var cronExpression = context.ExpressionExecutionContext.Get(CronExpression)!;
-        return new CronTriggerPayload(cronExpression);
+        // Treat a blank CRON expression as "no trigger" so the workflow can still be published
+        // (e.g. to temporarily disable the schedule) instead of failing validation. A non-blank but
+        // malformed expression still produces a payload and is caught by CronTriggerPayloadValidator.
+        var cronExpression = context.ExpressionExecutionContext.Get(CronExpression);
+
+        if (string.IsNullOrWhiteSpace(cronExpression))
+            return [];
+
+        return [new CronTriggerPayload(cronExpression)];
     }
 
     /// <summary>
--- a/src/modules/Elsa.Scheduling/Activities/Cron.cs
+++ b/src/modules/Elsa.Scheduling/Activities/Cron.cs
@@ -44,8 +44,17 @@ public class Cron : EventGenerator
             return;
         }
         
+        var cronExpression = context.ExpressionExecutionContext.Get(CronExpression);
+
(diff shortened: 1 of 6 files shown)


sfmskywalker:
Added a follow-up commit hardening the **runtime** path to match the publish/index-time behavior.

`Cron.ExecuteAsync` re-evaluates `CronExpression` when the activity runs **mid-workflow** (not as the workflow's start trigger, which short-circuits via `IsTriggerOfWorkflow()`). If the expression — e.g. a JavaScript expression — resolves to an empty/whitespace string at runtime, `ICronParser.GetNextOccurrence("")` previously threw a `CronFormatException`. It now completes the activity without scheduling, so "blank = disabled" is consistent across publish, trigger scheduling, and execution.

Note: for *triggers*, the cron value is frozen at index/publish time into the stored `CronTriggerPayload` and `DefaultTriggerScheduler` already guards blank/format — so this only affects the in-body execution path. Covered by `BlankCronExecutionTests` (blank and whitespace).

Issue #7397, opened 2026-04-16
Workflows sometimes don't resume after child/sub-workflows finish in distributed environment.


cobrafast:
## Description
DispatchWorkflow with WaitForCompletion = true intermittently leaves the parent workflow suspended forever after the child completes. No error is thrown; the parent simply never resumes.

## Expected Behavior
The parent workflow resumes after the child workflow reaches Finished status.

## Actual Behavior
The parent remains in Suspended state indefinitely. The child shows Finished. No faulted state, no log error.

## Environment
- Elsa Package Version: 3.x (distributed runtime, UseDistributedRuntime())
- Deployment: Multiple application instances (Azure App Service, ≥2 instances)
- Persistence: EF Core / SQL Server for workflow instances, bookmark store, and bookmark queue
- Distributed lock: SqlDistributedSynchronizationProvider (Medallion.Threading.SqlServer)
- Distributed cache/dispatch: MassTransit + Azure Service Bus

## Preliminary Cause Analysis

The failure is a race between DistributedBookmarkQueueWorker, BookmarkQueueSignaler, and DefaultBookmarkQueuePurger:

1. Child completes on Node B. ResumeDispatchWorkflowActivity.HandleAsync enqueues a BookmarkQueueItem into the DB, then calls bookmarkQueueSignaler.TriggerAsync() — which writes to Node B's in-memory Channel<T> ([BookmarkQueueSignaler.cs:27](https://github.com/elsa-workflows/elsa-core/blob/main/src/modules/Elsa.Workflows.Runtime/Services/BookmarkQueueSignaler.cs#L27)). Node A's channel is never written to.

2. Node B's DistributedBookmarkQueueWorker wakes and calls TryAcquireLockAsync with TimeSpan.Zero ([DistributedBookmarkQueueWorker.cs:15](https://github.com/elsa-workflows/elsa-core/blob/main/src/modules/Elsa.Workflows.Runtime.Distributed/Services/DistributedBookmarkQueueWorker.cs#L15)). If Node A holds the lock (e.g. its periodic trigger just fired), handle is null and the method returns immediately without processing — no retry, no re-signal.

[...]

PR #7748, opened 2026-06-21
Fix workflow commit atomicity and bookmark queue retries

### LINKED ISSUE #5961: [BUG] Workflow instance state persistence timeout after bookmark resumption leaves instance state showing suspended, yet further activities in workflow have executed
## Description
When a bookmark is resumed, the instance state is persisted after, if the call to persist the state times out the workflow is continued but updates are not persisted, this leaves the workflow instance state in a previous state (suspended) whilst some new activities have been executed.

## Steps to Reproduce
1. Create a workflow with a trigger of some kind and add in the Delay activity
2. Include a custom activity that incorporates a bookmark and a resume
3. Trigger the workflow in rapid succession (we triggered it 50 times and 4 times the instance remained suspended)

1. **Detailed Steps**: Provide a step-by-step description of what leads to the bug. Be as specific as possible.

4. **Code Snippets**: If the issue involves code (e.g., JavaScript error, server request failure), include the relevant snippets where the issue occurs.

5. **Attachments**:
- Stack Trace
```
Exception:*System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.AggregateException: One or more errors occurred. (An error occurred while saving the entity changes. See the inner exception for details.)
 ---> Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while saving the entity changes. See the inner exception for details.
 ---> Microsoft.Data.SqlClient.SqlException (0x80131904): Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.
 ---> System.ComponentModel.Win32Exception (258): Unknown error 258
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location ---
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.SqlServer.Update.Internal.SqlServerModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(IList`1 entriesToSave, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(StateManager stateManager, Boolean acceptAllChangesOnSuccess, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.SqlServer.Storage.Internal.SqlServerExecutionStrategy.ExecuteAsync[TState,TResult](TState state, Func`4 operation, Func`4 verifySucceeded, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync(Boolean acceptAllChangesOnSuccess, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync(Boolean acceptAllChangesOnSuccess, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Extensions.QueryableExtensions.BulkUpsertAsync[TDbContext,TEntity](TDbContext dbContext, IList`1 entities, Expression`1 keySelector, CancellationToken cancellationToken)

at Elsa.EntityFrameworkCore.Common.Store`2.SaveManyAsync(IEnumerable`1 entities, Expression`1 keySelector, Func`4 onSaving, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Common.Store`2.SaveManyAsync(IEnumerable`1 entities, Expression`1 keySelector, Func`4 onSaving, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Common.EntityStore`2.SaveManyAsync(IEnumerable`1 entities, Func`4 onSaving, CancellationToken cancellationToken)
   at Elsa.EntityFrameworkCore.Modules.Runtime.EFCoreWorkflowExecutionLogStore.SaveManyAsync(IEnumerable`1 records, CancellationToken cancellationToken)
   at Elsa.Workflows.Runtime.Middleware.Workflows.PersistWorkflowExecutionLogMiddleware.InvokeAsync(WorkflowExecutionContext context)
   at Elsa.Workflows.Runtime.Middleware.Workflows.PersistActivityExecutionLogMiddleware.InvokeAsync(WorkflowExecutionContext context)
   at Elsa.Workflows.Runtime.Middleware.Workflows.PersistBookmarkMiddleware.InvokeAsync(WorkflowExecutionContext context)
   at Elsa.Workflows.Runtime.Middleware.Workflows.ScheduleBackgroundActivitiesMiddleware.InvokeAsync(WorkflowExecutionContext context)
   at Elsa.Workflows.Pipelines.WorkflowExecution.WorkflowExecutionPipeline.ExecuteAsync(WorkflowExecutionContext context)
   at Elsa.Workflows.Services.WorkflowRunner.RunAsync(WorkflowExecutionContext workflowExecutionContext)
   at Elsa.Workflows.Services.WorkflowRunner.RunAsync(Workflow workflow, WorkflowState workflowState, RunWorkflowOptions options, CancellationToken cancellationToken)
   at Elsa.Workflows.Runtime.Services.WorkflowHost.ResumeWorkflowAsync(ResumeWorkflowHostParams params, CancellationToken cancellationToken)
   at Elsa.Workflows.Runtime.Services.DefaultWorkflowRuntime.ResumeWorkflowAsync(String workflowInstanceId, ResumeWorkflowRuntimeParams options)
   at Elsa.Workflows.Runtime.Services.DefaultWorkflowRuntime.ResumeWorkflowAsync(String workflowInstanceId, ResumeWorkflowRuntimeParams options)
   at Elsa.Workflows.Runtime.Handlers.DispatchWorkflowCommandHandler.HandleAsync(DispatchWorkflowInstanceCommand command, CancellationToken cancellationToken)
   at Elsa.Mediator.CommandStrategies.DefaultStrategy.ExecuteAsync[TResult](CommandStrategyContext context)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at InvokeStub_Task`1.get_Result(Object, Object, IntPtr*)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   at Elsa.Mediator.Middleware.Command.Components.CommandHandlerInvokerMiddleware.InvokeAsync(CommandContext context)
   at Elsa.Mediator.Middleware.Command.CommandPipeline.InvokeAsync(CommandContext context)
   at Elsa.Mediator.Services.DefaultMediator.SendAsync(ICommand command, ICommandStrategy strategy, CancellationToken cancellationToken)
   at Elsa.Mediator.HostedServices.BackgroundCommandSenderHostedService.ReadOutputAsync(Channel`1 output, CancellationToken cancellationToken)*
```
6. **Reproduction Rate**: Indicate how often the bug occurs when following these steps (e.g., "every time", "intermittently: about 50% of the time").
- Executing the workflow 50 times in rapid succession will produce this error most of the time

## Expected Behavior
- A workflow is triggered, it is then delayed and resumed with the correct state and status associated with the workflow instance state

## Actual Behavior
- The workflow instance state presents itself as being perpetually suspended, yet further activities have executed.

### LINKED ISSUE #7726: Inconsistency in bookmarks vs workflow instance due to non-atomic operations inside DefaultCommitStateHandler.CommitAsync
## Description
`DefaultCommitStateHandler.CommitAsync` persists workflow state as a sequence of independent, non-transactional await calls, with bookmark persistence first, and the workflow instance (state blob) saved last. If any operation between them fails (e.g. a SQL command timeout under load), the bookmark changes — including the removal of an auto-burned bookmark — are already committed, but the instance state save never happens. This leaves the relational bookmark store and the persisted instance state permanently divergent: the bookmark row is gone, while the (stale) instance blob still shows the activity as armed/suspended.
Because the event dispatcher resolves incoming stimuli against the bookmark store, the missing row means the activity can never be resumed again. The instance stays Running/Suspended forever with no way to recover through normal event delivery.
The relevant method is in `src/modules/Elsa.Workflows.Runtime/.../DefaultCommitStateHandler.cs`
```
public async Task CommitAsync(WorkflowExecutionContext workflowExecutionContext, WorkflowState workflowState, CancellationToken cancellationToken = default)
{
    var updateBookmarksRequest = new UpdateBookmarksRequest(workflowExecutionContext, workflowExecutionContext.BookmarksDiff, workflowExecutionContext.CorrelationId);
    await bookmarkPersister.PersistBookmarksAsync(updateBookmarksRequest);          // (1) bookmark removal committed here
    await activityExecutionLogRecordSink.PersistExecutionLogsAsync(workflowExecutionContext, cancellationToken);   // (2)
    await workflowExecutionLogRecordSink.PersistExecutionLogsAsync(workflowExecutionContext, cancellationToken);   // (3)
    await variablePersistenceManager.SaveVariablesAsync(workflowExecutionContext); // (4)
    var workflowInstance = await workflowInstanceManager.SaveAsync(workflowState, cancellationToken);              // (5) state blob saved last
    workflowExecutionContext.ExecutionLog.Clear();
    workflowExecutionContext.ClearCompletedActivityExecutionContexts();
    await workflowExecutionContext.ExecuteDeferredTasksAsync();
    await mediator.SendAsync(new WorkflowStateCommitted(workflowExecutionContext, workflowState, workflowInstance), cancellationToken);
}
```

**As there is no transaction / unit-of-work wrapping ops 1-to-5, a failure after the first line and before line 5 commits the bookmark mutation without the corresponding state advance.**
The auto-burn that feeds the removal into BookmarksDiff happens earlier, in DefaultActivityInvokerMiddleware.InvokeAsync (in-memory), and is only persisted at step (1) above:
```
var resumedBookmark = workflowExecutionContext.ResumedBookmarkContext?.Bookmark;
if (resumedBookmark is { AutoBurn: true })
    workflowExecutionContext.Bookmarks.Remove(resumedBookmark);
```

## Steps to Reproduce
Have Elsa with EFCore on SQL Server persistence (separate runtime and management stores), distributed runtime enabled, with single or multiple instances.
Run a long-running workflow that suspends on an auto-burn bookmark - any blocking activity with AutoBurn = true, e.g.(in my case) an event-wait activity.
While a resume is being committed, induce a failure on the persistence path after the bookmark write — most easily a SQL command timeout (the writes at steps 2–5 seem to produce this the easiest as the execution-log can be heavy for large workflows). A transient DB slowdown or connection-pool exhaustion under load gives me the reproduction.
The bookmark row is deleted from the bookmark store, but the workflow instance state was not updated (the save at step 5 did not run).
Send the event the workflow was waiting on. It will not resume — there is no bookmark row to match — and the instance remains Running indefinitely.

## Expected Behavior
Bookmark persistence and workflow-instance state persistence within a single `CommitAsync` should be **atomic**. A failure while persisting any part of a committed execution step should roll back the whole step — including the bookmark removal — so the bookmark store and the instance state never diverge. On failure, the resume should remain retriable with the bookmark intact.

## Actual Behavior
The bookmark removal is committed independently and first; a failure on a later write in the same CommitAsync leaves the bookmark deleted and the instance state unchanged. The instance is permanently unresumable via event delivery and stays Running. The workflow execution log shows the activity's last event as Suspended, never Resumed, because the Resumed log entry is only flushed at steps (2)/(3) and is lost when the commit fails — **making the failure hard to diagnose from the journal alone**.

## Screenshots
None, this is not UI related.

## Environment
- Elsa Package Version: 3.6.1 & 3.6.2 :
- Scenario is OS-independent.

## Log Output
The triggering failure is a SQL command timeout on the persistence path, e.g.:
```
System.ComponentModel.Win32Exception (258): Unknown error 258
  at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(...)
  at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute...
```
(258 = Windows WAIT_TIMEOUT.) 
Pretty much any failure in the CommitAsync produces this; a timeout on a heavy log instance writes is the most common trigger. Note the instance's own execution log will not contain a Resumed record for the affected activity, so the integrity loss is silent in the journal.

## Troubleshooting Attempts
1. Confirmed via the detection query that a set of Running instances had fewer live bookmarks than their definition requires (specifically zero auto-burn event-wait bookmarks where two were expected).
2. Decompressed the instance Data blob (Zstd) for affected instances and confirmed the bookmark is still present in the serialized state, while absent from the bookmark table — establishing table/blob divergence rather than a workflow-design issue.
3. Verified from the workflow execution log that the affected activities' last recorded event is Suspended, never Resumed — consistent with a failure early in CommitAsync (after the bookmark write at step 1, before the log flush at steps 2/3).
4. Traced the mechanism through 3.6.1 source: in-memory auto-burn in DefaultActivityInvokerMiddleware → WorkflowExecutionContext.CommitAsync() → DefaultCommitStateHandler.CommitAsync, where bookmark persistence (step 1) and instance save (step 5) are separate, unwrapped awaits.

## Additional Context
Intermittend reproduction rate, we only saw this after a very heavy load, while running with limited DTUs - this caused a cascade of DB write misses (IO Timeouts

## Related Issues
#5961 — Pretty sure this is related: Workflow instance state persistence timeout after bookmark resumption leaves instance state showing suspended, yet further activities have executed. (Same observable failure; this report also adds  the root-cause mechanism in DefaultCommitStateHandler.)
#7397 — Workflows sometimes don't resume after child/sub-workflows finish in distributed environments. (Different code path, bookmark queue purge race,  but similar theme: bookmark deletion not atomic with the operation that depends on it.)

sfmskywalker:
## Summary
- Wrap workflow commit persistence in a provider-safe commit transaction, enabled for SQL Server EF Core persistence only.
- Buffer commit-time mediator notifications until after successful persistence, and defer activity context/log cleanup until commit success.
- Fix bookmark queue paging so deleting processed/dead-lettered rows does not skip later queue items, and re-signal distributed queue processing after transient lock misses.

## Notes
- Generic/non-SQL EF Core persistence keeps the previous no-op transaction behavior to avoid enabling unsafe ambient/distributed transactions for providers that may not support them.
- The default bookmark queue purge TTL is increased to reduce the lock/purge race reported in the related queue issue.

Fixes #7726
Fixes #5961
Addresses #7397

## Validation
- `dotnet test test/unit/Elsa.Workflows.Runtime.UnitTests/Elsa.Workflows.Runtime.UnitTests.csproj --filter "FullyQualifiedName~DefaultCommitStateHandlerTests|FullyQualifiedName~BookmarkQueueDeadLetterTests|FullyQualifiedName~WorkflowCommitNotificationBufferTests" --no-restore`
- `dotnet build src/modules/Elsa.Persistence.EFCore/Elsa.Persistence.EFCore.csproj --no-restore`
- `dotnet build src/modules/Elsa.Persistence.EFCore.SqlServer/Elsa.Persistence.EFCore.SqlServer.csproj --no-restore`


src/modules/Elsa.Workflows.Runtime/Services/StoreActivityExecutionLogSink.cs
```diff
@@ -24,11 +24,6 @@ public async Task PersistExecutionLogsAsync(WorkflowExecutionContext context, Ca
 
         var records = await Task.WhenAll(activityExecutionContexts.Select(x => x.GetOrMapCapturedActivityExecutionRecordAsync()));
         await activityExecutionStore.SaveManyAsync(records, cancellationToken);
-
-        // Untaint activity execution contexts.
-        foreach (var activityExecutionContext in activityExecutionContexts)
-            activityExecutionContext.ClearTaint();
-
```
sfmskywalker, line 31: Why were these lines removed?

test/unit/Elsa.Workflows.Runtime.UnitTests/Services/DefaultCommitStateHandlerTests.cs
sfmskywalker, line 1: There seems to be a lot of repetition in this file. Can we DRY it up?

src/modules/Elsa.Workflows.Runtime/Services/StoreActivityExecutionLogSink.cs
sfmskywalker: Addressed by documenting the moved responsibility in StoreActivityExecutionLogSink: taint is now cleared by DefaultCommitStateHandler only after the commit transaction succeeds, so a rollback keeps dirty contexts retryable.

test/unit/Elsa.Workflows.Runtime.UnitTests/Services/DefaultCommitStateHandlerTests.cs
sfmskywalker: Addressed by refactoring DefaultCommitStateHandlerTests around a shared CommitTestFixture, removing the repeated substitute/handler setup while keeping each scenario-specific arrangement explicit.

PR #7756, opened 2026-06-24
fix: correct Oracle identifier quoting and NVARCHAR2 type cast in GenerateOracleUpsert

### LINKED ISSUE #7755: BUG: Oracle persistence fails with ORA-00904 and NVARCHAR2 datatype mismatch in GenerateOracleUpsert
# Bug Report: Oracle persistence broken in `GenerateOracleUpsert` due to unquoted identifiers and NVARCHAR2 type inference failure

## Description

Oracle persistence is completely non-functional when using `BulkUpsertAsync`. Two bugs
in `GenerateOracleUpsert` cause all bulk upsert operations to fail, meaning workflow
instances, bookmarks, activity execution records, triggers, and execution log records
cannot be persisted.

**Bug 1 — Unquoted identifiers (ORA-00904)**

The generated MERGE statement uses unquoted aliases in the `SELECT … FROM DUAL` source:

```sql
SELECT :p0 AS id, :p1 AS activity_id, … FROM DUAL
```

Oracle uppercases unquoted identifiers, so `AS id` defines the alias `ID`. However, the
`ON`, `WHEN MATCHED`, and `WHEN NOT MATCHED` clauses reference `Source."id"` (quoted,
case-sensitive). `ID ≠ "id"` in Oracle, causing: ORA-00904: "id": invalid identifier

The same unquoted identifier problem also exists in the `ON` condition, `UPDATE SET`, and
`INSERT`/`VALUES` clauses — all column references are unquoted, which will fail if any
column name contains a lowercase character (which all of Elsa's EF Core-generated column
names do).

**Bug 2 — NVARCHAR2 bind parameter type inference failure**

After fixing Bug 1, Oracle still fails on NVARCHAR2 columns. In a `SELECT … FROM DUAL`
source subquery, Oracle has no target column to derive bind parameter types from. ODP.NET
defaults to inferring `VARCHAR2` for string parameters. When the target column is
`NVARCHAR2`, Oracle raises a datatype mismatch error during the MERGE.

This does not affect any other provider because SQL Server, PostgreSQL, MySQL, and SQLite
all use patterns (`VALUES`, `INSERT … ON CONFLICT`, `INSERT … ON DUPLICATE KEY`) where
bind parameters sit directly in a row value context and type inference is not required.
`FROM DUAL` is Oracle-specific and uniquely suffers from this limitation.

## Steps to Reproduce

1. Configure Elsa with an Oracle database using EF Core persistence.
2. Dispatch any workflow.
3. Observe ORA-00904 during the first bulk upsert attempt (e.g. `workflow_instances`
   or `activity_execution_records`).
4. After fixing the alias quoting, observe datatype mismatch errors on NVARCHAR2 columns.

**Reproduction rate**: 100% on any Oracle deployment.

## Expected Behavior

Workflow instances, bookmarks, activity execution records, triggers, and execution log
records are persisted correctly via MERGE statements.

## Actual Behavior

All bulk upsert operations fail. Workflow execution cannot proceed past the first
persistence step.

## Environment

- Elsa version: 3.8.0_preview1 (confirmed against current `main`)
- Database: Oracle 23c.
- File affected: `src/modules/Elsa.EntityFrameworkCore/Extensions/BulkUpsertExtensions.cs`

## Proposed Fix

A fix is ready as a pull request. The changes to `GenerateOracleUpsert` are:

1. **Quote all identifiers** — aliases in the `SELECT … FROM DUAL` source and all column
   references in `ON`, `UPDATE SET`, and `INSERT`/`VALUES` are wrapped in double quotes,
   making Oracle treat them as case-sensitive identifiers that match the column names
   created by EF Core migrations.

2. **Explicit CAST for NVARCHAR2 columns** — for properties mapped to `NVARCHAR2(n)`
   columns, the bind parameter is wrapped in `CAST(:p AS NVARCHAR2(n))`. The length `n`
   is extracted from the EF Core column type string (e.g. `"NVARCHAR2(500)"`) so the
   cast matches the actual column definition exactly.

   This is an acknowledged workaround for the `FROM DUAL` type inference limitation in
   ODP.NET. A cleaner long-term fix would be to construct explicit `OracleParameter`
   objects with `OracleDbType.NVarchar2` set, which would allow Oracle to infer types
   correctly without any SQL-level casting. That would require restructuring the method
   to use `DbConnection` directly rather than `ExecuteSqlRawAsync`, which is a larger
   change deferred to a follow-up.

## Additional Context

All other providers (SQL Server, SQLite, PostgreSQL, MySQL) are unaffected. Their upsert
patterns do not use `FROM DUAL` and therefore do not require type inference in the source
subquery.

MohitGuptaC:
Fixes #7755

## Problem

`GenerateOracleUpsert` produced a `MERGE` statement with unquoted schema, table, column, and source-alias identifiers. Oracle folds unquoted identifiers to uppercase, while Elsa Oracle migrations create quoted mixed-case objects such as `"Elsa"."ActivityExecutionRecords"` and `"Id"`. The generated statement therefore did not address the EF-created objects consistently and failed with `ORA-00904`.

A second failure occurs inside the `SELECT ... FROM DUAL` source. ODP.NET has no target column there from which to infer bind parameter types and defaults .NET strings to `VARCHAR2`, while Elsa maps string properties to `NVARCHAR2`. The resulting type mismatch causes the `MERGE` to fail even after identifier quoting is corrected.

## Changes

- Delimit the Oracle schema, table, columns, aliases, and all downstream references consistently through the provider `ISqlGenerationHelper`.
- Cast parameters mapped to `NVARCHAR2` using the EF Core column type, for example `CAST({0} AS NVARCHAR2(450))`.
- Add focused SQL-generation coverage for:
  - mixed-case Oracle schema, table, and column identifiers;
  - `NVARCHAR2(450)` and `NVARCHAR2(2000)` casts, including a null value;
  - non-string columns remaining uncast;
  - multi-row `UNION ALL SELECT` generation;
  - key exclusion from `UPDATE SET`;
  - `INSERT`/`VALUES` clauses and parameter ordering.

Only the Oracle generator behavior is changed; the other provider generators are untouched.

## Verification

- `dotnet test test/unit/Elsa.Persistence.EFCore.UnitTests/Elsa.Persistence.EFCore.UnitTests.csproj --configuration Release` — passed
- `dotnet build src/modules/Elsa.Persistence.EFCore.Oracle/Elsa.Persistence.EFCore.Oracle.csproj --configuration Release` — passed for `net8.0`, `net9.0`, and `net10.0` with zero warnings

The contributor also verified the original fix end-to-end on Oracle: workflow dispatch and persistence succeeded for workflow instances, activity execution records, workflow execution logs, bookmarks, and triggers, with no `ORA-00904` or datatype mismatch errors.

## Release target

The PR remains targeted to `main`. A maintainer previously made `release/3.8.0` conditional on intending the fix for that release, and the contributor explicitly accepted a later release. Retargeting this main-based head directly would also introduce unrelated main commits into the diverged release branch; a dedicated cherry-pick/backport is the safe route if 3.8.0 GA must include the fix.

### DIFF
--- a/Elsa.sln
+++ b/Elsa.sln
@@ -413,6 +413,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Persistence.VNext.Runt
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Hosting.Management.UnitTests", "test\unit\Elsa.Hosting.Management.UnitTests\Elsa.Hosting.Management.UnitTests.csproj", "{39DE4EE7-0FDB-499F-9BC2-7ECC779FA5F6}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Persistence.EFCore.UnitTests", "test\unit\Elsa.Persistence.EFCore.UnitTests\Elsa.Persistence.EFCore.UnitTests.csproj", "{C6EFA89A-923E-4F7A-A63B-BC8DDF44A208}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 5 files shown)


MohitGuptaC:
@dotnet-policy-service agree company="Surya-Fintech"

sfmskywalker:
@greptile review

sfmskywalker:
Thanks for the PR, @MohitGuptaC. We are looking into it.

In any event, if this is intended to ship as part of 3.8.0, the PR should target the `release/3.8.0` branch instead of `main`.

MohitGuptaC:
@sfmskywalker 
I’ll make sure to target the correct branch from next time. Thanks for letting me know.

For this time, I’m okay if it ships later, as long as the code gets merged.

sfmskywalker:
@greptile review

PR #7894, opened 2026-07-26
Update Jint to 4.15.3 and stop blocking the calling thread on promises

lahma:
## Non-blocking promise resolution

```csharp
var result = engine.Evaluate(preparedScript);
return result.UnwrapIfPromise().ToObject();
```

`UnwrapIfPromise` blocks the calling thread while the engine's event loop drains — inside an `async` method. An expression that awaits a .NET `Task`, such as any expression calling `getSecret()`, therefore held a thread pool thread for the duration of that I/O.

`Engine.EvaluateAsync` awaits the returned promise instead, and observes the cancellation token while it does.

## Why 4.15.3 and not 4.14.0

The PR originally moved 4.4.2 → 4.14.0, the release where `EvaluateAsync` arrived. It now lands past 4.15.2 deliberately: once expressions genuinely suspend and resume instead of draining the event loop on the calling thread, they exercise the async suspension machinery 4.15.2 corrected — an `await` on a right-hand side no longer stores the suspension sentinel, async generators and `for await...of` preserve loop iteration state across a suspension, and a suspension node is unwrapped correctly. Shipping the non-blocking change on an earlier 4.14/4.15 would enable exactly the code paths those releases fixed. 4.15.3 is the current release.

## One default changed along the way

Since 4.14, `Options.Interop.ArrayConversion` defaults to `LiveView`: a CLR array reaches script as a live view over the original array rather than as a copy. That is observable, and it showed up in the existing suite —
`Scenarios/JavaScriptListsAndArrays` `Test5` ("Can sort array and list properties as mutable arrays") fails on the bare upgrade with `Cannot convert type 'double[]' to 'object[]'`, because a `double[]` now round-trips back as `double[]` rather than as a copied `object[]`. The same change also means a script calling `.sort()` on an array would reorder the workflow's own array in place.

The evaluator therefore pins the previous `Copy` semantics, so the version bump carries no behavioural change. Hosts that want the live view can opt in:

```csharp
elsa.UseJavaScript(jint => jint.ConfigureEngineOptions(options =>
    options.Interop.ArrayConversion = ArrayConversionMode.LiveView));
```

If you would rather adopt `LiveView` as the new Elsa default, that is a one-line change plus an update to `Test5` — happy to do it either way, but it seemed wrong to fold a semantic change into a version bump.

## Tests

New `ArrayConversionTests` pins the copy semantics from both sides: sorting a CLR array from a script does not mutate the original, and the value round-trips as `object[]`.

A second commit pins the *lane* rather than just its observable product, using the interop conversion counters that arrived in 4.15.1: a CLR array crosses through the copy lane and never the live view, and an ordinary evaluation converts no CLR array at all — Elsa converts collection-valued variables itself in `ObjectConverterHelper` before Jint's array lane could see them, which is what makes the `ArrayConversion` setting a narrow compatibility pin rather than something every evaluation depends on. The same commit adopts `JsString.Create` (public since 4.15.3) in `JsonElementConverter`, where the string case was the only one still routed through the whole `JsValue.FromObject` conversion pipeline to arrive at the same call the number and boolean cases beside it already make directly.

`Elsa.JavaScript.IntegrationTests`: 77 passing.

https://claude.ai/code/session_0179sA2T7HuRfRfSc2JirFik

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -128,7 +128,7 @@
         <PackageVersion Include="Humanizer.Core" Version="3.0.10"/>
         <PackageVersion Include="IronCompress" Version="1.7.0"/>
         <PackageVersion Include="JetBrains.Annotations" Version="2025.2.4"/>
-        <PackageVersion Include="Jint" Version="4.4.2"/>
+        <PackageVersion Include="Jint" Version="4.15.3"/>
         <PackageVersion Include="JsonSchema.Net" Version="9.4.0"/>
         <PackageVersion Include="LinqKit.Core" Version="1.2.11"/>
         <PackageVersion Include="MailKit" Version="4.14.1"/>
--- a/src/modules/Elsa.Expressions.JavaScript/ObjectConverters/JsonElementConverter.cs
+++ b/src/modules/Elsa.Expressions.JavaScript/ObjectConverters/JsonElementConverter.cs
@@ -26,7 +26,11 @@ internal class JsonElementConverter : IObjectConverter
         {
             JsonValueKind.Object => JsValue.FromObject(engine, JsonObject.Create(element)),
             JsonValueKind.Array => JsValue.FromObject(engine, JsonArray.Create(element)),
-            JsonValueKind.String => JsValue.FromObject(engine, element.GetString()),
+            // JsString.Create is the counterpart of the JsNumber.Create and JsBoolean uses below: it produces the
+            // string value directly, where JsValue.FromObject would re-enter the whole conversion pipeline — the
+            // registered object converters, this one included, followed by the default converter's type switch —
+            // to arrive at the same call. It became public in Jint 4.15.3.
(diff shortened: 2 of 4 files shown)


sfmskywalker:
@copilot resolve the merge conflicts in this pull request

lahma:
@sfmskywalker Rebased onto current main and the `Directory.Packages.props` conflict with #7896's dependency upgrades is resolved. While in there, the pin moved 4.14.0 → **4.15.3** (the current release) — the updated PR description explains why the non-blocking promise change wants the 4.15.2 async-suspension fixes underneath it rather than bare 4.14. The array-lane audit commit moved down from #7895 into this PR, since it pins what this PR changes. `Elsa.JavaScript.IntegrationTests` green (77 passing).

sfmskywalker:
Awesome, thanks @lahma !

PR #7901, opened 2026-07-30
fix: restore request services after tenant middleware exceptions

### LINKED ISSUE #7900: TenantResolutionMiddleware leaves HttpContext.RequestServices pointing to a disposed scope after downstream exceptions
## Description

`TenantResolutionMiddleware` temporarily replaces `HttpContext.RequestServices`
with the tenant scope's service provider.

The original provider is restored only after `await next(context)` completes:

```csharp
var originalServiceProvider = context.RequestServices;
context.RequestServices = tenantScope.ServiceProvider;
await next(context);
context.RequestServices = originalServiceProvider;
```

When downstream middleware or an endpoint throws, the restoration statement is
skipped. The `await using` declaration then disposes the tenant scope while
`HttpContext.RequestServices` still references that scope's service provider.

As a result, exception-handling middleware registered before `UseTenants()` can
receive an `HttpContext` whose `RequestServices` points to a disposed provider.
Resolving a service from it throws `ObjectDisposedException`, potentially
masking the original exception or preventing the intended error response.

Affected source:

https://github.com/elsa-workflows/elsa-core/blob/main/src/modules/Elsa.Tenants.AspNetCore/Middleware/TenantResolutionMiddleware.cs#L37-L41

## Steps to Reproduce

1. Enable Elsa multitenancy and register `UseTenants()`.
2. Register exception-handling middleware before `UseTenants()`.
3. Have downstream middleware or an endpoint throw an exception.
4. In the outer exception handler, resolve a service from
   `HttpContext.RequestServices`.
5. Observe that the provider is not the original request provider and has
   already been disposed.

A middleware-level regression test demonstrates the problem directly:

```csharp
var originalRequestServices = context.RequestServices;

await Assert.ThrowsAsync<InvalidOperationException>(
    () => middleware.InvokeAsync(context, tenantResolverPipelineInvoker));

Assert.Same(originalRequestServices, context.RequestServices);
_ = context.RequestServices.GetRequiredService<ScopedProbe>();
```

With the current implementation, the identity assertion fails and resolving a
service from the remaining provider throws `ObjectDisposedException`.

**Reproduction Rate:** Every time downstream throws.

**Additional Configuration:** Multitenancy middleware must be enabled. The
problem also occurs when tenant resolution returns `null`, because the
middleware still creates and installs a tenant scope.

## Expected Behavior

`HttpContext.RequestServices` should always be restored to the original request
provider before the tenant scope is disposed, regardless of whether downstream
execution succeeds, is cancelled, or throws.

Outer exception-handling middleware should be able to use the original request
services and observe the original downstream exception.

## Actual Behavior

After a downstream exception:

```text
Original provider restored: False
Current provider usable: false (ObjectDisposedException)
```

`HttpContext.RequestServices` still references the disposed tenant scope.

## Screenshots

Not applicable.

## Environment

- **Elsa Package Version:** Current `main` source at
  `ec9acd4f3f51b3b73584a4917a51472582beaa4a`
- **.NET SDK:** 10.0.302
- **Operating System:** macOS 26.5.2
- **Browser and Version:** Not applicable
- The defect is not expected to be runtime- or OS-specific.

## Log Output

```text
Assert.Same() Failure: Values are not the same instance
Expected: original request service provider
Actual: tenant scope service provider
```

Resolving a scoped service from the actual provider after the middleware
unwinds throws `ObjectDisposedException`.

## Troubleshooting Attempts

- Compared the middleware with a downstream fork and confirmed that both use
  the same source blob.
- Reproduced the failure using the real `Elsa.Tenants.AspNetCore` project and
  `DefaultTenantScopeFactory`.
- Confirmed that the normal successful path restores the provider; the failure
  is specific to exceptional downstream completion.
- Searched existing open and closed issues and pull requests; no matching
  report was found.

## Additional Context

A focused fix is to restore the provider in a `finally` block:

```csharp
await using var tenantScope = tenantScopeFactory.CreateScope(tenant);
var originalServiceProvider = context.RequestServices;
context.RequestServices = tenantScope.ServiceProvider;

try
{
    await next(context);
}
finally
{
    context.RequestServices = originalServiceProvider;
}
```

The `finally` executes before the `await using` declaration disposes the tenant
scope.

I can submit a focused pull request with this change and a regression test.

## Related Issues

None found.

DenDeline:
## Purpose

Restore the original `HttpContext.RequestServices` when
`TenantResolutionMiddleware` exits through a downstream exception.

Without restoration, outer middleware can receive a request-services provider
whose tenant scope has already been disposed.

Fixes #7900

---

## Scope

- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

### Problem

`TenantResolutionMiddleware` replaces `HttpContext.RequestServices` with the
tenant-scoped service provider before invoking the downstream middleware.

The original provider is restored only after `await next(httpContext)`.
If downstream middleware throws, that restoration statement is skipped. The
tenant scope is then disposed while `HttpContext.RequestServices` still points
to its provider.

An outer exception handler that subsequently resolves a request service can
therefore encounter an `ObjectDisposedException`, potentially obscuring the
original downstream exception.

### Solution

Restore the original request-services provider in a `finally` block. This
ensures restoration happens before the tenant scope is asynchronously disposed,
on both successful and exceptional paths.

A regression test verifies that:

- The original downstream exception propagates unchanged.
- The original request-services provider is restored.
- A scoped service can still be resolved from the restored provider afterward.

This change does not modify public APIs, configuration, routing, or path-base
handling.

---

## Verification

Steps performed:

1. Ran the focused `TenantResolutionMiddleware` regression test.
2. Ran all `Elsa.Tenants.UnitTests`.
3. Built `Elsa.Tenants.AspNetCore` for all configured target frameworks.
4. Ran `git diff --check`.

Results:

- Focused regression test: 1 passed.
- Tenant unit-test suite: 8 passed.
- `Elsa.Tenants.AspNetCore`: builds successfully for .NET 8, .NET 9, and
  .NET 10 with 0 warnings and 0 errors.
- No whitespace errors were reported.

Expected outcome:

When downstream middleware throws, the original request-services provider is
restored before tenant-scope disposal, and the original exception continues to
propagate.

---

## Screenshots / Recordings

Not applicable.

---

## Checklist

- [x] The PR is focused on one concern.
- [x] Commits follow the repository's commit convention.
- [x] Tests cover the changed behavior.
- [x] Documentation is not required because no public API or configuration changed.
- [x] No unrelated cleanup is included.
- [x] Relevant tests pass locally.


DenDeline:
@dotnet-policy-service agree

sfmskywalker (approved):
With a test and everything 🫶

Issue #7929, opened 2026-08-12
BPMN W6: triggers and correlation for message, signal and timer starts


sfmskywalker:
Part of #7909 (W6).

Let a BPMN process start from outside: message and signal start events, and recurring timer starts.

## Mechanism

`BpmnProcess` implements `ITrigger` and returns one bookmark datum per resolvable trigger from `GetTriggerPayloadsAsync(TriggerIndexingContext)`. Walk the graph for event-defined start events and emit a payload per resolved message or signal **name**. Timer starts go through the same path as `Elsa.Timer`'s recurring schedule; the library only supports a recurring start (ISO-8601 interval or cron), and a non-recurring one degrades to a plain start at import.

Elsa 4's equivalents total 138 lines across a stimulus provider and a recurring-schedule provider, so this is small.

## The rule that bites

**A nested `BpmnProcess` must not register start triggers.** An event subprocess body and a subprocess body are both nested scopes, and with D11 a `BpmnProcess` can also sit inside a `Flowchart`. In every one of those cases its start events are internal, not workflow entry points.

Root position is **not recoverable from a published activity node alone**, so this needs an explicit opt-out flag on the activity, set by the binder when it constructs a nested scope and by flowchart composition. Elsa 4 learned this the hard way; do not try to infer it.

## Correlation

Message and signal matching is on the resolved **name**, and error and escalation matching is on the **code**. That is the library's rule, not ours, and it is what makes a `.bpmn` file portable.

sfmskywalker:
Carrying one finding here from #7926 (merged in 65fe688), so it is not lost.

This issue says a nested `BpmnProcess` must not register start triggers, and that the flag is *"set by the binder when it constructs a nested scope and by flowchart composition."* The first half now exists; the second half does not, and there is a gap between them.

**What #7926 delivered:** `BpmnProcess.IsRootScope` reads Elsas existing `CanStartWorkflow` (default `false`) rather than introducing a second flag, so it cannot disagree with the gate `TriggerIndexer` reads. The applier refuses — before applying any command in the batch — when work it is about to start is a `BpmnProcess` that claims root position.

**The gap:** that refusal only sees work bound directly to the enclosing BPMN scope. A `BpmnProcess` nested via an intermediate container — a `Flowchart` inside a `BpmnProcess`, which D11 makes a real shape — never reaches the check, because the applier only ever inspects BPMN-bound work, not a Flowcharts own children.

[...]

PR #7950, opened 2026-08-17
fix(runtime): let a trigger index payloads under per-payload stimulus names

### LINKED ISSUE #7949: An ITrigger cannot register payloads under more than one stimulus name
Found while implementing BPMN start triggers (#7929). Filing separately because the fix belongs in `Elsa.Workflows.Core` / `Elsa.Workflows.Runtime` and should be reviewed on its own merits rather than as a rider on a feature.

## The constraint

`TriggerIndexingContext.TriggerName` is a single settable property:

```csharp
public string TriggerName { get; set; } = trigger.Type;
```

`TriggerIndexer.CreateWorkflowTriggersAsync` reads it **once**, after `GetTriggerPayloadsAsync` has returned every payload, and applies it to all of them:

```csharp
var triggerName = triggerIndexingContext.TriggerName;
...
var triggers = triggerData.Select(payload => new StoredTrigger
{
    Name = triggerName,
    Hash = _hasher.Hash(triggerName, payload),
    Payload = payload
});
```

So one `ITrigger` can only ever register payloads under **one** stimulus name. If an implementation sets `TriggerName` more than once — which the stimulus extension methods do, since `GetEventStimulus` and `GetTimerTriggerStimulus` each set it — the last write wins for every row.

It is not just a mislabelling. `Hash` is computed from the same name, so the earlier payloads are stored under a hash no publisher will ever compute. They are unmatchable.

## Why it surfaced

A BPMN process can legitimately declare several start events of different kinds — say a message start and a recurring timer start. Per #7929, `BpmnProcess` implements `ITrigger` and emits one payload per resolvable start. Reusing the existing payload types is the point: `EventStimulus` so an ordinary `PublishEvent` matches a BPMN message start, `TimerTriggerPayload`/`CronTriggerPayload` so `DefaultTriggerScheduler` schedules a BPMN timer start with no BPMN-specific machinery.

With the current model those cannot coexist on one activity. Whichever kind is processed last claims the name, and the others go unmatchable.

Nothing in the existing trigger set hits this, because each of them emits exactly one stimulus kind — so the constraint has been invisible rather than absent.

## What a fix probably looks like

The important property is that it stays **backwards compatible**. Every existing `ITrigger` should index byte-identically after the change; nothing should have to be rewritten.

That suggests adding a richer path rather than altering the existing one — for example letting an implementation contribute `(name, payload)` pairs, with the current single-`TriggerName` behaviour preserved for anything that does not. Whatever the shape, `Name` and `Hash` need to derive from the name that belongs to **that** payload rather than from one field shared across the call.

Worth deciding deliberately, since it touches how every trigger in the runtime is indexed.

## Also worth a look while in here

`TriggerIndexer.CreateWorkflowTriggersAsync` inserts a `Payload = null` placeholder row for **any** `ITrigger` that returns an empty payload list:

```csharp
if (!triggerData.Any()) triggerData.Add(null!);
```

That row is functionally inert — `WorkflowMatcher` and `TriggerBoundWorkflowService` match purely on `Hash`, and no real stimulus hashes to `(triggerType, null)` — so this is not a live defect, and #7929 verified that independently. But an activity that deliberately declines to register a trigger still leaves a row behind, which is at best confusing when reading the trigger store. Mentioned here rather than filed separately because anyone changing this method should see both.

sfmskywalker:
Lets one `ITrigger` register payloads under more than one stimulus name. **Isolated into its own PR because it changes how every trigger in the runtime is indexed**, and shared-runtime changes deserve review on their own merits rather than as a rider on a feature.

Closes #7949. Unblocks #7929 (BPMN start triggers).

## The defect

`TriggerIndexingContext.TriggerName` is a single settable property, and `TriggerIndexer.CreateWorkflowTriggersAsync` read it **once** after all payloads were collected, then applied it to every row:

```csharp
var triggerName = triggerIndexingContext.TriggerName;
...
Name = triggerName,
Hash = _hasher.Hash(triggerName, payload),
```

So an `ITrigger` could only ever register under one stimulus name. Set it twice — which the stimulus extensions do, since `GetEventStimulus` and `GetTimerTriggerStimulus` each assign it — and the last write won for every row.

That is not just a wrong label. `Hash` derives from the same name, so the earlier payloads were stored under a hash no publisher would ever compute. **Silently unmatchable.**

Nothing in the existing trigger set hit this, because each emits exactly one stimulus kind. The constraint was invisible rather than absent. A BPMN process with a message start *and* a recurring timer start is the first thing that needs it.

## The change

Additive and opt-in. A `NamedTriggerPayload` wrapper returned through the existing `IEnumerable<object>`; the indexer unwraps it and derives `Name`, `Hash` and `Payload` from **one source per row**:

```csharp
var namedPayload = payload as NamedTriggerPayload;
var triggerName = namedPayload != null ? namedPayload.Name : defaultTriggerName;
var stimulus    = namedPayload != null ? namedPayload.Payload : payload;
```

`ITrigger`'s signature is untouched — it gained an XML doc remark only. `TriggerName` keeps working exactly as before for anything that does not opt in.

Written as an explicit ternary rather than `namedPayload?.Payload ?? payload` on purpose: the null-coalescing form would silently store the *wrapper* if an inner payload were ever null.

**Why a wrapper rather than `(name, payload)` pairs on the context**, which the issue suggested: an identity-keyed map fails silently when a trigger registers a name for a payload instance it does not return — you get a row under the wrong name with no signal. The wrapper makes that mis-association structurally impossible and preserves order.

## Backwards compatibility, proven rather than asserted

This is the part that matters for a change at this level. The characterisation tests were run against the **pre-change** `TriggerIndexer` (via `git checkout HEAD -- TriggerIndexer.cs`) and they pass identically there — only the new named-payload tests fail. Same assertions, both sides.

They exercise the real built-in kinds — `Event`, `Timer`, `Cron`, `StartAt`, `HttpEndpoint` — through the real DI-registered `IStimulusHasher` and `ITriggerIndexer`, asserting `Name`, `Hash`, `Payload` and row count per kind. Not mocks.

## Two invariants made structural

**The wrapper is never persisted.** Review found the doc asserted this but only one level was unwrapped, so a double-wrapped payload would reach `StoredTrigger.Payload` — breaking `ValidateWorkflowRequestHandler`, which reflects on `Payload.GetType()` to resolve `ITriggerPayloadValidator<TPayload>`, and `WorkflowTriggerEqualityComparer`, which serialises by runtime type for the definition-refresh diff. The constructor now refuses to nest, so the guarantee holds by construction rather than by documentation.

**A blank stimulus name is refused at construction**, rather than storing a row hashed under an empty name.

## Verification

| Suite | Result |
| --- | --- |
| `dotnet build Elsa.sln` | pass |
| `Elsa.Workflows.Runtime.UnitTests` | 271 |
| `Elsa.Workflows.Core.UnitTests` | 267 |
| `Elsa.Scheduling.UnitTests` | 50 |
| `Elsa.Activities.UnitTests` | 411 |
| `Elsa.Workflows.IntegrationTests` | 289 |
| `Elsa.Bpmn*` (four projects) | 16 / 5 / 23 / 25, unchanged |

Also run and green: `Elsa.Http.UnitTests`, `Elsa.Http.IntegrationTests`, `Elsa.Alterations.IntegrationTests`, `Elsa.Workflows.Management.UnitTests` — all of which contain trigger implementations.

Mutations, each red then restored green:
- per-payload name falls back to the shared field → both multi-name tests fail;
- the nesting guard removed → the nested-wrapper test fails.

## Left alone deliberately

The `Payload = null` placeholder row that gets inserted for any trigger returning an empty list is unchanged, and is now **pinned by characterisation tests** (empty list, and a throwing trigger) so a future change to it is visible.

For the record, since anyone touching that method should know: it is inert for matching — `WorkflowMatcher` and `TriggerBoundWorkflowService` key purely on `Hash`, and nothing hashes to `(triggerType, null)`. But it does make `ValidateWorkflowRequestHandler` emit a "trigger should have a payload" validation error for any activity that legitimately declines to register. That is worth its own issue rather than a rider on this one.

Also noted and unchanged: `Cron`, `StartAt`, `WebhookEventReceived` and `AlterationPlanCompleted` never assign `TriggerName`, so they index under the activity type name and only coincidentally match `SchedulingStimulusNames.Cron`/`StartAt`. That coupling is now pinned by the integration test.

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Contexts/TriggerIndexingContext.cs
+++ b/src/modules/Elsa.Workflows.Core/Contexts/TriggerIndexingContext.cs
@@ -10,6 +10,15 @@ public class TriggerIndexingContext(WorkflowIndexingContext workflowIndexingCont
     public ExpressionExecutionContext ExpressionExecutionContext { get; } = expressionExecutionContext;
     public ITrigger Trigger { get; } = trigger;
     public CancellationToken CancellationToken { get; } = cancellationToken;
+
+    /// <summary>
+    /// The default stimulus name applied to the payloads returned by the trigger being indexed. Defaults to the activity type name.
+    /// </summary>
+    /// <remarks>
+    /// This is a single value shared by every payload of the trigger: when it is assigned more than once, the last write applies to all of them.
+    /// A trigger that needs to register payloads under more than one stimulus name should return <see cref="NamedTriggerPayload"/> instances instead,
+    /// which carry a name per payload and take precedence over this property.
+    /// </remarks>
     public string TriggerName { get; set; } = trigger.Type;
 
     public T? Get<T>(Input<T>? input) => ExpressionExecutionContext.Get(input);
--- a/src/modules/Elsa.Workflows.Core/Contracts/ITrigger.cs
+++ b/src/modules/Elsa.Workflows.Core/Contracts/ITrigger.cs
@@ -1,3 +1,5 @@
+using Elsa.Workflows.Models;
+
 namespace Elsa.Workflows;
 
 /// <summary>
@@ -8,5 +10,9 @@ public interface ITrigger : IActivity
     /// <summary>
     /// Implementors should return a list of objects where each object represents a bookmark datum. For each datum, a trigger is created.
     /// </summary>
+    /// <remarks>
+    /// Each returned object is stored under the stimulus name held by <see cref="TriggerIndexingContext.TriggerName"/>, which is shared by all payloads.
+    /// To register payloads under more than one stimulus name, return <see cref="NamedTriggerPayload"/> instances: each carries the name for its own payload.
+    /// </remarks>
     ValueTask<IEnumerable<object>> GetTriggerPayloadsAsync(TriggerIndexingContext context);
 }
--- a/src/modules/Elsa.Workflows.Core/Models/NamedTriggerPayload.cs
+++ b/src/modules/Elsa.Workflows.Core/Models/NamedTriggerPayload.cs
@@ -1,0 +1,36 @@
+namespace Elsa.Workflows.Models;
+
+/// <summary>
+/// A trigger payload that carries its own stimulus name.
+/// </summary>
+/// <remarks>
+/// Return instances of this type from <see cref="ITrigger.GetTriggerPayloadsAsync"/> when a single trigger needs to register payloads under
+/// more than one stimulus name. The wrapper is unwrapped during indexing: the stored trigger's name and hash are derived from <see cref="Name"/>
+/// and its payload from <see cref="Payload"/>. Payloads that are not wrapped are named using <see cref="TriggerIndexingContext.TriggerName"/>.
+/// </remarks>
+public record NamedTriggerPayload
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="NamedTriggerPayload"/> class.
+    /// </summary>
+    /// <param name="name">The stimulus name to register the payload under.</param>
+    /// <param name="payload">The payload to register.</param>
+    public NamedTriggerPayload(string name, object payload)
+    {
+        if (string.IsNullOrWhiteSpace(name))
+            throw new ArgumentException("A stimulus name is required.", nameof(name));
+
+        Name = name;
+        Payload = payload;
+    }
+
+    /// <summary>
+    /// The stimulus name to register the payload under. Publishers of this stimulus compute their hash from this same name.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// The payload to register. This is what gets stored; the wrapper itself never is.
+    /// </summary>
+    public object Payload { get; }
+}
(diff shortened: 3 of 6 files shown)


sfmskywalker:
Recording a CI flake rather than letting it pass silently.

The first `ubuntu-latest` run failed on `Elsa.Shells.Api.Tests.Endpoints.Reload.ReloadTests.Post_WhenBlueprintUnavailable_Returns503WithFailedStatus`:

```
System.InvalidOperationException : Endpoint HTTP: POST /shells/{shellId}/reload contains authorization
metadata, but a middleware was not found that supports authorization.
```

Nothing about trigger indexing touches the shells API middleware pipeline, so I checked rather than assumed:

- **On clean `origin/main` (7328a0f)**: `dotnet test test/unit/Elsa.Shells.Api.Tests` → 6/6 pass. So not a pre-existing breakage on main.
- **On this branch locally**: 6/6 pass. So not caused by this diff either.
- **Re-ran the failed job**: green, and `Elsa.Shells.Api.Tests` reports 6/6 in the passing run.

Passing on both sides locally and failing once in CI is a flake signature. The error is a test-host pipeline condition that can depend on ordering or parallelism within the run, and the failure landed during a GitHub partial outage window.

Flagging it because a green re-run is exactly where a real intermittent bug gets buried. I have not chased it further — it is unrelated to this change and reproduces nowhere I can drive it — but if `Elsa.Shells.Api.Tests` fails this way again on an unrelated PR, this comment is the second data point.

Issue #7982, opened 2026-08-24
Retire the legacy permission constant classes and finish the External Authentication catalog migration


sfmskywalker:
Follow-up to the authorization model rework (#7974 / #7980). Everything here is cleanup or completion work that was deliberately left out of that PR to keep it reviewable — none of it blocks the model, and none of it is a correctness bug.

Each module now declares its resources in a `<Module>ResourcePermissions` class. The old `<Module>Permissions` classes holding `verb:resource` strings were left in place so the cutover stayed additive. What remains of them falls into three tiers.

## Tier 1 — Dead constants, safe to delete

Zero references anywhere in `src/` or `test/`:

- `Elsa.Dashboard.Api/Permissions/DashboardPermissions.cs`
- `Elsa.Diagnostics.OpenTelemetry/Permissions/OpenTelemetryPermissions.cs`
- `Elsa.Secrets/Permissions/SecretsPermissions.cs`

Note `SecretsPermissions` also declares `use:`, `import:` and `export:secrets`, which never guarded an endpoint even before this work.

**Caveat:** these are `public const` in shipped packages, so deleting them is a source-breaking change for anyone referencing them. Either mark `[Obsolete]` now and remove at the next major, or remove outright if the team is comfortable — worth an explicit decision rather than a silent delete.

## Tier 2 — Referenced only by naming tests

- `Elsa.Diagnostics.ConsoleLogs/Permissions/ConsoleLogsPermissions.cs` — referenced by `ConsoleLogsNamingTests` and `ConsoleLogsModuleTests`
- `Elsa.Diagnostics.StructuredLogs/Permissions/StructuredLogsPermissions.cs` — referenced by `StructuredLogsNamingTests`

Those tests assert the constant still equals its legacy string, which is true and was left alone deliberately during the cutover — twice, in fact, after an over-eager find-and-replace broke them. They are now the *only* thing keeping these classes alive.

The tests are not worthless: they pin a naming convention. The question for whoever picks this up is whether that convention should now be asserted against the **new** resource names instead, at which point the legacy class and its test both go.

## Tier 3 — External Authentication: still genuinely in use

`Elsa.ExternalAuthentication/Permissions/ExternalAuthenticationPermissions.cs` has 14 constants across 6 files, and **is not dead**. It is used for two different things:

[...]

PR #7985, opened 2026-08-25
fix(external-auth)!: wildcard-aware permission grant boundary, and startup smoke tests for both hosts

sfmskywalker:
Follow-up to #7980, taking the first two items of #7982 in the order that issue proposes.

## 1 — The grant boundary was wildcard-blind, and it was exploitable

`PermissionGrantBoundary` held its allow and deny lists in a `HashSet<string>` with `StringComparer.Ordinal`. Under the new `{resource}:{verb}` vocabulary that cannot match a pattern, so the deployment deny list did not hold.

The issue flags one direction. **Both** are reachable:

| Deny list | Grant | Old result |
|---|---|---|
| `workflows/*:delete` | `workflows/definitions:delete` | admitted |
| `workflows/definitions:delete` | `workflows/*:delete` | admitted |

The second is the more surprising one: a wildcard grant outflanks a deny spelled out by name, so a fully concrete deny list is no protection either.

**Why this is exploitable rather than cosmetic.** `ElsaRolePermissionGrantSource` passes a role's permissions to the boundary verbatim; survivors are stored on the session as `ExternalGrants`, and `DefaultExternalAuthenticationTokenIssuer` concatenates them into the issued token's permission claims, where `PermissionEvaluator` *does* expand wildcards. So an ordinary role plus a deny list is enough, on every external sign-in, with no privileged actor and no connection-management access involved. `elsa-roles` is in the default `AllowedPermissionGrantSourceTypes`.

This was verified rather than argued: restoring the ordinal boundary under the new tests fails seven of them.

Deny is now matched in **both** directions and allow **one**-directionally, through `PermissionMatcher`. Allow stays one-directional deliberately — an allow entry must cover the whole grant, so a grant broader than anything allowed is refused rather than admitted for the overlapping part. A grant that is not a well-formed permission is dropped with a `malformed_permission` warning instead of being carried into a token it could never authorize anything in.

### A second problem, not in the issue

Five checks outside endpoint authorization still compared claim values against the **legacy** `ExternalAuthenticationPermissions` constants: delegation, role-reference removal, unsafe-settings confirmation, the final-login-path recovery override, and the boundary itself.

Those constants carry two colons (`external-authentication:connections:read`), which `Permission.TryParse` rejects outright — so no principal can hold one. Meanwhile `doc/migrations/authorization-model.md` instructs operators to replace exactly those strings. **Following the migration guide silently disables all five paths.** All now route through `IPermissionEvaluator`, so one matching rule governs the whole module.

Also here: the module consumed `IPermissionEvaluator` without registering it, relying on host ordering. `AddExternalAuthenticationServices` now calls `AddElsaAuthorization()` itself. To be precise about what that fixes — both hosts already register it independently, so this is defensive for consumers wiring the module directly, not a live host bug.

Non-core verbs move into `ExternalAuthenticationVerbs`, declared beside the resources they apply to, so a delegation check cannot spell a verb differently from the endpoint it guards.

The migration guide gains a section on the new matching semantics, including a warning that a carried-over deny list may now deny more than it did — which is the intent.

## 2 — Startup smoke tests for both hosts

Two parallel feature systems, every module registering in both, and nothing exercising either. The suites construct services directly, so a module registered in one path and not the other passes every test and fails only at startup. Three bugs in #7980 were found by running these hosts by hand.

Each host boots through `WebApplicationFactory` running its real `Program`, then is asked for routes it should serve behind a permission. The status code names the failure instead of collapsing it:

- **404** — the module was never registered (the classic-vs-shell divergence)
- **5xx** — endpoint found, dependencies not constructible (invisible to any test that builds services by hand)
- **200** — reachable without credentials, so no gate ran

**The gate is proven, not merely written.** Removing `AddExternalAuthenticationServices` from the shell feature fails the shell host on all five routes while the classic host stays green.

Assertions go through HTTP rather than the container on purpose. The hosts have genuinely different topologies: the classic root provider holds everything and registers **125** routes, while CShells gives each shell its own provider and mounts routes per shell, leaving **6** in the root. Any container or route-table assertion would have to encode that difference and would break whenever CShells changed internally.

Each host gains a namespaced entry-point marker, because both already declare a `Program` in the global namespace and a test project referencing both cannot tell them apart.

Coverage is off for that project: it references both hosts, so every module either pulls in would enter its denominator without adding real coverage, and coverlet cannot instrument a graph that size (`Unable to read beyond the end of the stream`). `TreatAsLocalProperty` keeps CI's `/p:CollectCoverage=true` from overriding it — verified against CI's exact Release invocation.

## Third commit: unrelated IDE cleanup

`34d64cbe9` carries a solution-wide IDE cleanup — redundant qualifiers and usings removed, primary-constructor and record syntax applied — across ConsoleLogs, StructuredLogs, Expressions.JavaScript and Identity. It ran alongside this work and is separated so the permission changes review on their own. It is behaviour-neutral and all touched projects build clean.

## Breaking

Marked `!` because a deployment carrying a deny list will now be denied more than before. That is the fix, but it is a behaviour change worth a release note.

## Verification

- External Authentication unit tests: **168 passed**
- External Authentication integration tests: **136 passed**
- Host smoke tests: **4 passed**, and confirmed failing when the divergence is injected
- Both hosts booted by hand: start clean, gated routes return 401

The clean run also surfaced a real bug an earlier contended run had masked — `ConnectionManagementTests` hand-registers its dependencies rather than calling `AddExternalAuthenticationServices`, so it needed the evaluator registered the way a host does. Fixed in the first commit.

## Not included

Item 3 of #7982, extending the endpoint coverage gate, is deliberately left out. Items 4 (legacy constants, duplicate descriptor types) and the untested migration-guide walkthrough remain open.

Refs #7982

### DIFF
--- a/Elsa.sln
+++ b/Elsa.sln
@@ -514,6 +514,7 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.UserTasks.Persistence.EFCore.UnitTests", "test\unit\Elsa.UserTasks.Persistence.EFCore.UnitTests\Elsa.UserTasks.Persistence.EFCore.UnitTests.csproj", "{A4F3CB08-759D-4FE6-B715-E8D91E3E3F3E}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.UserTasks.Persistence.ConformanceTests", "test\unit\Elsa.UserTasks.Persistence.ConformanceTests\Elsa.UserTasks.Persistence.ConformanceTests.csproj", "{0C16ED4D-2483-488D-9CA1-D269ED5BEB9F}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Elsa.Hosts.SmokeTests", "test\integration\Elsa.Hosts.SmokeTests\Elsa.Hosts.SmokeTests.csproj", "{0BE4BC01-D02D-4185-A433-E33780584261}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -2455,6 +2456,18 @@ Global
 		{0C16ED4D-2483-488D-9CA1-D269ED5BEB9F}.Release|x64.Build.0 = Release|Any CPU
 		{0C16ED4D-2483-488D-9CA1-D269ED5BEB9F}.Release|x86.ActiveCfg = Release|Any CPU
 		{0C16ED4D-2483-488D-9CA1-D269ED5BEB9F}.Release|x86.Build.0 = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|x64.Build.0 = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Debug|x86.Build.0 = Debug|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|Any CPU.Build.0 = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|x64.ActiveCfg = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|x64.Build.0 = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|x86.ActiveCfg = Release|Any CPU
+		{0BE4BC01-D02D-4185-A433-E33780584261}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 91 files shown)


sfmskywalker:
@greptile review

sfmskywalker:
### Runtime evidence for the malformed allow-list path

T-Rex could not verify this because `dotnet` is unavailable in the review sandbox (exit 127). Supplying the run it attempted, so the finding can be closed on evidence rather than left open on a tooling gap.

The guard is at `DefaultPermissionGrantResolver.cs:110`. `_isUnusable` is set when either list loses entries to parsing, and `Allows` returns `false` before any other check:

```csharp
_isUnusable = _allowed.Count != options.AllowedPermissions.Count || _denied.Count != options.DeniedPermissions.Count;

public bool Allows(string permission)
{
    if (_isUnusable || !Permission.TryParse(permission, out var candidate))
        return false;
```

So `_allowed.Count == 0` is no longer reachable with a non-empty configured list — an allow list that parses to nothing admits nothing, rather than reading as unrestricted.

Exactly the test T-Rex tried to run, plus the startup validation, on `net10.0`:

```
Passed  PermissionGrantPipelineTests.AGrantBoundaryThatDoesNotParseAdmitsNothing(allowed: ["not a permission"], denied: [])
Passed  PermissionGrantPipelineTests.AGrantBoundaryThatDoesNotParseAdmitsNothing(allowed: ["workflows/*:delete", "external-authentication:connections:read"], denied: [])
Passed  PermissionGrantPipelineTests.AGrantBoundaryThatDoesNotParseAdmitsNothing(allowed: [], denied: ["external-authentication:connections:read"])
Passed  ExternalAuthenticationOptionsValidatorTests.RejectsAGrantBoundaryEntryThatIsNotAWellFormedPermission(permission: "external-authentication:connections:read")
Passed  ExternalAuthenticationOptionsValidatorTests.RejectsAGrantBoundaryEntryThatIsNotAWellFormedPermission(permission: "not a permission")
Passed  ExternalAuthenticationOptionsValidatorTests.RejectsAGrantBoundaryEntryThatIsNotAWellFormedPermission(permission: "workflows/definitions:")
Passed  ExternalAuthenticationOptionsValidatorTests.AcceptsAGrantBoundaryOfWildcardPatterns
```

The three malformed cases cover all-malformed, partially-malformed, and a malformed deny entry — the deny mirror of the same bug, where an unparseable entry silently stops denying what it names.

Two layers, not one. `ExternalAuthenticationOptionsValidator` rejects a malformed boundary at startup, so the runtime guard is only reachable if validation was bypassed. That is what makes it safe for the runtime behaviour to be strict.

Full suites on this commit: **175** unit, **136** integration, **4** host smoke — all passing.

The two inline comments on this file are the original finding re-posted against shifted line numbers; line 110 is the guard itself and line 130 is `Parse`. Resolving them on that basis.

sfmskywalker:
@greptile review

sfmskywalker:
@greptile review

sfmskywalker:
Re-requesting review: the previous run scored this **5/5** and closed both boundary findings on executed checks. The only change since is `Elsa.sln`, resolving an add/add conflict with #7986 — both projects kept, solution verified with \`dotnet sln list\`.

Every one of the 90 source, test and doc files this PR touches is byte-identical to the commit that scored 5/5:

```
git diff dd018a4b7 HEAD -- $(git diff --name-only origin/main...HEAD | grep -v '^Elsa.sln$')   # empty
```

The 1/5 that followed was not a new finding — it reported the same two paths as unverified because T-Rex could not start (`dotnet: not found`), while noting "the current implementation contains protections for both paths".

Locally on this exact commit, `net10.0`:

- `Elsa.ExternalAuthentication.UnitTests` — **175 passed**
- `Elsa.ExternalAuthentication.IntegrationTests` — **138 passed**
- `Elsa.Hosts.SmokeTests` — **4 passed** (both hosts boot, gated routes return 401)

The two paths in question are covered by `AGrantBoundaryThatDoesNotParseAdmitsNothing` (3 cases), `RejectsAGrantBoundaryEntryThatIsNotAWellFormedPermission` (3 cases), and `DeploymentDenyBoundaryAppliesToRoleDerivedPermissionsAtIssuance` plus `RolePermissionsAreUnaffectedWhenNoBoundaryIsConfigured`, which pins that the default configuration is untouched.

@greptile review

PR #7987, opened 2026-08-25
refactor(auth)!: retire the legacy permission constants and duplicate descriptor types

sfmskywalker:
Closes the cleanup half of #7982, following #7985.

## The legacy `<Module>Permissions` classes are removed

Seven classes holding `verb:resource` strings: `AIPermissions`, `ConsoleLogsPermissions`, `DashboardPermissions`, `ExternalAuthenticationPermissions`, `OpenTelemetryPermissions`, `SecretsPermissions`, `StructuredLogsPermissions`.

`AIPermissions` was not in #7982's list — that list was written before the cutover finished, and it is dead by the same measure. Classes their own modules still reference (`WorkflowPermissions`, `IdentityPermissions`, and others) are untouched.

**Removed rather than `[Obsolete]`**, which #7982 asked to be an explicit decision rather than a silent one. The reasoning: every string these classes held carries two colons, so `Permission.TryParse` rejects it and no principal can hold it. Marking them obsolete would leave code that compiles, still reads as a permission check, and silently authorizes nothing — a suppressible warning in front of an invisible runtime failure. A compile error names the call site and is fixable against the migration guide's mapping table. Documented in the guide.

## The duplicate descriptor system is collapsed — and it was hiding a bug

Removed: the module's own `PermissionDescriptor` record, its `IPermissionDescriptorProvider` and `IPermissionDescriptorRegistry`, and `DefaultPermissionDescriptorRegistry`.

This was not only tidiness. **The module's registry was fed exclusively by its legacy permission names.** After the cutover, `DefaultPermissionGrantResolver` compared every grant against a catalog containing only unparseable legacy strings, so `unknown_permission_descriptor` fired for every *correct* grant — a permanent warning on healthy configuration.

The resolver now consults the core catalog, which is keyed by resource and lists the verbs each accepts. A wildcard counts as advertised, since it names a pattern rather than a resource to look up. New `UnknownDescriptorWarningTracksTheCoreCatalog` covers matched resource+verb, an undeclared verb, an unknown resource, and both wildcard forms.

The descriptor endpoint serves the core catalog too. Choosing what an external claim mapping may confer means choosing from everything Elsa declares, not just this module's own resources.

The module now contributes its resource descriptors explicitly rather than relying on the host's assembly scan, for the same reason it calls `AddElsaAuthorization` itself — a consumer wiring the module directly should not get an empty catalog.

## Naming tests repointed

#7982 asked whether the two naming tests should assert the new resource names. They now do. The convention worth holding was always that the module is called `diagnostics/console-logs`, not that a retired constant kept its old value — so the legacy classes and the assertions on them could both go.

## Verification

- `Elsa.ExternalAuthentication.UnitTests` — **179 passed**
- `Elsa.ExternalAuthentication.IntegrationTests` — **138 passed**
- `Elsa.Hosts.SmokeTests` — **4 passed** (both hosts boot, gated routes 401)
- `Elsa.Diagnostics.ConsoleLogs` unit **23** / integration **26**, `StructuredLogs` unit **36**
- All seven touched modules build clean

The host smoke tests matter here specifically: this changes DI registration, and the shell and classic hosts resolve it differently.

## Breaking

Marked `!` — referencing a removed constant is a compile error for anyone outside this repo.

Refs #7982


sfmskywalker:
@greptile review

PR #7989, opened 2026-08-25
test(auth): check the upgrade guide's mapping table against the catalog

sfmskywalker:
Closes the last open item in #7982: the upgrade guide had never been validated against a real deployment.

Walking it by hand once tells you about one afternoon. This checks it on every run instead.

## What it checks

`doc/migrations/authorization-model.md` tells operators how to rewrite every stored permission, and nothing verified that what it tells them to write is a permission Elsa actually accepts. Two assertions over the **published document**, not a copy — so it fails when the guide drifts from the code, which is the direction drift actually goes:

1. Every replacement parses under the `{resource}:{verb}` grammar.
2. Every replacement names a resource and verb some module advertises.

That gap was not hypothetical. The cutover left several checks comparing against legacy constants that the guide itself instructs you to replace, so following it silently disabled them. An entry that does not parse, or that names something nothing advertises, is a deployment locked out of an endpoint by doing exactly as it was told.

## Building the catalog took three attempts

Recording the two that under-reported, because both look correct:

| Approach | Why it fails |
|---|---|
| `AppDomain.CurrentDomain.GetAssemblies()` | Describes whatever earlier tests happened to touch. This class saw **29** resources as missing run alone and **none** in a full run — order-dependent, and worse than no test. |
| Walk `GetReferencedAssemblies()` from both hosts | The compiler drops references to assemblies whose types the app never names. AI, OpenTelemetry and Shells all vanished **despite being project references**. |
| Load every `Elsa.*.dll` in the output directory | Complete, and gives the same answer in isolation as in a full run. |

That last property is the one that matters, and it is verified both ways in this PR.

## Verification

Injected a malformed entry and an unadvertised one into the guide; each check fails naming exactly that entry and nothing else. Restored afterwards — the guide is unchanged in this diff.

Both checks are guarded against passing vacuously: the parser asserts it found a plausible number of rows, and the catalog asserts it is not empty.

**The table as it stands passes both.** So the guide is now verified correct, rather than merely unverified.

`Elsa.Hosts.SmokeTests`: **6 passed** (4 host smoke + 2 here), isolated and full runs agreeing.

Refs #7982

### DIFF
--- a/test/integration/Elsa.Hosts.SmokeTests/MigrationGuideTests.cs
+++ b/test/integration/Elsa.Hosts.SmokeTests/MigrationGuideTests.cs
@@ -1,0 +1,136 @@
+using System.Reflection;
+using System.Text.RegularExpressions;
+using Elsa.Authorization;
+using Elsa.Permissions;
+using Elsa.ModularServer.Web;
+using Elsa.Server.Web;
+
+namespace Elsa.Hosts.SmokeTests;
+
+/// <summary>
+/// Checks the upgrade guide's mapping table against the catalog a host actually builds.
+/// </summary>
+/// <remarks>
+/// The table tells operators how to rewrite every stored permission, and nothing had ever checked that what
+/// it tells them to write is a permission Elsa accepts. That is not hypothetical: the cutover left several
+/// checks comparing against legacy constants which the guide itself instructs you to replace, so following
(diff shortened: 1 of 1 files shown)


sfmskywalker:
@greptile review

PR #7990, opened 2026-08-25
refactor(auth): remove the vestigial per-author script permission plumbing

### LINKED ISSUE #7975: Per-author trust for host-code expressions (C#/Python) needs proper scoping
Follow-up from the authorization model rework (#7974).

## What changed

`exec:csharp-expressions` and `exec:python-expressions` are removed rather than translated into the new permission vocabulary. The host switch — `CSharpOptions.AllowHostCodeExecution` / `PythonOptions.AllowHostCodeExecution`, surfaced through `IExpressionDescriptorRegistry.IsBrowsable` — becomes the single control.

## Why they were removed

The permissions conflated two different concerns, enforced across three sites:

**Execution path** (`Execute`, `Dispatch`, `BulkDispatch` via `WorkflowDefinitionScriptAuthorizationService`) — incoherent. A workflow executes under the server's authority, not the caller's, so the check never constrained what a script could do; it only decided whether a given caller could trigger a definition someone else authored and published. `workflows/definitions:execute` already answers that. It also had a poor failure mode: an author adding a C# expression silently revoked an operator's ability to run a workflow they had been running for months, with no way for the operator to see why.

**Authoring path** (`Post`, `Import`, `ImportFiles`, `Publish`, `BulkPublish`) — meaningful. Saving a C# or Python expression means executing arbitrary code on the host; neither Roslyn scripting nor Python.NET is a sandbox. But this is a constraint on the *write* path, not an `execute` verb on a scripting resource, so the permission name mis-describes what it guards.

**Editor surface** (`Endpoints/Scripting/ExpressionDescriptors/List`) — filters which expression types are offered, using the same permissions.

Rather than carry a mis-scoped permission into the new vocabulary, it is removed and should be reintroduced deliberately.

## Consequence to address

Where host code is **disabled**, nothing changes.

Where host code is **enabled**, any author who may write workflow definitions may use C# and Python, and the editor offers those expression types to every such author. Deployments that enabled host code while permitting only *some* authors to use it lose that granularity.

This matters most for multi-team and multi-tenant deployments, where "this deployment allows host code" and "this author is trusted with host code" are genuinely different statements.

## What a proper design needs to settle

- The gate belongs on the **write** path — creating, importing, and publishing a definition — not on execution.
- Whether it is one permission or per-language, and how it composes with the existing host switch (which should remain the outer boundary).
- Where it sits in the resource tree without a broad `workflows/*` grant silently conferring it — the reason it was not simply relocated. See the discussion in `specs/013-rbac-authorization-model/contracts/permissions.md`.
- Whether JavaScript and Liquid expressions warrant the same treatment. They also execute server-side, though Jint is sandboxed in a way Roslyn and Python.NET are not.
- Enforcement must be declarative rather than inside handlers, per FR-019 of the authorization spec.

sfmskywalker:
#7975 is closed won't-do. This removes what the code was still half-carrying for it.

## The decision

Authoring a workflow is a trusted act, and a per-author gate would not change what a script can do once it runs — a workflow executes under the server's authority, not the caller's. The host switch stays the control, and since it is per language (`CSharpOptions` / `PythonOptions`), an author who should not have host code gets a host with the switch off rather than a permission.

## What was still there

Three places read as an authorization gate to anyone scanning the file, and none of them is one:

- `WorkflowDefinitionScriptAuthorizationService` took a `ClaimsPrincipal` it never read, and could return a `MissingPermission` reason nothing produced.
- Two call sites branched on that reason to send a **403 that cannot happen**.
- The expression-descriptor endpoint kept a map from expression type to per-author permission. Its values went unused *even before the permissions were retired* — `TryGetValue`'s `out var permission` was never referenced. It only ever tested membership; the decision was always `IsBrowsable`.

## What changed

The principal, the unreachable reason and both dead branches are gone. The map becomes a `FrozenSet` of the expression types the host can switch off, which is what it was being used as.

**Behaviour is unchanged.** The only possible failure is a language the host disabled — a property of the deployment, not the caller — so it stays a 400 naming the switch, never a 403.

`PermissionNames` loses `ExecuteCSharpExpressions` and `ExecutePythonExpressions`, which existed only for that map and the test mirroring it. Five other legacy constants there are also unreferenced, but they belong to other modules and are **left alone** rather than swept up here.

## Tests

Two tests asserting the "host and user allow it" case became exact duplicates of the host-only case once the principal stopped mattering, so they go with it. 7 remain, covering both languages across host-on, host-off, expression usage and `RunCSharp`/`RunPython` activity usage.

- `Elsa.Workflows.IntegrationTests` — **295 passed**
- `Elsa.Hosts.SmokeTests` — **6 passed**

## Docs

The migration guide told operators they lose per-author granularity *"until #7975 lands"* and advised disabling host code until then. That promise is now false, so it is withdrawn and replaced with the real guidance: the per-language switch, and a separate host for untrusted authors.

The `specs/013-*` design records are left as the historical account of the decision at the time.

Closes #7975

### DIFF
--- a/doc/migrations/authorization-model.md
+++ b/doc/migrations/authorization-model.md
@@ -39,7 +39,11 @@ Today they are literal claim values, not patterns: `read:*` authorizes only the 
 - Where host code is **disabled**, nothing changes.
 - Where host code is **enabled**, any author who may write workflow definitions may use C# and Python, and the editor offers those expression types to every such author.
 
-Deployments that enabled host code while trusting only *some* authors lose that granularity until [#7975](https://github.com/elsa-workflows/elsa-core/issues/7975) lands. If that matters to you, disable host code until then.
+Deployments that enabled host code while trusting only *some* authors lose that granularity, and this is now the
+intended posture rather than a gap awaiting a fix: [#7975](https://github.com/elsa-workflows/elsa-core/issues/7975)
+was closed as won't-do. Authoring a workflow is a trusted act, and a per-author gate would not have changed what a
+script can do once it runs. If some of your authors are not trusted with host code, give them a host with the
+switch off; the switch is per language, so C# and Python can be decided separately.
 
 ## Revocation
 
@@ -146,8 +150,8 @@ If you have duplicate names across tenants today, they were impossible to create
 | `read:output-converters` | `workflows/descriptors/output-converters:view` |
 | `read:workflow-activation-strategies` | `workflows/descriptors/activation-strategies:view` |
 | `read:javascript-type-definitions` | `workflows/scripting/javascript:view` |
-| `exec:csharp-expressions` | *removed* — see #7975 |
-| `exec:python-expressions` | *removed* — see #7975 |
+| `exec:csharp-expressions` | *removed* — the host switch is the control; see #7975 |
+| `exec:python-expressions` | *removed* — the host switch is the control; see #7975 |
 | `read:user` | `identity/users:view` |
 | `create:user` | `identity/users:create` |
 | `update:user` | `identity/users:update` |
--- a/doc/wiki/expressions-and-scripting.md
+++ b/doc/wiki/expressions-and-scripting.md
@@ -61,13 +61,13 @@ elsa.UseCSharp(options =>
 });
 ```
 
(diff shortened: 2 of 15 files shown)


sfmskywalker:
@greptile review

PR #7991, opened 2026-08-25
feat(secrets)!: scope secrets to tenants

### LINKED ISSUE #7972: Elsa.Secrets is not tenant-aware
## Problem

`Elsa.Secrets` has no notion of tenancy. `Secret` is a plain POCO that does not derive from `Elsa.Common.Entities.Entity`, so it carries no `TenantId`, and there is no tenant filtering anywhere in the module or in its persistence packages (`Elsa.Secrets.Persistence.EFCore*`, `Elsa.Secrets.Persistence.VNext`).

Every other user-facing entity in Elsa — `WorkflowDefinition`, `WorkflowInstance`, `Label`, `AlterationPlan`, `User`, `Role`, `Application` — derives from `Entity` and is scoped by the global query filter installed by `SetTenantIdFilter`. Secrets are the exception.

## Impact

In a multi-tenant deployment, every tenant sees and can resolve every other tenant's secrets. Permission-based access control does not help: `read:secrets` / `write:secrets` are evaluated against the caller's permissions, not against which tenant owns the secret, so any caller holding the permission reaches the whole set.

This also interacts with External Authentication, where a tenant-scoped `IdentityProviderConnection` can bind to a secret that is effectively global.

## Scope

This is a feature, not a hardening fix — the data is not partitioned at all, so there is no filter to tighten:

- Add `TenantId` to `Secret` (derive from `Entity`, or add the property and wire the saving/model-creating handlers).
- EF Core migrations across all five providers (Sqlite, SqlServer, PostgreSql, MySql, Oracle).
- Decide the semantics of a tenant-agnostic secret (`"*"`, per ADR 0009) — deployments will want shared platform secrets alongside tenant-owned ones.
- `ISecretResolver` must resolve within the ambient tenant context; workflow execution already runs inside a tenant scope.
- `ConfigurationSecretStore` is read-only and configuration-backed — decide whether it is tenant-agnostic by definition.
- Backfill/migration story for existing secrets on upgrade.

## Notes

Surfaced while designing the authorization model rework (`specs/013-rbac-authorization-model/`). That work deliberately excludes secrets tenancy so it can be scoped on its own merits.

sfmskywalker:
Closes #7972. `Secret` was the one user-facing entity with no notion of tenancy — every tenant could see and resolve every other tenant's secrets, and permissions did not help because `secrets:view` is evaluated against the caller, not against which tenant owns the secret.

## The change is smaller than the issue suggests

The infrastructure was already in place: `SecretsElsaDbContext` derives from `ElsaDbContextBase`, and `EFCoreSecretsPersistenceFeature` from `PersistenceFeatureBase`, which registers `SetTenantIdFilter`. The handler was skipping secrets for exactly one reason:

```csharp
if (!typeof(Entity).IsAssignableFrom(entityType.ClrType))
    return;
```

`Secret : Entity` plus a column per provider is most of it.

## No backfill, deliberately

The issue lists "backfill/migration story" as open. It largely dissolves. `SetTenantIdFilter` already contains a clause written for exactly this case:

```
TenantId == context.TenantId || TenantId == "*" || (TenantId == null && context.TenantId == "")
```

So the column is added **nullable with no data step**:

| Deployment | Existing secrets |
| --- | --- |
| Single-tenant | Unchanged — the filter is only installed when multitenancy is enabled, so nothing applies |
| Multi-tenant | Not visible until assigned — a visible failure rather than continued exposure |

## Two things that were not obvious

**`Secret.Id` would have silently become null.** It self-initializes to a Guid and *nothing else ever assigns one* — there is no identity generator on the create path. `Entity.Id` is `null!`, so deriving naively means a null id on every insert. Any test that builds a `Secret` by hand would have missed it. A constructor preserves the behaviour.

**The unique index had to move.** From `NormalizedName` to `(TenantId, NormalizedName)`, matching what `User`, `Role` and `Application` did in the same release. Left global, secret names become a shared resource — the second tenant to want `smtp-password` could not create one. Asserted at the schema level, because the repository's own duplicate check queries `dbContext.Secrets`, which the query filter already scopes; testing through it would exercise `SetTenantIdFilter` rather than this change.

## VNext cannot support this

`Elsa.Secrets.Persistence.VNext` keys documents by secret name alone, and `Elsa.Persistence.VNext` has no tenant concept to filter on. It now **throws outside the default tenant** rather than silently serving one tenant's secret to another. Single-tenant deployments are unaffected.

Making it tenant-aware means changing the document id scheme, which relocates existing documents — a storage change to make deliberately, not fold into this.

## Verification

- `Elsa.Secrets.UnitTests` — **65 passed** (63 before, +2). These run `MigrateAsync`, so the new migration is applied against a real SQLite database on every run.
- `Elsa.Persistence.VNext.UnitTests` — **49 passed**. The default-tenant path needs no case of its own: all 49 now run through the new guard with an empty tenant id, which is what a single-tenant deployment does.
- `Elsa.Hosts.SmokeTests` — **6 passed**; `Elsa.ExternalAuthentication.IntegrationTests` — **139 passed**.
- All five providers' migrations verified identical in shape: nullable column, composite index, **zero** `Sql`/`UpdateData` calls.

## Breaking

Marked `!` — multi-tenant deployments must assign existing secrets to a tenant, or `*` to share them. Upgrade guide in `doc/migrations/secrets-tenancy.md`, including the downgrade caveat (recreating the global unique index fails if two tenants hold the same name by then).

## Note

`Elsa.Secrets.Persistence.EFCore.MySql` targets only `net8.0;net9.0`, unlike its siblings which include `net10.0`. Pre-existing; its migration was generated against `net9.0`. Worth a look separately.

Refs #7972


sfmskywalker:
@greptile review

PR #7992, opened 2026-08-26
feat(external-auth)!: require a permission to author policy default roles

### LINKED ISSUE #7977: Should setting unlinked-policy defaultRoleIds require external-authentication:roles:assign?
## Observation

The `external-authentication:roles:assign` descriptor reads:

> Assign external authentication default roles — Assign or remove Elsa Roles used when an external authentication policy creates a user.

That overstates what it guards. Setting `defaultRoleIds` on an unlinked-identity policy is validated by `RoleAuthorizationService.CanAssignRolesAsync` in `IdentityProviderConnectionManagementService.ValidatePolicyAsync` — the ordinary "you may only grant roles whose permissions you already hold" subset rule. `roles:assign` itself is enforced in exactly one place: `ExternalAuthenticationRoleDeletionDependencyContributor.ValidateRemovalAsync`, when removing policy references to a role being deleted.

## Why this is not a vulnerability

The subset rule applies either way, so no caller can grant privileges they do not already hold. An actor without `roles:assign` who can edit a connection can still set `defaultRoleIds`, but only to roles within their own permission set.

## The question

Is that intended? Two readings:

1. **Working as designed.** `roles:assign` governs the role-deletion remediation path specifically, and the subset rule is the right guard for authoring. The descriptor text simply overstates it and should be narrowed.
2. **A missing guard.** `roles:assign` was meant to gate *authoring* default roles as well, and the check was never wired into the write path. Deployments wanting "may configure connections, may not decide what auto-created users receive" cannot express that today.

The second is a real separation-of-duties posture, particularly where connection administration and identity governance sit with different teams.

## Related work

The authorization model rework (#7974) corrects the descriptor text to describe what it actually guards, and maps the permission to `external-authentication/policies/default-roles:update`. That deliberately does **not** settle this question — adding a guard is a behaviour change that would break roles which set default roles today without holding the permission, and it belongs with the module owner rather than inside a vocabulary migration.

sfmskywalker:
Implements the "missing guard" reading of #7977.

> ⚠️ **Ready but deliberately not merged.** This is a breaking authorization change that makes currently-working operations start failing validation. #7977 says the decision "belongs with the module owner", so this waits on your word. Merge it, or close it and I'll narrow the descriptor text instead.

## The gap, demonstrated rather than argued

With the guard stubbed out, a caller holding only `connections:create` and `policies:update` creates a connection whose policy assigns `workflow-user` — response **201 Created**.

The subset rule does not object, because it answers a different question: it prevents *escalation* (you cannot grant roles carrying permissions you lack), not *delegation of the decision*.

## Why this reads as unwired rather than intentional

`external-authentication/policies:update` is already enforced on the write path at both the create and update sites, through the same `RequiresPolicyManagement` condition covering the very policy the roles live inside. Meanwhile `policies/default-roles:update` was enforced in exactly one place: removing policy references while deleting a role.

## What the guard actually keys on

Review found three ways an initial "is the set non-empty" check was wrong. It now triggers on the **effective set changing**:

| Scenario | Requires the permission |
| --- | --- |
| Adding, editing or clearing default roles | Yes |
| Switching the policy away from create-user, which drops its roles | Yes |
| Editing an unrelated field, enabling, or validating, roles untouched | **No** |
| Creating a policy that assigns none | **No** |

The baseline comes from the **connection registry**, not the database store — a configuration-owned connection has no database row, so a store-only baseline made its configured roles read as newly assigned on every validation, locking out callers with only `connections:view`.

## Who breaks

Anyone who held legacy `external-authentication:roles:assign` maps to the new permission and is unaffected. The break is for roles holding `policies:manage` (→ `policies:view` + `policies:update`) but *not* `roles:assign`, that change default roles.

If it is going to be added, now costs least: operators are already rewriting every stored permission, so this is one more grant inside an edit they are making anyway.

## Verification

- `Elsa.ExternalAuthentication.IntegrationTests` — **145 passed** (139 before, +6)
- `Elsa.Hosts.SmokeTests` — **6 passed**
- All 17 CI checks green

Every one of the four behaviours above has a test, and each was checked by reverting the fix and confirming the expected failure:

| Revert | Symptom |
| --- | --- |
| Remove the guard | `Expected: BadRequest, Actual: Created` |
| Key on presence not change | 400 on an unrelated field edit |
| Scope to create-user candidates | `Expected: BadRequest, Actual: OK` on abandonment |
| Baseline from store only | Permission error validating a configuration-owned connection |

The fixture gained a per-test permission override, since it was previously all-or-nothing and could not express "manages policies but not default roles" — the very distinction at issue.

Closes #7977

### DIFF
--- a/doc/migrations/authorization-model.md
+++ b/doc/migrations/authorization-model.md
@@ -115,8 +115,9 @@ permission and is unaffected. The break is for roles holding `external-authentic
 (→ `policies:view` + `policies:update`) but *not* `roles:assign`, which set default roles today. Grant them
 `external-authentication/policies/default-roles:update`, or move that responsibility to a role that has it.
 
-Clearing the list, or a policy that sets no default roles, needs nothing extra — the permission is required
-only when roles are actually being assigned.
+The permission is required when the default-role set **changes** — adding, removing, or clearing. Leaving a
+stored set alone needs nothing extra, so an administrator without the permission can still edit other fields
+on a connection whose default roles someone else configured, enable it, or validate it.
 
 ## Third-party modules
 
--- a/doc/migrations/authorization-model.md
+++ b/doc/migrations/authorization-model.md
@@ -115,9 +115,10 @@ permission and is unaffected. The break is for roles holding `external-authentic
 (→ `policies:view` + `policies:update`) but *not* `roles:assign`, which set default roles today. Grant them
 `external-authentication/policies/default-roles:update`, or move that responsibility to a role that has it.
 
-The permission is required when the default-role set **changes** — adding, removing, or clearing. Leaving a
-stored set alone needs nothing extra, so an administrator without the permission can still edit other fields
-on a connection whose default roles someone else configured, enable it, or validate it.
+The permission is required when the default-role set **changes** — adding, removing, clearing, or switching
+the policy away from one that creates users, which drops its roles just as surely. Leaving a stored set alone
+needs nothing extra, so an administrator without the permission can still edit other fields on a connection
+whose default roles someone else configured, enable it, or validate it.
 
 ## Third-party modules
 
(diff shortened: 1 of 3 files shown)


sfmskywalker:
@greptile review

sfmskywalker:
@greptile review

sfmskywalker:
@greptile review

PR #7994, opened 2026-08-26
fix(core): clear stale blocked tokens when a flowchart activity completes (#7993)

### LINKED ISSUE #7993: Token flow: WaitAny (Race) join inside a loop is not re-triggered from a different inbound connection (stale blocked tokens)
## Description

In token-based flowchart execution, a `FlowJoin` with `Mode = WaitAny` (mapped to `MergeMode.Race`) that is re-entered through a loop fails to fire when it is later triggered from a **different inbound connection** than the previous activation. The fresh trigger is silently swallowed and the flowchart completes early without running the remaining activities. No exception is thrown and the workflow status is still `Finished`, so the failure is silent.

## Steps to Reproduce

1. Enable token-based flowchart execution (not the default; see "Additional Configuration" below).
2. Create a flowchart where a `FlowJoin` (WaitAny) has **two forward inbound connections** (`A → join`, `B → join`) and sits inside a **loop**, so that the join is activated more than once, from different inbounds.
3. Run the workflow so that:
   - Pass 1: the join fires via inbound `A` (this blocks the `B → join` connection);
   - Loop: the join fires again via a loop back-edge (this blocks both forward inbounds again);
   - Pass 2: the join should fire via inbound `B`.

Minimal repro workflow (code-based, no designer JSON needed):

```csharp
using Elsa.Workflows.Activities;
using Elsa.Workflows.Activities.Flowchart.Activities;
using Elsa.Workflows.Activities.Flowchart.Models;

public class RaceJoinLoopWorkflow : WorkflowBase
{
    {
        var counter = builder.WithVariable(0);
        var start = new WriteLine("Start");
        var whichPath = new FlowDecision(context => counter.Get(context) == 0);
        var isSecondPass = new FlowDecision(context => counter.Get(context) == 2);
        var pathA = new WriteLine("A");
        var pathB = new WriteLine("B");
        var join = new FlowJoin(); // WaitAny by default → MergeMode.Race
        var body = new WriteLine("Body");
        var increment = new SetVariable<int>(counter, context => counter.Get(context) + 1);
        var loopDecision = new FlowDecision(context => counter.Get(context) <= 1);
        var retry = new WriteLine("Retry");
        var end = new WriteLine("End");

        builder.Root = new Flowchart
        {
            Start = start,
            Activities =
            {
                start, whichPath, isSecondPass, pathA, pathB, join, body, increment, loopDecision, retry, end
            },
            Connections =
            {
                new(start, whichPath),
                new(new(whichPath, "True"), new Endpoint(pathA)),
                new(new(whichPath, "False"), new Endpoint(isSecondPass)),
                new(new(isSecondPass, "True"), new Endpoint(pathB)),
                new(new(isSecondPass, "False"), new Endpoint(end)),
                // Two forward inbounds into the same WaitAny join.
                new(pathA, join),
                new(pathB, join),
                new(join, body),
                new(body, increment),
                new(increment, loopDecision),
                // Loop back-edge into the join.
                new(new(loopDecision, "True"), new Endpoint(retry)),
                new(retry, join),
                // Exit the loop and re-enter the join through the other forward inbound.
                new(new(loopDecision, "False"), new Endpoint(whichPath)),
            }
        };
    }
}
```

Failing test (xunit, follows the existing `JoinBehaviors` test style):

```csharp
[Fact]
public async Task WaitAny_join_reentered_via_loop_can_be_triggered_from_another_inbound()
{
    var options = new RunWorkflowOptions().WithTokenBasedFlowchart();
    var result = await _fixture.RunWorkflowAsync<RaceJoinLoopWorkflow>(options);
    var lines = _fixture.CapturingTextWriter.Lines.ToList();
    Assert.Equal(new[] { "Start", "A", "Body", "Retry", "Body", "B", "Body", "End" }, lines);
    Assert.Equal(WorkflowStatus.Finished, result.WorkflowState.Status);
}
```

**Reproduction rate:** every time.

**Attachments:** the two files above are self-contained; I can also attach them as a ready-to-drop-in test (`RaceJoinLoopTests.cs` + `Workflows/RaceJoinLoopWorkflow.cs` for `test/integration/Elsa.Workflows.IntegrationTests/Scenarios/JoinBehaviors`).

## Expected Behavior

The join fires on every activation, producing:

```
Start, A, Body, Retry, Body, B, Body, End
```

## Actual Behavior

The join's third activation (via `B`) is swallowed; the flowchart completes early:

```
Start, A, Body, Retry, Body, B
```

(`Body` never runs a third time, `End` is never reached, yet the workflow reports `Finished`.)

The persisted workflow state of a stalled instance shows the smoking gun — the token for the `B → join` connection is `blocked: true, consumed: true`, i.e. it was consumed by the "consume the block without scheduling" branch instead of triggering the join.

## Screenshots

N/A — the attached xunit test demonstrates the issue deterministically; its assertion output shows the truncated line sequence above.

## Environment

- **Elsa Package Version**: built from source, tag `3.7.1`. `Flowchart.Tokens.cs` is identical in `3.8.0-rc2`, so the issue persists there.
- **Operating System**: Windows 10 (10.0.19044), .NET SDK 10.0.400 — the repro is platform-independent.
- **Browser and Version**: N/A (engine-level issue)

## Log Output

No error is logged. Test assertion failure (unpatched):

```
Assert.Equal() Failure: Collections differ
Expected: string[]     [···, "Retry", "Body", "B", "Body", "End"]
Actual:   List<string> [···, "Retry", "Body", "B"]
```

## Troubleshooting Attempts

Root cause: every time the join is scheduled, the `MergeMode.Race` branch in `OnChildCompletedTokenBasedLogicAsync` (`Flowchart.Tokens.cs`) adds **blocked tokens for all other forward inbound connections** of the join. These blocked tokens are never removed once the join has executed. When a token later arrives on such a connection (e.g. the next loop iteration), the `existingBlockedToken` lookup finds the stale block and the code path `// Consume the block without scheduling.` swallows the fresh trigger, so the join is never scheduled again.

Proposed fix — the activation wave for an activity is decided once it completes, so blocked inbound tokens should not outlive it. In `OnChildCompletedTokenBasedLogicAsync`, right after consuming the inbound tokens of the completed activity:

```csharp
// Consume inbound tokens to the completed activity.
var inboundTokens = tokens.Where(t => t.ToActivityId == completedActivity.Id && t is { Consumed: false, Blocked: false }).ToList();
foreach (var t in inboundTokens)
    t.Consume();

// The activation wave for the completed activity is decided once it completes.
// Clear blocked inbound tokens so future activations (e.g. via loop-back edges) are not swallowed by stale blocks.
tokens.RemoveWhere(t => t.ToActivityId == completedActivity.Id && t.Blocked);
```

This keeps the race guard working *within* a wave (a losing branch that completes while the join is already scheduled still consumes its block without double-scheduling), while allowing future activations in loops.

Verification with the fix applied: the repro test passes, and the existing flowchart integration suites still pass (JoinBehaviors, ExplicitJoins, FlowchartCompletion, FlowchartNextActivity — 35 tests, 0 failures).

## Additional Context

**Additional configuration required to reproduce:** token-based flowchart execution must be enabled, since counter-based remains the default (`Flowchart.UseTokenFlow = false` / `FlowchartOptions.DefaultExecutionMode = CounterBased`). In tests this is done per run via `new RunWorkflowOptions().WithTokenBasedFlowchart()`; in an app via `.UseFlowchart(f => f.UseTokenBasedExecution())`.

The bug only manifests when all of the following hold: the join uses WaitAny/Race semantics, it has more than one forward inbound, and it is activated multiple times from different inbounds (typical for loops that re-enter a join).

## Related Issues

- #6229 — the counter-based engine has an analogous join re-trigger defect; this issue is its token-flow counterpart.
```

Alfxjx:
## Purpose

Fixes a silent failure in token-based flowchart execution where a `FlowJoin` (WaitAny / `MergeMode.Race`) re-entered through a loop could not fire again from a different inbound connection — the fresh trigger was swallowed by a stale blocked token.

Fixes #7993

---

## Scope

Select one primary concern:

- [ ] Documentation update
- [ ] New feature

---

### Problem

In token-based flowchart execution, every time a `MergeMode.Race` target is scheduled, blocked tokens are added for its other forward inbound connections. These blocked tokens were never removed once the join had executed. When a token later arrived on such a connection — e.g. a join re-entered through a loop back-edge and then triggered from a different forward inbound — the stale block was found by the `existingBlockedToken` lookup and the trigger was consumed by the "consume the block without scheduling" branch. The join never fired again and the flowchart completed early, silently, with status `Finished`.

Full root cause, minimal repro workflow, and a failing test are in #7993.

### Solution

The activation wave for an activity is decided once it completes, so blocked inbound tokens should not outlive it. In `OnChildCompletedTokenBasedLogicAsync` (`Flowchart.Tokens.cs`), right after consuming the completed activity's inbound tokens, blocked inbound tokens for that activity are removed:

```csharp
tokens.RemoveWhere(t => t.ToActivityId == completedActivity.Id && t.Blocked);

### DIFF
--- a/src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.Tokens.cs
+++ b/src/modules/Elsa.Workflows.Core/Activities/Flowchart/Activities/Flowchart.Tokens.cs
@@ -38,6 +38,10 @@ public partial class Flowchart
         foreach (var t in inboundTokens)
             t.Consume();
 
+        // The activation wave for the completed activity is decided once it completes.
+        // Clear blocked inbound tokens so future activations (e.g. via loop-back edges) are not swallowed by stale blocks.
+        tokens.RemoveWhere(t => t.ToActivityId == completedActivity.Id && t.Blocked);
+
         // Schedule next activities based on merge modes.
         foreach (var connection in activeOutboundConnections)
         {
--- a/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/JoinBehaviors/RaceJoinLoopTests.cs
+++ b/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/JoinBehaviors/RaceJoinLoopTests.cs
@@ -1,0 +1,22 @@
+using Elsa.Testing.Shared;
+using Elsa.Workflows.Activities.Flowchart.Extensions;
+using Elsa.Workflows.IntegrationTests.Scenarios.JoinBehaviors.Workflows;
+using Elsa.Workflows.Options;
+using Xunit.Abstractions;
+
+namespace Elsa.Workflows.IntegrationTests.Scenarios.JoinBehaviors;
+
+public class RaceJoinLoopTests(ITestOutputHelper testOutputHelper)
+{
+    private readonly WorkflowTestFixture _fixture = new(testOutputHelper);
+
+    [Fact(DisplayName = "WaitAny join re-entered via loop can be triggered from another inbound")]
+    public async Task WaitAny_join_reentered_via_loop_can_be_triggered_from_another_inbound()
+    {
+        var options = new RunWorkflowOptions().WithTokenBasedFlowchart();
+        var result = await _fixture.RunWorkflowAsync<RaceJoinLoopWorkflow>(options);
+        var lines = _fixture.CapturingTextWriter.Lines.ToList();
+        Assert.Equal(new[] { "Start", "A", "Body", "Retry", "Body", "B", "Body", "End" }, lines);
+        Assert.Equal(WorkflowStatus.Finished, result.WorkflowState.Status);
+    }
+}
--- a/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/JoinBehaviors/Workflows/RaceJoinLoopWorkflow.cs
+++ b/test/integration/Elsa.Workflows.IntegrationTests/Scenarios/JoinBehaviors/Workflows/RaceJoinLoopWorkflow.cs
@@ -1,0 +1,57 @@
+using Elsa.Workflows.Activities;
+using Elsa.Workflows.Activities.Flowchart.Activities;
+using Elsa.Workflows.Activities.Flowchart.Models;
+
+namespace Elsa.Workflows.IntegrationTests.Scenarios.JoinBehaviors.Workflows;
+
+/// <summary>
+/// Reproduces https://github.com/elsa-workflows/elsa-core/issues/7993:
+/// a WaitAny (Race) join with two forward inbounds that is re-entered through a loop
+/// must fire again when triggered from a different inbound than the previous activation.
+/// </summary>
+public class RaceJoinLoopWorkflow : WorkflowBase
+{
+    protected override void Build(IWorkflowBuilder builder)
+    {
+        var counter = builder.WithVariable(0);
+        var start = new WriteLine("Start");
+        var whichPath = new FlowDecision(context => counter.Get(context) == 0);
+        var isSecondPass = new FlowDecision(context => counter.Get(context) == 2);
+        var pathA = new WriteLine("A");
+        var pathB = new WriteLine("B");
+        var join = new FlowJoin(); // WaitAny by default → MergeMode.Race
+        var body = new WriteLine("Body");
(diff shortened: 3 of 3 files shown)


Alfxjx:
@dotnet-policy-service agree

sfmskywalker:
@greptile

PR #7996, opened 2026-08-26
fix(external-auth): keep the default-roles guard when an update omits the policy

sfmskywalker:
## Summary

Closes a gap in the `PolicyDefaultRoles` guard introduced in #7992 (refs #7977): a PUT that **omitted** `unlinkedPolicy` hit the null-policy early return in `IdentityProviderConnectionManagementService.ValidatePolicyAsync` before the guard ran, so an actor holding only `external-authentication/connections:update` could clear a stored create-user policy — silently dropping its default-role assignments — and get 200 OK. This contradicted the documented model ("adding, removing, clearing, or switching the policy away from one that creates users" all require the permission).

## Changes

- The candidate's effective default-role set is computed **before** the null-policy early return (empty when the policy is omitted or does not create users), so clearing, adding, or switching a policy all count as changing default roles. The remaining policy validations stay in the non-null branch.
- The cheap permission check now runs before the registry-backed role comparison, so the common permitted path skips building the registry.

## Tests

Three new integration tests alongside the existing guard coverage:
- omitting a stored create-user policy → 400 with the permission error (the reported hole),
- introducing a create-user policy on a policy-less connection → 400,
- clearing a policy that assigns no roles → still permitted.

## Verification

`Elsa.ExternalAuthentication.IntegrationTests`: 148/148 green (net10.0); unit suite 185/185. Also verified standalone on this branch in an isolated worktree.


sfmskywalker:
@greptile review

PR #7997, opened 2026-08-26
fix(auth): validate wildcard permission patterns and warn on deny-list stripping

sfmskywalker:
## Summary

Hardens wildcard permission-pattern handling on two fronts:

1. **Inert patterns are rejected at authoring time.** `Permission.IsValidPattern` refuses wildcard spellings that parse but can never match anything (e.g. `workflows*:delete`). The grant boundary validator, `StoredPermissionValidator`, `PermissionGrantValidator`, and `ExternalAuthenticationOptionsValidator` all apply it, so a meaningless pattern is a validation error rather than a grant that silently authorizes nothing.
2. **Deny-list wildcard stripping is surfaced.** A non-empty `DeniedPermissions` list refuses every wildcard grant that could reach a denied permission — including the seeded administrator role's `*`. `ExternalAuthenticationOptionsValidator` now logs a warning (never fails) when `DeniedPermissions` is configured, and the migration guide documents the operational consequence: externally-authenticating administrators need enumerated grants instead of `*`.

## Note for reviewers

One production-DI line rides along: `AddExternalAuthenticationServices` now calls `services.AddLogging()`. The options validator gained an `ILogger<>` ctor dependency and is registered with `ValidateOnStart`, which makes a resolvable logger part of the module's registration contract — same pattern as the module's existing `AddDataProtection`/`AddRateLimiter` self-registrations. `AddLogging` is TryAdd-based, so host logging configuration always wins.

## Verification

`Elsa.Api.Common.UnitTests`, `Elsa.ExternalAuthentication.UnitTests` (185/185), and `Elsa.Identity.UnitTests` (110/110, incl. new `StoredPermissionValidatorTests`) green on net10.0; also verified standalone on this branch in an isolated worktree.

### DIFF
--- a/doc/migrations/authorization-model.md
+++ b/doc/migrations/authorization-model.md
@@ -61,7 +61,10 @@ by exact string, so they read the way a role does.
   connection granting `workflows/*:delete` is denied by a deny list naming only `workflows/definitions:delete`.
   Before this release both comparisons were exact, so either spelling slipped past the other and a deployment's
   deny list did not hold. If you carried a deny list across the upgrade, re-read it: it may now deny more than it
-  used to, which is the intent.
+  used to, which is the intent. A consequence to plan for: any non-empty deny list refuses every wildcard grant
+  that could reach a denied permission, and `*` (which parses to `*:*`) reaches all of them — so a role holding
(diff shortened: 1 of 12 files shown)


sfmskywalker:
@greptile review

PR #7998, opened 2026-08-26
docs(secrets): document null-tenant index gaps and the MySql TFM pin

sfmskywalker:
## Summary

Documentation-only: records two facts about the secrets/tenancy release that operators and future maintainers would otherwise rediscover the hard way. No behavior changes.

1. **Null-tenant rows sit outside the per-tenant unique indexes.** The composite `(TenantId, NormalizedName)` indexes only backstop rows whose `TenantId` is non-null (SQL Server filters them out of the index; SQLite/PostgreSQL/MySQL treat nulls as distinct). Single-tenant deployments and pre-upgrade rows therefore rely on the pre-save existence checks for name uniqueness — sequential duplicates are blocked, concurrent ones are not schema-enforced. Documented in `secrets-tenancy.md`, cross-referenced from the authorization-model guide (same caveat applies to user/role/application names), and noted at the EFCore secret repository's write path.
2. **The MySql provider's TFM pin stays.** `Elsa.Secrets.Persistence.EFCore.MySql` cannot target net10.0: Pomelo.EntityFrameworkCore.MySql tops out at EF Core 9, and the project references `Elsa.Persistence.EFCore.MySql`, which carries the same net8.0/net9.0 pin. The csproj comment now explains this instead of leaving the pin looking like an oversight.

## Verification

`Elsa.Secrets.UnitTests` 65/65 green; model snapshot verified byte-identical to the `SecretTenancy` migration's designer target (no pending model changes). Branch builds standalone in an isolated worktree.

### DIFF
--- a/doc/migrations/authorization-model.md
+++ b/doc/migrations/authorization-model.md
@@ -130,6 +130,8 @@ Included in the same release: `User.Name`, `Role.Name`, `Application.Name` and `
 
 If you have duplicate names across tenants today, they were impossible to create, so no data conflict can arise. Going the other way — downgrading — will fail if duplicates exist by then.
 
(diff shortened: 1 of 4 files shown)


sfmskywalker:
@greptile review

Issue #7976, opened 2026-08-23
ContinueLogout requires authentication, which can 401 the upstream logout callback


sfmskywalker:
## Problem

`ContinueLogout` in `src/modules/Elsa.ExternalAuthentication/Endpoints/Broker/Logout.cs` declares neither `ConfigurePermissions` nor `AllowAnonymous`, so it inherits the FastEndpoints default and requires an authenticated principal.

It should be anonymous. The endpoint takes a `handle` from the route and calls `broker.ContinueLogoutAsync(handle)` — it never touches `HttpContext.User`. The handle *is* the authority, exactly as with every other broker callback (`CompleteCallback`, `CompleteLogout`, `InitiateExternal`, `ExchangeToken`), all of which declare `AllowAnonymous`.

```csharp
internal sealed class ContinueLogout(IExternalAuthenticationBroker broker) : ElsaEndpointWithoutRequest
{
    public override void Configure() => Get("/external-authentication/logout/continue/{handle}");

    public override async Task HandleAsync(CancellationToken cancellationToken)
    {
        var result = await broker.ContinueLogoutAsync(Route<string>("handle")!, cancellationToken);
        ...
    }
}
```

## Impact

The identity provider redirects the browser to this endpoint during upstream logout. At that point the Elsa session may already be gone — that is the whole point of a logout flow — so requiring authentication can reject a callback that should succeed, leaving the logout half-completed: signed out upstream, with local state not cleaned up.

Whether it reproduces depends on how the host is configured and how much of the Elsa session survives to the callback, which is worth confirming before fixing.

## Note on the sibling endpoint

`Logout` in the same file also declares nothing, but for it the inherited default is *correct*: it reads the external session id from `HttpContext.User`, so it genuinely needs an authenticated principal — just no permission. Only `ContinueLogout` looks mis-gated.

## Suggested fix

Declare `AllowAnonymous()` on `ContinueLogout`, and declare `Logout` explicitly as authenticated-only so neither relies on an inherited default.

## Context

Surfaced by the endpoint coverage audit for the authorization model rework (#7974), which requires every endpoint to declare its access explicitly. Not introduced by that work.

sfmskywalker:
Confirmed, and it is worse than the report described. Fix in #7979.

**The authorization defect is certain, not probable.** Two independent reasons the endpoint can never authenticate:

[...]

Issue #7974, opened 2026-08-23
Structured authorization model: hierarchical resources, open verbs, single evaluator


sfmskywalker:
Tracking issue for the authorization model rework.

Design lives in `specs/013-rbac-authorization-model/` — [spec](specs/013-rbac-authorization-model/spec.md), [plan](specs/013-rbac-authorization-model/plan.md), [decision record](specs/013-rbac-authorization-model/research.md), [vocabulary contract](specs/013-rbac-authorization-model/contracts/permissions.md), [API contract](specs/013-rbac-authorization-model/contracts/rest-api.md), [tasks](specs/013-rbac-authorization-model/tasks.md).

**71 tasks across 5 phases**, tagged by user story, with dependencies and parallel guidance in `tasks.md`.

*This description is kept current; the comments are history, not required reading.*

## Why

Elsa has most of an authorization system — roles carry permissions, tokens carry permission claims, and 151 of 160 endpoint files declare one — but no model behind it. A census of all 150 guarded endpoints found:

- **`read:*` is a literal claim value, not a pattern.** It authorizes 12 endpoints out of roughly 40 read endpoints. Granting it produces inconsistent, unpredictable coverage.
- **No implication.** `read:workflow-definitions` does not imply `read:workflow-definitions:versions`. Every relationship is hand-enumerated at the call site.
- **No catalog.** 57 permission strings appear as inline literals across 174 call sites in three competing naming schemes. A typo silently produces an unreachable endpoint, and no UI can render a role editor.
- **Fails open.** Omitting the declaration inherits the FastEndpoints default; there is no Elsa-level fallback and no startup guard.
- **Four parallel enforcement mechanisms** — FastEndpoints permissions, ASP.NET policies, mid-handler `AuthorizeAsync`, and ~15 hand-rolled claim greps (one using a different string comparison than the rest) — with no single place to audit.

## The model

A permission is `{resource}:{verb}`. **Both axes are open** and string-keyed, contributed by modules through descriptors:

```
resource  := hierarchical, '/'-separated
verb      := flat string
satisfies := resourceMatches(granted, required) && verbMatches(granted, required)
```

**Wildcards** are the only construct with forward reach, on either axis. A trailing `*` matches the named node **and all descendants** — `workflows/definitions/*` covers `workflows/definitions` itself plus `.../versions` and `.../labels`. `*:*` is superuser, with no special-cased sentinel.

[...]

PR #8000, opened 2026-08-27
docs(013): reconcile RBAC and User Tasks task lists with the shipped code

sfmskywalker:
## Summary

The task lists for feature 013 had drifted badly out of step with the code. `specs/013-rbac-authorization-model/tasks.md` read **65 open / 6 done** while the authorization-model migration was in fact complete — 17 module catalogs, 155 endpoint files on `RequirePermission`, zero remaining legacy `ConfigurePermissions` callers. Only the Phase 6 documentation tasks had ever been ticked. Anyone opening that file would either redo landed work or stop trusting it.

Every open task was re-verified against the working tree at `origin/main`:

| List | Before | After |
|---|---|---|
| `013-rbac-authorization-model` | 65 open / 6 done | **12 open / 59 done** |
| `013-user-tasks` | 13 open / 45 done | **10 open / 48 done** |

## What remains genuinely open

The 22 still-open tasks each carry an inline note naming the missing evidence, so a real gap is distinguishable from unticked bookkeeping. The substantive ones:

- **T027** — `ConfigurePermissions` is preserved and functional but is *not* marked `[Obsolete]`, and does not resolve legacy strings through a migration table, register an unverified descriptor, or warn. It simply prepends `PermissionNames.All`.
- **T040** — the three `Policies(IdentityPolicyNames.SecurityRoot)` usages remain in `Secrets/Hash`, `Roles/Create` and `Applications/Create`; the constant is `[Obsolete]` but still consumed.
- **T038** — two hand-rolled claim inspections remain (`AIHttpContextIdentity`, still case-insensitive, and `RoleDeletionCoordinator`).
- **T049** — `PermissionGrantOptions` still lives in `Elsa.ExternalAuthentication` rather than being promoted to `Elsa.Identity`.
- **T054** — `MemoryStore` is still not tenant-aware.
- Missing test coverage: T022b, T025, T044, T050, T051, T056.
- Three User Tasks items (T047, T048, T050) are Studio-side and live in `elsa-studio`, so they cannot be closed from this repo.

## Notes

Two tasks landed in a different shape than specified and say so inline rather than being ticked silently: the coverage gate is a shared `EndpointCoverage.AssertEveryEndpointDeclaresAccess(assembly)` helper invoked per module rather than one test in `Elsa.Api.Common.UnitTests` (T041), and the permission stamp is computed by `PermissionStampCalculator` rather than persisted on `User`, which is why it needed no migrations (T047a).

Issue hygiene was checked at the same time: every issue referenced by these specs (#7972, #7974, #7975, #7976, #7977, #7982) is already closed, and no currently-open issue is resolved by this work. Nothing to close.

Docs only — no code changes. Verified by inspection rather than a full build, which is why T063 stays open.


sfmskywalker:
Confirmed, and thank you — that localhost bootstrap finding is real. It is out of scope for this PR, which is a documentation reconciliation: the checklist now records it as an open defect under T042 rather than fixing it.

The code fix is in #8002 (restores the structured spellings and evaluates them through `PermissionEvaluator`). Note that #8003 retires the `SecurityRoot` policy, which was the only attachment point for that grant — so depending on which of the two lands, the bootstrap path is either repaired or deliberately retired. Both are noted in #8003.

Resolving here as documented-not-fixed.

sfmskywalker:
@greptile review

PR #8002, opened 2026-08-27
fix(identity): restore the localhost bootstrap grant under the structured vocabulary

sfmskywalker:
## Summary

The documented **localhost first-run bootstrap path grants nothing**. `LocalHostPermissionRequirementHandler` injects three permissions for local, unauthenticated requests so an operator can stand up an instance:

```csharp
private static readonly string[] BootstrapPermissions =
[
    "create:application",
    "create:user",
    "create:role"
];
```

Those are legacy-vocabulary strings. The three endpoints they exist to unlock have since moved to structured permissions — `identity/applications:create`, `identity/users:create`, `identity/roles:create` — and a legacy string does not merely fail to match, it parses to a **different `(resource, verb)` pair**: `"create:user"` reads as resource `create`, verb `user`. So the injected claims authorize none of the three endpoints, and the local bootstrap cannot create the first application, user or role.

This is the `LocalHostPermissionRequirement` half of **T042**, which the task explicitly names alongside the token issuers.

## Changes

- `BootstrapPermissions` is now `Permission[]` in the structured vocabulary. The literals are spelled out rather than referenced from `Elsa.Identity`, because `Elsa.Api.Common` sits beneath it.
- The authenticated branch evaluates through `PermissionEvaluator.HasAllPermissions` instead of comparing claim values by set containment, so a caller holding a covering wildcard such as `identity/*:create` now satisfies the requirement. `*` continues to satisfy it, as before.

## Tests

The existing tests asserted the legacy spellings, so they would have kept the defect in place — they now assert the structured ones. Added:

- a theory asserting the injected grant **actually authorizes** each of the three endpoints when run through the evaluator, rather than merely appearing as a claim string (the assertion that would have caught this);
- a case for an authenticated caller holding `identity/*:create`.

`Elsa.Identity.UnitTests` **115/115** green on net10.0.

## Related

Found while working through the task-list reconciliation in #8000, which now records T042 as partially done. Same root cause as #8001 (`RoleDeletionCoordinator` gating on the legacy `"delete:role"`): code holding legacy permission strings that silently stopped matching when the endpoints around it migrated.

### DIFF
--- a/src/common/Elsa.Api.Common/Requirements/LocalHostPermissionRequirement.cs
+++ b/src/common/Elsa.Api.Common/Requirements/LocalHostPermissionRequirement.cs
@@ -1,4 +1,5 @@
 using System.Security.Claims;
+using Elsa.Authorization;
 using Elsa.Extensions;
 using Elsa.Options;
 using JetBrains.Annotations;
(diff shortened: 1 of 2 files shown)


sfmskywalker:
Superseded by #8003. That PR retires the `SecurityRoot` policy and, with it, deletes `LocalHostPermissionRequirement` entirely — so the legacy strings this PR repairs no longer exist to repair.

Keeping this open only as the fallback if we decide to preserve the localhost grant instead. Otherwise it should be closed as moot.

PR #8166, opened 2026-09-14
fix(workflows): persist named WithVariable values across suspend/resume

### LINKED ISSUE #8159: [BUG] WithVariable<T>(string name, T value) configures no storage driver, so variables silently lose their value across suspend/resume
## Describe the bug

`IWorkflowBuilder` has two overloads that behave very differently with respect to persistence:

```csharp
// Elsa.Workflows.Core/Builders/WorkflowBuilder.cs
public Variable<T> WithVariable<T>()
{
    var variable = new Variable<T>(null!, default(T));
    Variables.Add(variable);
    variable.WithWorkflowStorage();   // <-- WorkflowInstanceStorageDriver
    variable.Id = null!;
    return variable;
}

public Variable<T> WithVariable<T>(string name, T value)
{
    var variable = new Variable<T>(name, value);
    Variables.Add(variable);
    return variable;                  // <-- no storage driver at all
}
```

The named overload never sets `StorageDriverType`. Per the XML docs on `Variable.StorageDriverType`, such a variable "will remain in memory for as long as the expression execution context exists" — so its value is **not** persisted with the workflow instance and is gone the moment the workflow suspends and resumes.

This is easy to miss because 3.8.0 marked the parameterless overload obsolete:

```csharp
[Obsolete("Use the overload that takes a name instead. This overload will be removed in a future version.")]
Variable<T> WithVariable<T>();
```

Following that guidance — a one-line, mechanical change — silently turns every workflow variable from persisted into in-memory-only. There is no warning, no exception at build or start time, and the workflow runs correctly right up until the first suspension point.

The failure that eventually surfaces is far away from the cause. In our case a variable read fine before a `Delay`, and after the `Delay` the same expression threw:

```
Elsa.Workflows.Exceptions.InputEvaluationException: Failed to evaluate activity input 'MemberId'
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
```

because `MemoryBlockReference.Get(context)` returned `null` for a variable that had been populated earlier. Variables of value types are worse: a `Variable<Guid>` just quietly comes back as `Guid.Empty` and the workflow takes a wrong branch with no error at all.

## To Reproduce

```csharp
public class ReproWorkflow : WorkflowBase
{
    {
        Variable<string> message = builder.WithVariable<string>("message", null!);

        builder.Root = new Sequence
        {
            Activities =
            {
                new SetVariable { Variable = message, Value = new("hello") },
                new WriteLine(context => $"before suspend: {message.Get(context) ?? "<null>"}"),
                new Delay(TimeSpan.FromSeconds(10)),
                new WriteLine(context => $"after resume:   {message.Get(context) ?? "<null>"}"),
            },
        };
    }
}
```

Run with any persistent store (we use `Elsa.Persistence.Dapper` on SQL Server) so the workflow really suspends and is rehydrated.

Swapping line 1 for the obsolete `builder.WithVariable<string>()` makes the problem disappear.

## Expected behavior

`WithVariable<T>(string name, T value)` persists the variable with the workflow instance, exactly like `WithVariable<T>()` does — or, if the difference is deliberate, it is called out in the overload's XML documentation and in the obsolete message that steers people toward it.

## Actual behavior

The variable is in-memory only. After suspend/resume the value is `null`/`default`, with no error, warning or log entry.

## How often does this bug happen?

Every time.

## Additional context

- This is **not** a 3.8 regression. I decompiled `Elsa.Workflows.Core` 3.7.1, 3.8.0 and 3.8.1 and the two method bodies are identical in all three; `main` still matches. The named overload has never set a storage driver. What changed in 3.8.0 is the `[Obsolete]` attribute on the parameterless overload, which is what makes people migrate onto it.
- The overload was added for ergonomics around `Output<T>` bindings (#7694, #7701); persistence does not appear to have been considered in either the issue or the PR, and the 3.8.0 release notes list it only as a developer-facing convenience with no behavioural note.
- #7701 also adds `WithVariable<T>(string name)`, which likewise sets no storage driver, so it will inherit the same behaviour when it ships.
- Relatedly, #6750 asks for a workflow-wide default storage driver, which would make this whole class of mistake impossible.

### Possible fixes

1. **Make the named overload match** — add `variable.WithWorkflowStorage();`. Behaviour-compatible with what people migrating from the obsolete overload expect, and matches what almost every caller wants. This does change behaviour for anyone already relying on the current in-memory semantics, so it may want to be a 3.9 change.
2. **At minimum**, document it: mention the missing driver on the overload's XML docs, and reword the obsolete message to something like *"Use the overload that takes a name instead, and call `WithWorkflowStorage()` if the value must survive suspension."*
3. Optionally, log a warning at workflow-build time for a variable with no `StorageDriverType`, since it is almost never what the author meant.

### Workaround

```csharp
public static Variable<T> WithPersistentVariable<T>(this IWorkflowBuilder builder, string name, T? value = default) =>
    builder.WithVariable(name, value!).WithStorageDriver<WorkflowInstanceStorageDriver>();
```

### Environment

- **Elsa version:** reproduced on 3.8.0 and 3.8.1; same code path present in 3.7.1 and on `main`
- **Persistence:** `Elsa.Persistence.Dapper` 3.8.1 / SQL Server
- **Runtime:** .NET 10, Windows

sfmskywalker:
## Purpose

Fix #8159: named `WithVariable<T>(string name, T value)` now defaults to workflow instance storage, so values survive suspend/resume like the parameterless overload.

---

## Scope

Select one primary concern:

- [ ] Documentation update
- [ ] New feature

---

### Problem

`IWorkflowBuilder.WithVariable<T>()` calls `WithWorkflowStorage()` (`WorkflowInstanceStorageDriver`). The named overload added for ergonomics never set `StorageDriverType`.

An unset driver means in-memory only for the current expression context. After a bookmark (`Delay`, `Event`, etc.) the value is gone: reference types come back `null`, value types come back `default`. Elsa 3.8 marked the parameterless overload obsolete and pointed people at the named one, so a mechanical migration silently turned persisted variables into memory-only ones. The failure shows up far from the cause (e.g. `InputEvaluationException` on a later activity).

Investigated all `WithVariable` overloads on `WorkflowBuilder` / `IWorkflowBuilder`:

| Overload | Creates variable? | Storage before | Change |
| --- | --- | --- | --- |
| `WithVariable<T>()` | yes | `WithWorkflowStorage()` | none (obsolete) |
| `WithVariable<T>(T value)` | yes (delegates to parameterless) | workflow storage | none (obsolete) |
| `WithVariable<T>(string name, T value)` | yes | **none** | **now calls `WithWorkflowStorage()`** |
| `WithVariable<T>(Variable<T>)` / `WithVariable(Variable)` / `WithVariables(...)` | no (caller-owned) | caller-configured | none — do not override an existing driver |

There is no `WithVariable<T>(string name)` on `main` (it landed in #7701 and is not present on current `main`).

### Solution

Call `WithWorkflowStorage()` from the named factory, same as the parameterless path. Document that default on the public API. No new storage API.

---

## Verification

Steps:
1. `dotnet test test/unit/Elsa.Workflows.Core.UnitTests/Elsa.Workflows.Core.UnitTests.csproj --filter FullyQualifiedName~WorkflowBuilderTests`
2. `dotnet test test/integration/Elsa.Workflows.IntegrationTests/Elsa.Workflows.IntegrationTests.csproj --filter FullyQualifiedName~NamedVariablePersistence`

Expected outcome:
- Named `WithVariable("message", value)` has `StorageDriverType == typeof(WorkflowInstanceStorageDriver)`, matching the parameterless overload.
- A workflow that sets a named variable, bookmarks, and resumes still reads the set value after resume.

Local result: both targeted test runs passed (3 unit + 1 integration).

---

## Commit Convention

- `fix:` – Bug fixes (behavior change)

---

## Checklist

- [x] The PR is focused on a single concern
- [x] Commit messages follow the recommended convention
- [x] Tests added or updated (if applicable)
- [x] Documentation updated (if applicable)
- [x] No unrelated cleanup included
- [x] All tests pass

Fixes #8159

<div><a href="https://cursor.com/agents/bc-0197c670-c6a5-5d89-9be1-19b41f6309cb?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0197c670-c6a5-5d89-9be1-19b41f6309cb&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


sfmskywalker:
@greptileai

sfmskywalker (commented):
## Code Review — Round 1/4 @ `206cde2` → `main`

**Verdict: APPROVE + HIGH** (formal event COMMENT — own-PR APPROVE blocked for `sfmskywalker`)

### What changed (fixes #8159)
- Named `WithVariable<T>(string name, T value)` now calls `WithWorkflowStorage()` — same default as the (obsolete) parameterless overload so values survive suspend/resume.
- Caller-owned `WithVariable(Variable)` / `WithVariables` unchanged (no driver override).
- Docs on `IWorkflowBuilder` updated.
- Unit: named → `WorkflowInstanceStorageDriver`; parity with parameterless; existing `Variable` not overridden.
- Integration: set before Event bookmark → resume still has value.

Clear subtractive fix for the 3.8 migration footgun. Intentional 3.9 behavior change for hosts that relied on memory-only named vars.

### Soft steers (non-blocking)
- Call out in 3.9 release notes for anyone who depended on unnamed persistence vs named in-memory asymmetry (unlikely intentional).

### Merge gates
- Code Review: **APPROVE + HIGH** @ `206cde2`
- Greptile: tip **5/5**
- CI: green on tip

Issue #7173, opened 2025-12-28
ITenantsProvider.ListAsync causes memory/performance bottlenecks with high tenant counts


JoPetra:
Currently, the ITenantsProvider interface requires implementing ListAsync(), which returns Task<IEnumerable<Tenant>>. While this works for small deployments, it does not scale for multi-tenant SaaS environments with hundreds or thousands of tenants.

ListAsync() loading all tenant configurations into memory.

PR #8183, opened 2026-09-19
fix(runtime): process bookmark queue per tenant scope

tulrichtrimble:
## Problem

With multitenancy enabled in Elsa Core (`TenantsFeature`), enqueued bookmarks 
remain stuck indefinitely and fail to resume workflows.

**Root Cause:** `BookmarkQueueWorker.ProcessAsync` executes on a background 
DI scope without established tenant context. When `SetTenantIdFilter` installs 
EF Core global query filters (on all Entity-derived tables when 
`TenantsOptions.IsEnabled = true`), it compares `e.TenantId == dbContext.TenantId`. 
With null ambient tenant, the filter blocks all rows carrying real `TenantId` values.

**Result:** Queue appears empty, bookmark drain silently completes with zero 
items processed. Only bookmarks resumed via HTTP request scope (which resolves 
tenant context) succeed.

## Solution

Modified `BookmarkQueueWorker.ProcessAsync` to detect multitenancy infrastructure 
and iterate tenants when available:

1. Resolve `ITenantsProvider` and `ITenantScopeFactory` from root DI container
2. If unavailable → fall back to single-scope behavior (backward compatible)
3. If available:
   - Fetch all active tenants via `ITenantsProvider.ListAsync()`
   - Process queue once per tenant (including `null` for tenant-agnostic bookmarks)
   - Create tenant-scoped DI container via `ITenantScopeFactory.CreateScope(tenant)`
   - Retrieve `IBookmarkQueueProcessor` from tenant scope
   - Process queue under that tenant's context

This ensures EF Core global filters match the ambient tenant context.

## Impact

- ✅ Fixes bookmark queue drain in multitenancy scenarios
- ✅ Fixes recurring `TriggerBookmarkQueueRecurringTask` backstop
- ✅ Backward compatible: falls back to original behavior when multitenancy not configured
- ✅ No data model changes required
- ✅ No breaking API changes

## Verification Steps

1. Enable multitenancy: `.UseTenants(f => f.UseTenantsProvider<T>())`
2. Enqueue a bookmark before persistence completes
3. Trigger `BookmarkQueueWorker.ProcessAsync()`
4. Verify bookmark is processed and workflow resumes

## Validation

- Tested with ProM workflow engine (multitenancy enabled)
- Confirmed compatibility with `ITenantAccessor`, `ITenantScopeFactory`, and `ITenantsProvider`
- Single clean commit with minimal changes (26 insertions, 3 deletions in one file)


sfmskywalker:
**Triage (catch-up 2026-09-20):** Community fix on `main` — bookmark queue processed per tenant scope. Checks green / MERGEABLE. Classification: **needs-review** (escalate to Crew Lead for merge go). No review labels available on this repo.

sfmskywalker:
@greptile

sfmskywalker:
Thanks a lot for this, @tulrichtrimble. Your diagnosis is spot on: with multitenancy enabled, `BookmarkQueueWorker.ProcessAsync` opens a plain `IServiceScopeFactory` scope, which drops the tenant. The EF tenant filter then sees an empty queue, so queued bookmarks never resume.

We've decided to fix it a bit differently than this PR does, and I wanted to explain why.

**Why not list all tenants here.** Resolving tenants through `ITenantsProvider.ListAsync` adds another place that loads every tenant. #7173 already calls out that pattern as a real problem for systems with many tenants. It also turns out this worker isn't on a slow timer. It's signalled on nearly every workflow instance and bookmark save. And there's already one `BookmarkQueueWorker` per tenant, because each tenant's `TriggerBookmarkQueueRecurringTask` resolves it from that tenant's scope. So with this change, each of the N workers would list all N tenants on almost every save, and the cost would grow with tenant count times write traffic.

**What we're doing instead.** Each worker stays in its own tenant:

1. The worker opens its processing scope with `ITenantScopeFactory.CreateScope(tenant)` for its own tenant instead of a plain scope.
2. The default `BookmarkQueueSignaler` routes signals per tenant. Today it's a single capacity-1 channel, so a signal wakes one arbitrary tenant's worker. The public interface doesn't change.
3. `DistributedBookmarkQueueWorker` includes the tenant id in its lock name, so tenants no longer block each other across nodes.

This path never calls `ListAsync` and needs no store changes. The tests will check that two tenants with queued bookmarks both resume, that `ListAsync` is never called, that an empty queue does nothing, and that a signal from one tenant doesn't wake another tenant's worker.

**Where it lands.** The fix is tracked in #8382 and targets **3.9** (`release/3.9.0`, then forward-ported to `main`). We decided it's too large a change for a 3.8.x patch. The wider "loads every tenant" scaling problem stays in #7173.

We'll close this PR once the fix for #8382 merges, and that PR will credit you for the diagnosis. Thanks again. This was a really helpful find.

sfmskywalker:
Closing as superseded by #8385 (merged into release/3.9.0 as c5472665), which builds on this PR's approach for per-tenant bookmark queue processing (#8382). Thank you @tulrichtrimble for identifying the issue and for the original implementation. It shaped the final fix.

PR #8185, opened 2026-09-19
fix(workflows): isolate nested loop CurrentValue memory blocks

### LINKED ISSUE #7908: Nested loops shadow their parent memory blocks
## Description
If I nest a `ParallelForEach` within a `ForEach`, it's impossible to access the correct `CurrentValue` and `CurrentIndex`. 

## Steps to Reproduce
To help us identify the issue more quickly, please follow these guidelines:

1. **Detailed Steps**: 
- Add a `ForEach` to the workflow
- Add a nested `ParallelForEach` to the workflow (didn't try, but any activities that shadow there parent's variables should be a problem)
- Try to access `currentValue`, `currentIndex`
- I've attached a sample workflow json

## Expected Behavior
`CurrentValue` and `CurrentIndex` should point to the scope of the current loop within `ForEach` and not override the parents variables.

## Actual Behavior
`CurrentValue` and `CurrentIndex` point to values of the last iteration in the `ParallelForEach`.

## Screenshots

Image

Image

[workflow-definition-parallel-for-each-test-28dde9b65c54cab9-4fb7b983f8489938.json](https://github.com/user-attachments/files/30869653/workflow-definition-parallel-for-each-test-28dde9b65c54cab9-4fb7b983f8489938.json)

Image

## Environment
- **Elsa Package Version**: 3.6.1, 3.7.1
- **Operating System**: Windows 11

## Troubleshooting Attempts
I've tried to debug the chain and came down to those two conclusions:
1. Accessing the currentIndex/currentValue variables in an Expression will call `ExpressionExecutionContext.GetVariableInScope`. This will look for memory blocks of those variables, which are not available. Interestingly enough though: `ExpressionExecutionContext.GetVariablesInScope` will find them - there is a fallback to the `activityExecutionContext` implemented:
```
if (!currentScope.TryGetActivityExecutionContext(out var activityExecutionContext))
                {
                    var variables = currentScope.Memory.Blocks.Values
                        .Where(x => x.Metadata is VariableBlockMetadata)
                        .Select(x => x.Metadata as VariableBlockMetadata)
                        .Select(x => x!.Variable)
                        .ToList();

                    foreach (var variable in variables)
                        yield return variable;
                }
                else
                {
                    var variables = activityExecutionContext.Variables;

                    foreach (var variable in variables)
                        yield return variable;
                }
```
2. I've checked why the memory blocks aren't there. `CreateActivityExecutionContextAsync` in `WorkflowExecutionContext` will iterate through `variablesToDeclare` and call `activityExecutionContext.ExpressionExecutionContext.CreateVariable(variable.Name, variable.Value);`. There, a shadow variable will be created:
```
var variable = new Variable(name, value)
            {
                StorageDriverType = storageDriverType ?? typeof(WorkflowInstanceStorageDriver)
            };
```
As we don't provide an id for the variable, it will be the default value `currentIndexVariable`/`currentValueVariable`.  `ExpressionExecutionContext.Set` will then use `GetBlockInternal` to determine if the block exists and as the parent `ForEach` already declares them, it will override the values of those memory-blocks. So I guess the fix needs to be at this place already. I've tried to write my own `ParallelForEach` and schedule the items like this:
```
// Schedule a body of work for each item.
var tag = Guid.NewGuid();
tags.Add(tag);

// For each item, declare a new variable for the work to be scheduled.
// We specify our own variable id, as otherwise there would be clashes with the memory-blocks in nested ForEach/ForEachParallel
var currentValueVariable = new Variable<T>("CurrentValue", item, $"currentValueVariable-{tag}")
{
    // TODO: This should be configurable, because this won't work for e.g. file streams and other non-serializable types.
    StorageDriverType = typeof(WorkflowInstanceStorageDriver)
};

var currentIndexVariable = new Variable<int>("CurrentIndex", currentIndex++, $"currentIndexVariable-{tag}") 
{ 
    StorageDriverType = typeof(WorkflowInstanceStorageDriver) 
};
var variables = new List<Variable> { currentValueVariable, currentIndexVariable };

await context.ScheduleActivityAsync(Body, OnChildCompleted, tag, variables);
```
But due to the shadow variable creation above, the `id` will never be forwarded to the `ExpressionExecutionContext.Set`.

sfmskywalker:
## Purpose

Prevent nested `ForEach` / `ParallelForEach` (and `For`) from overwriting the parent loop's `CurrentValue` and `CurrentIndex` memory blocks.

Closes #7908

## Scope

- [ ] Documentation update
- [ ] New feature

### Problem

When a `ParallelForEach` (or another loop that writes `CurrentValue` / `CurrentIndex`) is nested inside a `ForEach`, the inner loop overwrites the parent's memory-block variables. After the inner loop finishes, expressions that should read the outer `CurrentValue` / `CurrentIndex` see the last inner iteration instead.

The scheduled-variable path in `CreateActivityExecutionContextAsync` called `CreateVariable(name, value)`, which:

1. Built a **new** `Variable` with a name-derived id (`currentValueVariable` / `currentIndexVariable`)
2. `Set()` walked parent expression scopes and reused the parent's existing block

So nested loops shared one memory block instead of isolating per activity scope.

### Solution

- Declare the original scheduled `Variable` (id, storage driver, value) on the **child activity's own memory register**, so `Set()` cannot walk up and overwrite a parent loop block.
- Give each loop scope unique variable ids (`{activityExecutionId}:CurrentValue` for sequential loops, `{tag}:CurrentValue` for `ParallelForEach` iterations) so even id-based lookups stay isolated.
- Cover the regression with a unit test on `CreateActivityExecutionContextAsync` and integration tests for `ForEach` + `ParallelForEach` and `ForEach` + `ForEach`.

Single-loop and sequential nesting behavior is unchanged: name-based `getVariable('CurrentValue')` still resolves the innermost loop while the parent block remains intact after the inner loop completes.

## Verification

Steps:
1. `dotnet test test/unit/Elsa.Workflows.Core.UnitTests/Elsa.Workflows.Core.UnitTests.csproj --filter ScheduledVariableIsolationTests`
2. `dotnet test test/integration/Elsa.Activities.IntegrationTests/Elsa.Activities.IntegrationTests.csproj --filter "FullyQualifiedName~NestedLoopMemoryBlockTests|FullyQualifiedName~ForEachTests|FullyQualifiedName~ParallelForEachTests"`
3. Repro from #7908: nest `ParallelForEach` inside `ForEach` and read the outer `currentValue` / `currentIndex` after the inner loop.

Expected outcome:

- After an inner `ParallelForEach` or `ForEach`, the outer loop still reports its own `CurrentValue` / `CurrentIndex`.
- Inside the inner loop, `CurrentValue` / `CurrentIndex` still refer to the inner iteration.
- Existing single-loop `ForEach` and `ParallelForEach` tests continue to pass.

Local verification (net10.0):

- `ScheduledVariableIsolationTests`: passed
- Nested loop + existing `ForEach` / `ParallelForEach` integration tests: 20 passed
- `Elsa.Activities.UnitTests` looping tests: 74 passed
- Related variable / expression-context unit tests: 36 passed

## Checklist

- [x] The PR is focused on a single concern
- [x] Commit messages follow the recommended convention
- [x] Tests added or updated (if applicable)
- [x] Documentation updated (if applicable)
- [x] All targeted tests pass

<div><a href="https://cursor.com/agents/bc-95499cc6-ef4a-5448-a365-2c260ff5780c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-95499cc6-ef4a-5448-a365-2c260ff5780c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


sfmskywalker:
@greptileai

sfmskywalker (commented):
## Code Review — Round 1/4 (tip-bound)

**Verdict:** APPROVE + **HIGH**  
**Tip:** `0ffc2fa63a339e34dc5d069b4ff42f778ab5ac7b` → `main`  
**Fixes:** #7908

### Finding
Nested loop `CurrentValue` / `CurrentIndex` isolation is fixed at the right layer. `DeclareScheduledVariable` registers the original scheduled `Variable` on the child activity’s own memory instead of `CreateVariable` + `Set` (which walked parents and reused a shared name-derived block). `ForEach` / `For` / `ParallelForEach` also give per-scope variable ids (`{activityExecutionId|tag}:…`), so id-based lookups stay isolated. Name-based `getVariable` still resolves the innermost scope while the outer block survives after the inner loop completes — matching the intended semantics.

The change is small, well-commented (`#7908`), and easy for a human to maintain. Unit coverage of `CreateActivityExecutionContextAsync` plus integration coverage for `ForEach`+`ParallelForEach` and `ForEach`+`ForEach` are the right regression guards; existing single-loop tests remain the baseline.

### Soft steers (non-blocking)
- Merge when tip CI is fully green (`ubuntu-latest` / csharp Analyze were still in progress when reviewed). Greptile had not commented yet — prefer tip **5/5** if it scores; otherwise **APPROVE+HIGH ∧ CI** is enough.
- Optional later: an explicit `ForEach`+`For` nesting case (same declare path, lower urgency).

### Merge gate
Prefer Greptile tip **5/5** when scoring; else **APPROVE+HIGH ∧ CI green**.

---
*Formal event: COMMENT (own-PR APPROVE blocked for author). Round 1 tip-frozen at `0ffc2fa`.*

